PR #5, opened 2018-09-26
Filter by registry namespace

tkashem:
For an `OperatorSource` type, allow the admin to specify a registry namespace to filter by

### DIFF
--- a/pkg/appregistry/adapter.go
+++ b/pkg/appregistry/adapter.go
@@ -15,8 +15,10 @@ const (
 
 // This interface (internal to this package) encapsulates nitty gritty details of go-appr client bindings
 type apprApiAdapter interface {
-	// ListPackages returns a list of package(s) available to the user
-	ListPackages() (appr_models.Packages, error)
+	// ListPackages returns a list of package(s) available to the user.
+	// When namespace is specified, only package(s) associated with the given namespace are returned.
+	// If namespace is empty then visible package(s) across all namespaces are returned.
+	ListPackages(namespace string) (appr_models.Packages, error)
 
 	// GetPackageMetadata returns metadata associated with a given package
 	GetPackageMetadata(namespace string, repository string, release string) (*appr_models.Package, error)
@@ -29,8 +31,12 @@ type apprApiAdapterImpl struct {
 	client *appr.Appregistry
 }
 
-func (a *apprApiAdapterImpl) ListPackages() (appr_models.Packages, error) {
+func (a *apprApiAdapterImpl) ListPackages(namespace string) (appr_models.Packages, error) {
 	params := appr_package.NewListPackagesParams()
+
+	if namespace != "" {
+		params.SetNamespace(&namespace)
+	}
 
 	packages, err := a.client.PackageAppr.ListPackages(params)
 	if err != nil {
--- a/pkg/appregistry/client.go
+++ b/pkg/appregistry/client.go
@@ -6,6 +6,17 @@ import (
 	"strings"
 )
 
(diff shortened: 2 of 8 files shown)


pkg/appregistry/client.go
```diff
@@ -6,6 +6,17 @@ import (
 	"strings"
 )
 
+// Client exposes the functionality of app registry server
+type Client interface {
+	// RetrieveAll retrieves all visible packages from the given source
+	// When namespace is specified, only package(s) associated with the given namespace are returned.
+	// If namespace is empty then visible package(s) across all namespaces are returned.
+	RetrieveAll(namespace string) ([]*OperatorMetadata, error)
```
aravindhp: Like I mentioned above it is cleaner to also have `RetrieveAll()`. You can either add that as part of this PR or open a new PR for that.

pkg/appregistry/adapter.go
```diff
@@ -29,9 +31,13 @@ type apprApiAdapterImpl struct {
 	client *appr.Appregistry
 }
 
-func (a *apprApiAdapterImpl) ListPackages() (appr_models.Packages, error) {
+func (a *apprApiAdapterImpl) ListPackages(namespace string) (appr_models.Packages, error) {
 	params := appr_package.NewListPackagesParams()
 
+	if namespace != "" {
```
aravindhp: Not a huge fan of these sort of interfaces. I would prefer having `ListPackages()` in addition to this API.

pkg/operatorsource/reconciler.go
```diff
@@ -51,7 +51,7 @@ func (r *reconciler) Reconcile(opsrc *v1alpha1.OperatorSource) error {
 		return err
 	}
 
-	manifests, err := registry.RetrieveAll()
+	manifests, err := registry.RetrieveAll(opsrc.Spec.RegistryNamespace)
```
ravisantoshgudimetla: nit: Guarding against nil values is something that should be done.

pkg/appregistry/mock_adapter.go
```diff
@@ -34,16 +34,16 @@ func (m *MockapprApiAdapter) EXPECT() *MockapprApiAdapterMockRecorder {
 }
 
 // ListPackages mocks base method
-func (m *MockapprApiAdapter) ListPackages() (models.Packages, error) {
-	ret := m.ctrl.Call(m, "ListPackages")
+func (m *MockapprApiAdapter) ListPackages(namespace string) (models.Packages, error) {
```
ravisantoshgudimetla: I'd rather have all these moved to a new package called generated, if you want to have the generated packages.

pkg/appregistry/adapter.go
```diff
@@ -29,9 +31,13 @@ type apprApiAdapterImpl struct {
 	client *appr.Appregistry
 }
 
-func (a *apprApiAdapterImpl) ListPackages() (appr_models.Packages, error) {
+func (a *apprApiAdapterImpl) ListPackages(namespace string) (appr_models.Packages, error) {
```
ravisantoshgudimetla: The more I am thinking about it, I feel we should have the args to functions as fields in apprApiAdapterImpl struct

pkg/appregistry/adapter.go
ravisantoshgudimetla: I think logging it here and adding some message to end-user is really useful because.

tkashem:
Merging for now. @ravisantoshgudimetla I will address the feedback separately. Here is a link to the corresponding JIRA issue - https://jira.coreos.com/browse/MKTPLC-76

PR #6, opened 2018-09-28
add documentation on how to push operator manifest to app registry

tkashem:


aravindhp:
Please add documentation for deleting too.

tkashem:
@aravindhp done, take a look when you have time

docs/how-to-upload-manifest.md
```diff
@@ -0,0 +1,120 @@
+# How To Push Operator Manifest To App Registry
```
aravindhp: "Manifest(s)" here too

docs/how-to-upload-manifest.md
```diff
@@ -0,0 +1,120 @@
...
+This document shows how one can push operator manifest(s) into `quay.io` or any other compatible app registry server.
+
+```
```
aravindhp: Why are you using ``` here? This should be a regular paragraph. If you want to call out something, please use bold.

docs/how-to-upload-manifest.md
```diff
@@ -0,0 +1,120 @@
...
+The example script(s) in this document refers to `quay.io`
+```
+
+## Format of Operator Manifest
+```
```
aravindhp: Why are you using ``` here? This should be a regular paragraph. If you want to call out something, please use bold.

docs/how-to-upload-manifest.md
```diff
@@ -0,0 +1,120 @@
...
+publisher: redhat
+
+data:
+  customResourceDefinitions: |-
+  clusterServiceVersions: |-
+  packages: |-
+```
+
+## Pre-Requisite
+```
```
aravindhp: Why are you using ``` here? This should be a regular paragraph. If you want to call out something, please use bold.

docs/how-to-upload-manifest.md
```diff
@@ -0,0 +1,120 @@
...
+You need to have an account with quay.io. If you don't have one you can sign up for it. 
+```
+
+### Obtain Access Token
+All API requests made by executing HTTP verb (GET, POST, PUT or DELETE) against the API endpoint URL require an `Authorization header` with an access token. By logging in to `quay.io` you will receive an access token.  
+
+* Save the following excerpt into a file, for example `login.sh`  
+```bash
```
aravindhp: Please create a scripts directory and reference this script from there?

docs/how-to-upload-manifest.md
```diff
@@ -0,0 +1,120 @@
...
+curl -H "Content-Type: application/json" -XPOST https://quay.io/cnr/api/v1/users/login -d '
+{
+    "user": {
+        "username": "'"${USERNAME}"'",
+        "password": "'"${PASSWORD}"'"
+    }
+}'
+```
+
+* Execute `login.sh` by providing your `quay.io` accouunt username and password as arguments
```
aravindhp: "account" has a spelling error.

docs/how-to-upload-manifest.md
```diff
@@ -0,0 +1,120 @@
...
+```bash
+$ chmod a+x login.sh
+$ ./login.sh {username} {password}
+```
+
+On successful invokation, You will receive a JSON response with the access token, as shown below. Save the value of the `token` field from the JSON response. We will use this token to make calls to `quay.io` API.
```
aravindhp: "invocation" has a spelling error. Please run this through a spell checker in case I missed any errors.

docs/how-to-upload-manifest.md
```diff
@@ -0,0 +1,120 @@
...
+    "error": {
+        "code":"package-exists","details":{},"message":"package exists already"
+    }
+}
+```
+
+If you change your operator manifest then bump up the version in `RELEASE` before pushing the manifest.
+
+### Delete Operator Manifest
+For now, we have been deleting manifest(s) manually by going to `quay.io`. Go to the `Settings` tab of your desired repository and click the `Delete Application` button.
```
aravindhp: Add a note mentioning you need to be admin of the organization

docs/how-to-upload-manifest.md
```diff
@@ -0,0 +1,120 @@
...
+{
+    "token": "basic XWtgc2hlbTpsZWR6ZXCwbYlf"
+}
+```
+
+## Push Operator Manifest
+* Save the manifest into a file, for example `myapp.yaml`
+* Copy the following excerpt and save it into a file, for example `push.sh`. Put appropriate values for the variables.
+
+```bash
```
aravindhp: This can be placed in the `scripts` directory too.

docs/how-to-upload-manifest.md
```diff
@@ -0,0 +1,120 @@
...
+#!/bin/sh
+
+FILENAME="file name (without the extension)"
```
aravindhp: Make these as inputs like you did with the login script.

docs/how-to-upload-manifest.md
```diff
@@ -0,0 +1,120 @@
...
+curl -H "Content-Type: application/json" -H "Authorization: ${TOKEN}" -XPOST https://quay.io/cnr/api/v1/packages/${NAMESPACE}/${REPOSITORY} -d '
+{
+    "blob": "'"${BLOB}"'",
+    "release": "'"${RELEASE}"'",
+    "media_type": "helm"
+}'
+```
+
+If you have saved the manifest into `myapp.yaml` and want to push it to `myoperators/myapp:1.0.0` then the variables will have the following values.
+```bash
```
aravindhp: This can be replaced with a command line example of the script

docs/how-to-upload-manifest.md
tkashem: It's like an info section, looks more aesthetic in my opinion :). I would reserve making an entire line bold for something really really critical.

docs/how-to-upload-manifest.md
tkashem: This document is a guide on how to push manifest to app registry, I don't see it as a tool. I am open to writing a tool (using shell script or golang/python) that automates pushing of operator manifest. Given that we are moving toward image based solution I don't think it's feasible to invest into tooling now.

docs/how-to-upload-manifest.md
tkashem: Don't think it's relevant, it's not a complete solution. See my comments about tooling below.

docs/how-to-upload-manifest.md
tkashem: Inline is fine, I believe. In my mind a script file should be functional. If we turn this solution into a tool then this is what the usage would look like, off the top of my head
```
manifest push {file name} --username={user name} --password={password}
``` 
or change `appr` to recognize operator manifest file.

docs/how-to-upload-manifest.md
tkashem: See my comments below.

docs/how-to-upload-manifest.md
aravindhp: What is aesthetic is a matter of personal choice. I would rather follow the general standard of using ``` ``` to indicate code and stuff that shouldn't be formatted. I definitely want this standard to be followed in our docs.

docs/how-to-upload-manifest.md
aravindhp: Agreed. We can revisit this if we ever go back to using the app registry.

tkashem:
@aravindhp done, give it a look when you have time.

PR #9, opened 2018-10-03
Add Dockerfile to build without operator-sdk

kevinrizza:


Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN make build
+
+FROM centos:7
+RUN useradd marketplace-operator
+USER marketplace-operator
+COPY --from=builder /go/src/github.com/operator-framework/operator-marketplace/tmp/_output/bin/marketplace-operator /usr/bin
+
+LABEL io.k8s.display-name="OpenShift marketplace-operator" \
+      io.k8s.description="This is a component of OpenShift Container Platform and manages the OpenShift Marketplace." \
+      io.openshift.tags="openshift" \
```
tkashem: I would recommend `io.openshift.tags=openshift,marketplace` to be more specific.

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+WORKDIR /go/src/github.com/operator-framework/operator-marketplace
+COPY . .
+RUN make build
+
+FROM centos:7
+RUN useradd marketplace-operator
+USER marketplace-operator
+COPY --from=builder /go/src/github.com/operator-framework/operator-marketplace/tmp/_output/bin/marketplace-operator /usr/bin
+
+LABEL io.k8s.display-name="OpenShift marketplace-operator" \
```
tkashem: For display, `OpenShift Marketplace Operator` maybe?

Dockerfile
```diff
@@ -0,0 +1,18 @@
+
+FROM openshift/origin-release:golang-1.10 as builder
+WORKDIR /go/src/github.com/operator-framework/operator-marketplace
+COPY . .
+RUN make build
```
aravindhp: Is multi-stage builds being done in other operator repos?

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+FROM centos:7
```
aravindhp: Shouldn't this be a RHEL image? We are also now diverting from using a different base image here vs what we are using in `tmp\build\Dockerfile`. Apparently that is actually not temporary from a SDK perspective which is why it has been checked in. So we should closely align the two otherwise the image we are developing with will differ from what we are shipping.

Dockerfile
aravindhp: +1

Makefile
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+all: build
+
+build:
```
aravindhp: Please call this `osbs-build` as I can see us using this Makefile for other purposes too.

Dockerfile
aravindhp: We should follow whatever convention that needs to be followed for these tags.

Dockerfile
kevinrizza: I'm going off this example, which uses the multistage build:

https://github.com/openshift/console-operator/blob/master/Dockerfile

Dockerfile
kevinrizza: Is

`FROM alpine:3.6`

A RHEL image? 

I was using the console-operator as an example, and assumed that what they were doing is correct. I can definitely update this to use RHEL or the same base image that is being used in the generated dockerfile, I just am not sure which you think makes the most sense.

aravindhp: `alpine` is not a RHEL image. It is the default image that the SDK uses. Is there a mandate from OSBS as to what the base image to use? If not we should follow what other OpenShift operators are basing their image off of.

aravindhp:
Please use proper case in your commit message and follow the 50/72 rule

Dockerfile
kevinrizza: Fair, though I'm not exactly sure what your suggestion is here. Are you saying we should update this Docker image to try to match what's in the dockerfile in tmp\build ? Or are you just referring to the base image itself? I do think it makes sense to update this base image to use the same one as what the operator-sdk uses, though maybe it makes sense to reach out to them and make sure that should not be RHEL? I don't believe there is anything about the OSBS build that requires RHEL, as far as I know.

I think no matter what we do here there is going to be some pain in trying to manage the fact that OSBS is requiring a dockerfile in the root of the directory, and that that file isn't going to be built with the same method as the operator-sdk for right now.

aravindhp: If OSBS doesn't have a requirement for it to be RHEL, then we should stick to using what the SDK started with which is Alpine. I prefer that given it is more lightweight. We need to be aware that any changes to the SDK Dockerfile will need to be reflected here too.

PR #12, opened 2018-10-16
Add unit tests rule to Makefile.

kevinrizza:


kevinrizza:
/cc @aravindhp 

Looks like it didn't add you by default.

Makefile
```diff
@@ -20,6 +20,9 @@ osbs-build:
 	# hack/build.sh
 	./tmp/build/build.sh
 
+unit:
+	go test -v ./pkg/...
```
aravindhp: Is this what other teams are doing? We should ensure that this doesn't cause trouble when launched by the CI framework. Is there a way to do that?

kevinrizza: It's being used exactly this way by the OLM team. I actually just copied their pattern for this bit:

https://github.com/operator-framework/operator-lifecycle-manager/blob/master/Makefile#L21

https://github.com/openshift/release/blob/master/ci-operator/config/operator-framework/operator-lifecycle-manager/operator-framework-operator-lifecycle-manager-master.yaml#L38

aravindhp:
@tkashem please review

Makefile
tkashem: @kevinrizza My bad, I forgot to mention that we do not commit the generated mock files into git. So we would need to generate the mock files before the unit tests can be run.  `make generate-mocks` will generate the mocks for you in the right place.

aravindhp: Please open a Jira bug to track this.

PR #13, opened 2018-10-16
Update Dockerfile base image

kevinrizza:
-Use openshift/origin-base when building image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,8 +4,7 @@ WORKDIR /go/src/github.com/operator-framework/operator-marketplace
 COPY . .
 RUN make osbs-build
 
-FROM alpine:3.6
-RUN apk update && apk add ca-certificates && rm -rf /var/cache/apk/*
+FROM openshift/origin-base
 RUN useradd marketplace-operator
 USER marketplace-operator
 COPY --from=builder /go/src/github.com/operator-framework/operator-marketplace/tmp/_output/bin/marketplace-operator /usr/bin
--- a/tmp/build/Dockerfile
+++ b/tmp/build/Dockerfile
@@ -1,7 +1,6 @@
-FROM alpine:3.6
+FROM openshift/origin-base
 
-RUN apk update && apk add ca-certificates && rm -rf /var/cache/apk/*
-RUN adduser -D marketplace-operator
+RUN useradd marketplace-operator
 USER marketplace-operator
 
 ADD tmp/_output/bin/marketplace-operator /usr/local/bin/marketplace-operator


Dockerfile
```diff
@@ -4,8 +4,7 @@ WORKDIR /go/src/github.com/operator-framework/operator-marketplace
 COPY . .
 RUN make osbs-build
 
-FROM alpine:3.6
-RUN apk update && apk add ca-certificates && rm -rf /var/cache/apk/*
```
aravindhp: You have do something similar with the origin-image too otherwise the operator will have trouble talking to Quay.

kevinrizza: @aravindhp I don't think the image built here will need to talk to Quay, since this image should just get pushed up to the openshift internal CI registry

aravindhp: So this is not the image  that will be used for the product? What if we need to run e2e tests that need to talk to Quay? In other words, what exactly does this image get used for?

kevinrizza: This image gets built and published up to an image stream for the internal registry for the openshift ci instance (https://api.ci.openshift.org/console/). That registry contains all the built images from our ci process, but the image we release should be from the OSBS build and not this one. The e2e tests would be run from it when we added them.

If this dockerfile is temporary we probably want to put as little config in it as possible. Do we have a real dependency on making api calls to Quay right now? If not, maybe we should hold off on adding this until it's needed.

aravindhp: Yes, we do have a real dependency on making API calls to Quay. I am not sure what you want to hold off on. Plus I don't understand why this Dockerfile would be temporary if it is needed to get the product released.

kevinrizza: If we have a real dependency on external apis right now, we can make a change to have similar functionality for now on top of the new base image.

Just fyi we will eventually need to replicate this stuff for our operator-sdk build.

aravindhp: Once you figured out how to install CA certs, please try using `openshift/origin-base` image in the SDK Dockerfile. That way we are using the same base image in all scenarios.

tkashem: +1

aravindhp: Please test this with an OperatorSource CR to ensure we are able to pull the artefacts from Quay appregistry, make the same changes to the SDK Dockerfile and we can merge this PR.

kevinrizza: @aravindhp tested and it successfully communicates with quay. I also updated the sdk dockerfile to use the same base image.

aravindhp: >fyi it looks like a default ca certificate is available on our base image and any fedora/redhat install since fedora 19, so we don't actually need to take action here:

Please mention the above in your commit message otherwise we are going to forget why we removed `RUN apk update && apk add ca-certificates && rm -rf /var/cache/apk/*`

kevinrizza: All set

PR #15, opened 2018-10-18
Get all dependencies for unit tests

kevinrizza:
-Update the make unit call to get mockgen dependency
-Pull down the mockgen repo before installing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,11 +20,13 @@ osbs-build:
 	# hack/build.sh
 	./tmp/build/build.sh
 
-unit:
+unit: generate-mocks unit-test
+
+unit-test:
 	go test -v ./pkg/...
 
 generate-mocks:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -20,11 +20,13 @@ osbs-build:
 	# hack/build.sh
 	./tmp/build/build.sh
 
-unit:
+unit: generate-mocks unit-test
```
SamiSousa: Does `unit` follow a convention? Maybe this command could be named `test`

kevinrizza: I don't think 'test' makes sense, since we want to leave room for more than one test type. In the future, we could also have something like 'functional' and 'component' that also refer to specific tests. At that point, we could use something like 'test' that would refer to 'unit' 'component' and 'functional' as a group.

PR #17, opened 2018-10-18
Add service account definition

kevinrizza:
-Create sa.yaml to conform to operator-framework format

### DIFF
--- a/deploy/rbac.yaml
+++ b/deploy/rbac.yaml
@@ -45,7 +45,7 @@ metadata:
   name: default-account-marketplace-operator
 subjects:
 - kind: ServiceAccount
-  name: default
+  name: marketplace-operator-serviceaccount
   namespace: marketplace
 roleRef:
   kind: ClusterRole
--- a/deploy/sa.yaml
+++ b/deploy/sa.yaml
@@ -1,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: marketplace-operator-serviceaccount
+  namespace: marketplace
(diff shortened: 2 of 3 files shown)


deploy/sa.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: default
```
aravindhp: Please call it `marketplace-operator-serviceaccount.` This change should also be reflected in rbac.yaml. And please run the deployment to ensure that everything works as before.

deploy/rbac.yaml
```diff
@@ -45,7 +45,7 @@ metadata:
   name: default-account-marketplace-operator
 subjects:
 - kind: ServiceAccount
-  name: default
+  name: marketplace-operator-serviceaccount
   namespace: marketplace
```
tkashem: You would also need to change change the deployment spec to use this service account in `operator.yaml` file, I believe.
```
   spec:
      serviceAccountName: marketplace-operator-serviceaccount
```

SamiSousa:
For #19, should be setting `namespace: marketplace` in `operator.yaml` so that marketplace operator's namespace matches the namespace for CatalogSourceConfig and OperatorSource.

deploy/sa.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: marketplace-operator-serviceaccount
+  namespace: marketplace
```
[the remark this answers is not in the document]
SamiSousa: I also recommend setting `namespace: marketplace` for the example CRs as well.

PR #19, opened 2018-10-23
CatalogSourceConfig requires field "metadata.namespace"

SamiSousa:
I noticed that the marketplace operator expects CatalogSourceConfig CRs to specify the field "metadata.namespace" with the namespace that created it.

### DIFF
--- a/deploy/examples/catalogsourceconfig.cr.yaml
+++ b/deploy/examples/catalogsourceconfig.cr.yaml
@@ -2,6 +2,7 @@ apiVersion: "marketplace.redhat.com/v1alpha1"
 kind: "CatalogSourceConfig"
 metadata:
   name: "example"
+  namespace: "marketplace"
 spec:
   targetNamespace: "global"
   packages: "operators/marketplace-operators"
--- a/deploy/examples/operatorsource.cr.yaml
+++ b/deploy/examples/operatorsource.cr.yaml
@@ -2,6 +2,7 @@ apiVersion: "marketplace.redhat.com/v1alpha1"
 kind: "OperatorSource"
 metadata:
   name: "global-operators"
+  namespace: "marketplace"
 spec:
   type: appregistry
   endpoint: "https://quay.io/cnr"


SamiSousa:
@aravindhp @kevinrizza @tkashem @awgreene @galletti94

awgreene:
The metadata you added specifies that the namespace is `default` but it  looks like the namespace is set to `global` in the spec. Is this intended?

SamiSousa:
The namespace under metadata should be the namespace where the CSC is created, presumably the namespace the marketplace operator resides in. The targetNamespace under spec is the namespace that the CSC points to, which is where the corresponding CatalogSource should be created.

If `global` is the namespace where the marketplace operator is being deployed, then metadata.namespace should be set to `global`

deploy/examples/catalogsourceconfig.cr.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: "marketplace.redhat.com/v1alpha1"
 kind: "CatalogSourceConfig"
 metadata:
   name: "example"
+  namespace: "default"
```
awgreene: The metadata you added specifies that the namespace is default but it looks like the namespace is set to global in the spec. Is this discrepancy intended?

SamiSousa: The namespace under metadata should be the namespace where the CSC is created, presumably the namespace the marketplace operator resides in. The targetNamespace under spec is the namespace that the CSC points to, which is where the corresponding CatalogSource should be created.

If `global` is the namespace where the marketplace operator is being deployed, then metadata.namespace should be set to `global`

SamiSousa: Might be clearer to put the namespace `marketplace` rather than `default` if that's where the marketplace operator is intended to be deployed.

aravindhp: Yes `marketplace` would be the better option. I also think rather than making piecemeal changes like this, it would be better if this was tackled from a holistic stand point of deploying the marketplace operators with proper naming prefixes etc.

tkashem: And I would also suggest adding `namespace: marketplace` for `operatorsource.cr.yaml` as well just to be consistent.

aravindhp:
Please reword your commit message by just stating the problem without using the first person. Moreover you are not setting the `namespace` in the `operator.yaml` for consistency. If you did not do this, the example `CatalogSourceConfig` would not work.

SamiSousa:
> Moreover you are not setting the namespace in the `operator.yaml` for consistency. If you did not do this, the example `CatalogSourceConfig` would not work.
#17 should be setting namespace in `operator.yaml`

aravindhp:
> > #17 should be setting namespace in `operator.yaml`

Please make this comment in that PR. This is also why I said previously, `Rather than making piecemeal changes like this, it would be better if this was tackled from a holistic stand point of deploying the marketplace operator with proper naming prefixes etc.`.

SamiSousa:
Closing in favor of adding these changes as a part of #17 as a more holistic solution.

aravindhp:
Looks like there are more changes coming down with pipe with migration to SDK 0.1.0. Given that it makes sense to just merge the example changes you made. So I will reopen and merge this in.

PR #25, opened 2018-11-07
Updating README to include flow description and deploy using OLM

galletti94:
Thought we could use a section to explain the flow of the operator (even if subject to change in the future). Also including a description of how to deploy marketplace using OLM (for when @awgreene merges the CSV).

Let me know if you think this is needed - happy to change the wording, style etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,14 +22,22 @@ $ export REGISTRY=<SOME_REGISTRY> \
 ```
 2. Update the `deploy/operator.yaml` to pull the Marketplace Operator image you just pushed. You should update the `spec.template.spec.containers[0].image` field with the `$REGISTRY/$NAMESPACE/$REPOSITORY:$TAG` value.
 
-## Deploying the Marketplace Operator
+## Using the Marketplace Operator
 
-### In an OKD Cluster
+### Description
+
+The marketplace operator manages two CRDs: [OperatorSource](./deploy/operatorsource.crd.yaml) and [CatalogSourceConfig](./deploy/catalogsourceconfig.crd.yaml). When an OperatorSource CR is created in the same namespace as where the marketplace operator is running (we recommend the namespace be called "marketplace"), the operator will download manifests stored in the registry specified in this OperatorSource CR (for now, please see documentation about using [quay](https://quay.io)'s appregistry API). For an example of this OperatorSource CR please see the [examples](./deploy/examples/) folder.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,14 +22,26 @@ $ export REGISTRY=<SOME_REGISTRY> \
 ```
 2. Update the `deploy/operator.yaml` to pull the Marketplace Operator image you just pushed. You should update the `spec.template.spec.containers[0].image` field with the `$REGISTRY/$NAMESPACE/$REPOSITORY:$TAG` value.
 
-## Deploying the Marketplace Operator
+## Using the Marketplace Operator
 
-### In an OKD Cluster
+### Description
+
+The marketplace operator manages two CRDs: [operatorsource](./deploy/operatorsource.crd.yaml) and [catalogsourceconfig](./deploy/catalogsourceconfig.crd.yaml). When an operatorsource CR is created in the same namespace as where the marketplace operator is running, the marketplace operator will download manifests stored in the external registry specified in this operatorsource CR (for now, please see documentation about using quay's appregistry API). For an example of this operatorsource CR please see the [examples](./deploy/examples/) folder.
```
aravindhp: Please use 'OperatorSource` and `CatalogSource` etc when mentioning CRs.

README.md
```diff
@@ -22,14 +22,26 @@ $ export REGISTRY=<SOME_REGISTRY> \
...
+#### In an OKD Cluster
 ```bash
-$ oc apply -f deploy
+$ oc apply -f deploy/
 ```
 
-### In a Kubernetes Cluster
+#### In the UI
+
+When deploying OLM, include the contents of the CSV file located in the [deploy](./deploy/) folder in the rh-operators configmap shipped with the latest version of OLM. Follow the remaining instructions for deploying OLM. Finally, navigate to the `package manifest` navigation tab in the `Operators` section, find the marketplace operator, and follow the steps after clicking `Create Subscription` in the OLM UI to deploy the operator.
```
aravindhp: This should go in only after we have added the CSV to repo. Moreover once that is in and is part of rh-operators, you only need to mention how to subscribe to it. It should also be called out that this is a OpenShift UI.

README.md
```diff
@@ -22,14 +22,26 @@ $ export REGISTRY=<SOME_REGISTRY> \
...
+
+#### In a Kubernetes Cluster
 ```bash
-$ kubectl apply -f deploy
+$ kubectl apply -f deploy/
```
aravindhp: This would need to be updated if we want users to go the CSV route.

galletti94:
@aravindhp I think I addressed all the issues you raised - let me know if there is anything else you'd like to tweak

galletti94:
@aravindhp is this good to go?

PR #31, opened 2018-11-16
Make MemoryDatastore a Singleton

kevinrizza:
-Remove New() memoryDatastore method
-Initialize datastore in init() func

### DIFF
--- a/pkg/operatorsource/operatorsource.go
+++ b/pkg/operatorsource/operatorsource.go
@@ -10,7 +10,7 @@ import (
 // NewHandler returns an instance of the Handler interface
 // that can be used to reconcile an OperatorSource type object.
 func NewHandler() (Handler, datastore.Reader) {
-	datastore := datastore.New()
+	datastore := datastore.Cache
 	kubeclient := kube.New()
 	transitioner := phase.NewTransitioner()
 
--- a/pkg/datastore/datastore.go
+++ b/pkg/datastore/datastore.go
@@ -5,9 +5,12 @@ import (
 	"strings"
 )
 
-// New returns a new instance of datastore for Operator Manifest(s).
-func New() *memoryDatastore {
-	return &memoryDatastore{
+var (
+	Cache *memoryDatastore
+)
+
+func init() {
+	Cache = &memoryDatastore{
 		manifests: map[string]*OperatorManifest{},
 		parser:    &manifestYAMLParser{},
 	}


pkg/operatorsource/operatorsource.go
```diff
@@ -10,7 +10,7 @@ import (
 // NewHandler returns an instance of the Handler interface
 // that can be used to reconcile an OperatorSource type object.
 func NewHandler() (Handler, datastore.Reader) {
-	datastore := datastore.New()
+	datastore := datastore.MemoryDatastore
```
tkashem: Can we avoid stuttering by renaming `MemoryDatastore` to `Memory` / `Transient` or something appropriate?

kevinrizza: Updated to `Cache`

kevinrizza:
@aravindhp To address your comment on the commit message, I'm not sure I follow your second point. We no longer need to pass the datastore around. It just gets initialized in the init() function when it's first referenced, and at that point the cache should be able to be referenced globally by using datastore.Cache. I think leaving the code as is otherwise is the best option, because this variable should just get replaced by some kind of client eventually.

aravindhp:
> @aravindhp To address your comment on the commit message, I'm not sure I follow your second point. We no longer need to pass the datastore around. It just gets initialized in the init() function when it's first referenced, and at that point the cache should be able to be referenced globally by using datastore.Cache. I think leaving the code as is otherwise is the best option, because this variable should just get replaced by some kind of client eventually.

I was talking about the [handler.NewHandler()](https://github.com/operator-framework/operator-marketplace/blob/master/pkg/stub/handler.go#L15) function. I was under the impression you were going to not return `datastore.Reader` in [operatorsource.NewHandler()](https://github.com/operator-framework/operator-marketplace/blob/master/pkg/operatorsource/operatorsource.go#L12) and have it be initialized inside [catalogsourceconfig.NewHandler()](https://github.com/operator-framework/operator-marketplace/blob/master/pkg/catalogsourceconfig/catalogsourceconfig.go#L10). Maybe I misunderstood.

Either way please mention the rationale behind this (required to move to v0.1.0 etc) in the commit message.

kevinrizza:
@aravindhp ah I see what you meant. Yeah I didn't bother because those files are getting blown away, and datastore is just referenced directly in the controller add functions.

I updated the commit message to reference the sdk update.

PR #33, opened 2018-11-20
Improve crds

aravindhp:
Add
 - Annotations
 - Short name
 - Additional printer columns
 - Open API V3 Schema validation

### DIFF
--- a/deploy/catalogsourceconfig.crd.yaml
+++ b/deploy/catalogsourceconfig.crd.yaml
@@ -2,6 +2,9 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: catalogsourceconfigs.marketplace.redhat.com
+  annotations:
+    displayName: Catalog Source Config
+    description: Represents a CatalogSourceConfig.
 spec:
   group: marketplace.redhat.com
   names:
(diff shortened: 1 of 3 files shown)


deploy/catalogsourceconfig.crd.yaml
```diff
@@ -2,12 +2,57 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: catalogsourceconfigs.marketplace.redhat.com
+  annotations:
+    displayName: Catalog Source Config
+    description: Represents an CatalogSourceConfig.
```
SamiSousa: `Represents a CatalogSourceConfig.`

deploy/catalogsourceconfig.crd.yaml
```diff
@@ -2,12 +2,57 @@ apiVersion: apiextensions.k8s.io/v1beta1
...
   scope: Namespaced
   version: v1alpha1
+  additionalPrinterColumns:
+  - name: TargetNamespace
+    type: string
+    description: The namespace where the operators will be enabled
+    JSONPath: .spec.targetNamespace
+  - name: Packages
+    type: string
+    description: List of operator(s) which will be enabled in the target namespace
```
SamiSousa: I think this should mention how to format the list. I recall it being comma separated without spaces, it should be specified here.

deploy/operatorsource.crd.yaml
```diff
@@ -2,12 +2,65 @@ apiVersion: apiextensions.k8s.io/v1beta1
...
+    description: Represents an OperatorSource.
 spec:
   group: marketplace.redhat.com
   names:
     kind: OperatorSource
     listKind: OperatorSourceList
     plural: operatorsources
     singular: operatorsource
+    shortNames:
+    - os
```
tkashem: `os` or `opsrc`? Mostly to avoid an overloaded name like `os` which is also used as shorthand for `OpenShift`

deploy/operatorsource.crd.yaml
```diff
@@ -2,12 +2,65 @@ apiVersion: apiextensions.k8s.io/v1beta1
...
+          - spec
+          type: object
+          description: Spec for an OperatorSource.
+          required:
+          - type
+          - endpoint
+          - registryNamespace
+          properties:
+            type:
+              type: string
```
tkashem: Can you specify a predefined set of values here?

aravindhp: Not that I know of for strings. What predefined set of values would you want here?

aravindhp:
@tkashem @SamiSousa I have addressed your comments. Please review.

deploy/operatorsource.crd.yaml
aravindhp: I have addressed using a pattern which is just `appregistry` for now. Please review.

deploy/operatorsource.crd.yaml
aravindhp: Done.

PR #35, opened 2018-11-26
Update dockerfiles for osbs

kevinrizza:
-Use openshift ci registry for base images
-Add downstream ocp dockerfile

### DIFF
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -4,10 +4,10 @@ WORKDIR /go/src/github.com/operator-framework/operator-marketplace
 COPY . .
 RUN make osbs-build
 
-FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
+FROM registry.svc.ci.openshift.org/ocp/4.0:base
 RUN useradd marketplace-operator
 USER marketplace-operator
-COPY --from=builder /go/src/github.com/operator-framework/operator-marketplace/build/_output/bin/marketplace-operator /usr/bin
+COPY --from=builder /go/src/github.com/operator-framework/operator-marketplace/tmp/_output/bin/marketplace-operator /usr/bin
 
 LABEL io.k8s.display-name="OpenShift Marketplace Operator" \
       io.k8s.description="This is a component of OpenShift Container Platform and manages the OpenShift Marketplace." \
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,4 +1,4 @@
-FROM openshift/origin-base
+FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -1,10 +1,10 @@
 
-FROM openshift/origin-release:golang-1.10 as builder
+FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.10 AS builder
```
aravindhp: Is VPN access required for this? We need ensure this works if anyone tried using the operator. Moreover I see that the OLM [Dockerfile](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/Dockerfile#L1) does not do this and they have gone through the OSBS process.

tmp/build/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openshift/origin-base
+FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
```
aravindhp: Same question here

Dockerfile.rhel7
```diff
@@ -0,0 +1,18 @@
+
```
aravindhp: If this is this identical to the Dockerfile won't a symbolic link suffice?

kevinrizza: They are not identical, the base images are different.

Dockerfile
kevinrizza: I did just test this and it does work without VPN. Moreover, I believe this is a standard that the release team is trying to enforce. Take a look at this documentation for more detail:

https://mojo.redhat.com/docs/DOC-1179058#jive_content_id_Request_Addition_to_ART_Build_System

There are several operator repos that are using these versions of the base image. It seems likely that the OLM source code is just not following the standard.

kevinrizza: For example:

https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/Dockerfile.rhel7#L1

https://github.com/openshift/cluster-samples-operator/blob/master/Dockerfile#L1

https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/Dockerfile.rhel7#L1

https://github.com/openshift/cluster-ingress-operator/blob/master/Dockerfile.rhel7#L1

kevinrizza:
@tkashem @aravindhp Merged to master due to conflicts, so this needs to be re-approved (no actual changes, just a file path change for the sdk dockerfile). Can someone double check?

aravindhp:
Why are we seeing a merge commit? Did you rebase correctly?

kevinrizza:
> Why are we seeing a merge commit? Did you rebase correctly?
 
Must not have rebased properly last time, I just did it again and there is only one commit. Should be good now.

PR #42, opened 2018-11-30
Reconcile OperatorSource Spec update

tkashem:
When an admin updates the spec of an OperatorSource object we need to reconcile so that the new spec takes effect.

What needs to be done:
* Save OperatorSource spec to datastore. 
* Determine if the Spec has been changed, we will consider change in spec as an update event.
* Reconcile update event appropriately.

### DIFF
--- a/pkg/apis/marketplace/v1alpha1/operatorsource_types.go
+++ b/pkg/apis/marketplace/v1alpha1/operatorsource_types.go
@@ -1,6 +1,8 @@
 package v1alpha1
 
 import (
+	"strings"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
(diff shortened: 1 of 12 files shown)


pkg/datastore/datastore.go
```diff
@@ -183,6 +203,25 @@ func (ds *memoryDatastore) GetPackageIDs() string {
 	return strings.Join(keys, ",")
 }
 
+func (ds *memoryDatastore) RegisterOperatorSource(opsrc *v1alpha1.OperatorSource) {
+	ds.rows[opsrc.GetUID()] = &operatorSourceRow{
```
aravindhp: This feels more like an addition rather than a registration. `AddOperatorSource` or `InsertOperatorSource` especially given the counter is `RemoveOperatorSource`. Or remove should be called `UnregisterOperatorSource`

pkg/apis/marketplace/v1alpha1/operatorsource_types.go
```diff
@@ -66,3 +64,26 @@ func (opsrc *OperatorSource) EnsureGVK() {
...
 }
+
+// IsEqual returns true if the Spec specified in this is the same as the other.
+// Otherwise, the function returns false.
+//
+// The function performs a case insensitive comparison of corresponding
+// attributes.
+//
+// If the Spec specified in other is nil then the function returns false.
+func (this *OperatorSourceSpec) IsEqual(other *OperatorSourceSpec) bool {
```
aravindhp: Java `this`? :-) How about `spec` or `o` or `s`?

PR #45, opened 2018-12-04
Add Marketplace e2e tests

kevinrizza:


test/e2e/marketplace_test.go
```diff
@@ -0,0 +1,186 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+var (
```
SamiSousa: I think `const` might be better here since they don't change.

test/e2e/marketplace_test.go
```diff
@@ -0,0 +1,186 @@
...
+
+	// Now lets update the catalogsourceconfig and see that the configmap and catalog source are updated
+	// First set the target namespace field for the catalogsourceconfig to a new namespace
+	testCatalogSourceConfig.Spec.TargetNamespace = testNamespace
+	err = f.Client.Update(context.TODO(), testCatalogSourceConfig)
+	if err != nil {
+		return err
+	}
+
+	// Check for the config map created in the new namespace
```
SamiSousa: Should the old `CatalogSource` and `ConfigMap` be automatically deleted upon updating a `CatalogSourceConfig`?

test/e2e/marketplace_test.go
```diff
@@ -0,0 +1,186 @@
...
+		t.Fatal(err)
+	}
+}
+
+func WaitForResult(t *testing.T, f *framework.Framework, result runtime.Object, namespace, name string) error {
+	namespacedName := types.NamespacedName{Name: name, Namespace: namespace}
+	err := wait.Poll(retryInterval, timeout, func() (done bool, err error) {
+		err = f.Client.Get(context.TODO(), namespacedName, result)
+		if apierrors.IsNotFound(err) {
+			if err != nil {
```
SamiSousa: Is this check necessary? The two return statements following are equivalent, so is this just to control when to print logs?

test/e2e/e2e-tests.sh
```diff
@@ -0,0 +1,43 @@
...
+CATALOG_SOURCE_CONFIG_CRD="./deploy/crds/catalogsourceconfig.crd.yaml"
+OPERATOR="./deploy/operator.yaml"
+SERVICE_ACCOUNT="./deploy/service_account.yaml"
+ROLE_BINDING="./deploy/role_binding.yaml"
+ROLE="./deploy/role.yaml"
+
+# Create openshift resources if they don't exist
+echo "Creating openshift resources"
+oc apply -f $OPERATOR_SOURCE_CRD
+oc apply -f $CATALOG_SOURCE_CONFIG_CRD
```
aravindhp: Any reason not to use the `oc get --ignore-not-found` for the CRDs?

test/e2e/marketplace_test.go
```diff
@@ -0,0 +1,186 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	apis "github.com/operator-framework/operator-marketplace/pkg/apis"
```
aravindhp: Any reason for the explicit name given it will be called `apis` without it?

test/e2e/marketplace_test.go
```diff
@@ -0,0 +1,186 @@
...
+	operator "github.com/operator-framework/operator-marketplace/pkg/apis/marketplace/v1alpha1"
+
+	framework "github.com/operator-framework/operator-sdk/pkg/test"
```
aravindhp: Is the naming for clarity when reading the code as I don't see another `test` package being imported here?

test/e2e/marketplace_test.go
```diff
@@ -0,0 +1,186 @@
...
+	"github.com/operator-framework/operator-sdk/pkg/test/e2eutil"
+
+	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
aravindhp: Same comment as for `framework`

test/e2e/marketplace_test.go
```diff
@@ -0,0 +1,186 @@
...
+
+func TestMarketplace(t *testing.T) {
+	operatorsourceList := &operator.OperatorSource{
+		TypeMeta: metav1.TypeMeta{
+			Kind: operator.OperatorSourceKind,
+			APIVersion: fmt.Sprintf("%s/%s",
+				operator.SchemeGroupVersion.Group, operator.SchemeGroupVersion.Version),
+		},
+	}
+	catalogsourceconfigList := &operator.CatalogSourceConfig{
```
aravindhp: I suggest using the builders for OperatorSource and CatalogSource here. Or are you worried about using them because they don't have unit tests? Comment applies to all places where you are creating these CRs.

test/e2e/marketplace_test.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		t.Fatalf("failed to add catalogsourceconfig custom resource scheme to framework: %v", err)
+	}
+	// run subtests
+	t.Run("marketplace-group", func(t *testing.T) {
+		t.Run("Cluster", MarketplaceCluster)
+	})
+}
+
+func defaultCreateTest(t *testing.T, f *framework.Framework, ctx *framework.TestCtx) error {
```
aravindhp: Please add a comment as to what the test does.

test/e2e/marketplace_test.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err = defaultCreateTest(t, f, ctx); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func WaitForResult(t *testing.T, f *framework.Framework, result runtime.Object, namespace, name string) error {
```
aravindhp: Please add a doc

test/e2e/marketplace_test.go
```diff
@@ -0,0 +1,186 @@
...
+	err = WaitForResult(t, f, &olm.CatalogSource{}, testNamespace, catalogSourceName)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func MarketplaceCluster(t *testing.T) {
+	t.Parallel()
```
aravindhp: What does this parallelize?

test/e2e/marketplace_test.go
aravindhp: Yeah, how can `err` be `nil` at this point?

test/e2e/marketplace_test.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Then check for the catalog source.
+	err = WaitForResult(t, f, &olm.CatalogSource{}, namespace, catalogSourceName)
+	if err != nil {
+		return err
+	}
+
```
aravindhp: Please assert that the CatalogSource points to the expected ConfigMap.

test/e2e/e2e-tests.sh
```diff
@@ -0,0 +1,43 @@
+TEST_NAMESPACE="openshift-marketplace"
```
aravindhp: Please add `#!/bin/bash`
Please add `set -e` so that it stops on the first error.

test/e2e/marketplace_test.go
```diff
@@ -0,0 +1,186 @@
...
+	})
+}
+
+func defaultCreateTest(t *testing.T, f *framework.Framework, ctx *framework.TestCtx) error {
+	namespace, err := ctx.GetNamespace()
+	if err != nil {
+		return fmt.Errorf("could not get namespace: %v", err)
+	}
+
+	testOperatorSource := &operator.OperatorSource{
```
aravindhp: Is there a way to use the example CR here?

test/e2e/marketplace_test.go
tkashem: I have misspoken, it will not panic.

test/e2e/e2e-tests.sh
kevinrizza: oc apply does an update if the resource type already exists. the oc get for the namespaces is there because if the namespace already exists it will fail

test/e2e/marketplace_test.go
kevinrizza: Right now nothing because the test is only called once. It's a flag for test framework that, if other tests are called at the same time, they will be run in parallel rather than serially.

test/e2e/marketplace_test.go
kevinrizza: Afaik the only way for us to do that would be to parse it and pass it along. The client in the test framework doesn't have the capability to read yaml files.

test/e2e/marketplace_test.go
```diff
@@ -0,0 +1,186 @@
...
+var (
+	retryInterval        = time.Second * 5
+	timeout              = time.Second * 60
+	cleanupRetryInterval = time.Second * 1
+	cleanupTimeout       = time.Second * 5
+	testNamespace        = "operator-test"
+)
+
+func TestMarketplace(t *testing.T) {
+	operatorsourceList := &operator.OperatorSource{
```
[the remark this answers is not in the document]
kevinrizza: Yeah, it was copied from here:

test/e2e/marketplace_test.go
```diff
@@ -0,0 +1,186 @@
...
+			Kind: operator.OperatorSourceKind,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "global-operators",
+			Namespace: namespace,
+		},
+		Spec: operator.OperatorSourceSpec{
+			Type:              "appregistry",
+			Endpoint:          "https://quay.io/cnr",
+			RegistryNamespace: "operators",
```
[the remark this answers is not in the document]
kevinrizza: It might actually make sense for us to have a separate registry specifically for these e2e tests that differs from our example (at least in terms of if and when it is updated).

test/e2e/marketplace_test.go
kevinrizza: Yes, I think that adding an additional import (since right now we are only importing the api package which was created by the framework) to use the builders is extra code that could itself introduce unexpected bugs. I do think if they were well tested ahead of time (with something like unit tests) then they would be safer to use.

test/e2e/marketplace_test.go
aravindhp: Do we want this behavior by default? What if we introduce an update CatalogSourceConfig test case that depends on the OperatorSource / CatalogSourceConfig created here to be present? I suggest we introduce this when we really have tests that we want to run in parallel.

test/e2e/marketplace_test.go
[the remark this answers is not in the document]
aravindhp: I prefer using a separate namespace for e2e tests and one for development. Not only that, we need to convert that namespace to have individual operator repositories like we are planning to have in the product. Though it wouldn't hurt to have one repository with two or more operators to ensure we don't break that functionality.

test/e2e/marketplace_test.go
```diff
@@ -0,0 +1,186 @@
...
+		return err
+	}
+
+	// Check that we created the catalogsourceconfig.
+	testCatalogSourceConfig := &operator.CatalogSourceConfig{}
+	err = WaitForResult(t, f, testCatalogSourceConfig, namespace, catalogSourceConfigName)
+	if err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
aravindhp: > Assert that the all expected package(s) have been downloaded from quay.io?
How do we do this? Do you want to make direct app registry calls to get the blobs and introspect the blobs and ensure they match? I would suggest against this as that code could potentially have bugs too. We could take the approach of having a separate appregistry namespace for e2e tests with a known set of operator repositories.

test/e2e/marketplace_test.go
```diff
@@ -0,0 +1,186 @@
...
+	// Then check for the catalog source.
+	err = WaitForResult(t, f, &olm.CatalogSource{}, namespace, catalogSourceName)
+	if err != nil {
+		return err
+	}
+
+	// Now lets update the catalogsourceconfig and see that the configmap and catalog source are updated
+	// First set the target namespace field for the catalogsourceconfig to a new namespace
```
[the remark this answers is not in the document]
kevinrizza: I see your point, but the framework automatically cleans up the artifacts of each of these tests. So we can create additional tests for each piece of functionality, but they will share a lot of common code (ie we will need to create an operator source that downloads from the quay and creates a catalogsourceconfig for each test)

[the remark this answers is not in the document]
aravindhp: Is there a way to ask for the clean up only after all tests are run?

test/e2e/marketplace_test.go
aravindhp: That happens only if the `targetNamespace` changes.

test/e2e/marketplace_test.go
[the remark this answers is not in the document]
aravindhp: I also think the update can be handled in a separate PR. That will involve ensuring that the `ConfigMap` and `CatalogSource` in the old namespace is deleted. With this PR, let us focus on testing the basic functionality and having it run automatically as part of CI for each PR.

test/e2e/marketplace_test.go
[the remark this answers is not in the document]
tkashem: I believe a free-tier user is allowed to have only one organization, I originally created `operators`. So  either @kevinrizza / @aravindhp creates an org for us or have someone from redhat create an org for us.

test/e2e/marketplace_test.go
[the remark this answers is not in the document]
tkashem: No, by inspecting the `Status.Packages` field of the `OperatorSource` object. ( My current PR adds this ), or inspecting the `CatalogSourceConfig` object  (Spec.Packages).  Yeah, I assumed it would be done in separate PR(s).

test/e2e/marketplace_test.go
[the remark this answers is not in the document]
tkashem: @kevinrizza I agree, and to add to it, here are some thoughts I have:
- We will have lots of common code, we can organize them into test helper functions to keep the test(s) concise.
- Ideally, each test (when it runs) should be independent of other test(s). This way we don't allow side effect from one test to bleed into another test. So one failing test should not have any impact on the other. I would recommend this to be the default approach and then break this rule on special cases if there is a need. 
- Every test validated one use case. So if a test fails we know what's broken.
- Every test has a setup and tear down, we will use the framework and our helper functions to reduce the boilerplate code.

test/e2e/marketplace_test.go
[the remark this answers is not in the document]
aravindhp: Sounds good. Let us restrict comments in this PR targeted towards basic functionality so that we can get this into the CI pipeline soon.

test/e2e/marketplace_test.go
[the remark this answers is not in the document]
aravindhp: Please use `marketplace_e2e`.  They have been populated with `federationv2 jboss-middleware mongodb-enterprise service-catalog couchbase-enterprise descheduler dynatrace-monitoring amq-streams etcd prometheus` operators.

kevinrizza:
@tkashem @aravindhp @SamiSousa I believe have addressed all of your comments, please re-review.

kevinrizza:
@aravindhp `go test` requires files with tests in them to be named *something*_test.go. I simplified the helper filename to helpers.go, but in order for the tests to run I left marketplace_test.go and main_test.go as is.

kevinrizza:
@aravindhp added a section to the readme in the latest update

scripts/e2e-tests.sh
```diff
@@ -0,0 +1,21 @@
...
+echo "Creating openshift resources"
+kubectl apply -f $OPERATOR_SOURCE_CRD
+kubectl apply -f $CATALOG_SOURCE_CONFIG_CRD
+if ! kubectl get namespace $TEST_NAMESPACE; then
+    kubectl create namespace $TEST_NAMESPACE
+fi
+
+# Run the tests through the operator-sdk
+echo "Running operator-sdk test"
+operator-sdk test local ./test/e2e --up-local --kubeconfig=$KUBECONFIG --namespace $TEST_NAMESPACE
```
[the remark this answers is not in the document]
kevinrizza: Yes, exactly that. That way, we can run the operator from source rather than relying on an image built from an external process. Without that tag the test relies on the image referenced in the operator.yaml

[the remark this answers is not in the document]
tkashem: Okay, so due to tools limitation we can't deploy marketplace operator on the cluster using the individual yaml files and then run the `e2e` tests on it. 

(In the future) I believe e2e tests should run on an environment provisioned from the deploy manifests.  For example, (thinking out loud here) we could templatize `namespace` and `image` referred to by individual yaml files. We build the image, generate the yaml files from templates, deploy and run tests.I am sure there are more elegant solutions we can come up with in the future. It would be great if we track this issue in jira.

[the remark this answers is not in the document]
aravindhp: A lot of what you are asking for is done by the CI operator once Marketplace is part of the cluster payload image-stream and will run by default in a cluster.

PR #46, opened 2018-12-06
Ensure datastore cache is in sync when marketplace operator restarts

tkashem:
When marketplace operator process dies it loses in-memory cache for already reconciled OperatorSource object(s). We need to rebuild the datastore cache for all existing OperatorSource object(s) when the process starts.

* When the operator process starts it is informed of all existing OperatorSource object(s). We use it to rebuild datastore cache.
* Implement logic to detect  whether an OperatorSource object is out of sync.
* Introduce a new phase called "Purging" and a reconciler associated with it. Purge all resource(s) created during the course of reconciliation when an OperatorSource object is purged. 
* For out of sync OperatorSource object(s) kick off reconciliation from Purging phase.


pkg/datastore/datastore.go
```diff
@@ -45,11 +45,18 @@ type Reader interface {
 // Writer is an interface that is used to manage the underlying datastore
 // for operator manifest.
 type Writer interface {
-	// GetPackageIDs returns a comma separated list of operator ID(s). Each ID
+	// GetPackageIDs returns a comma separated list of operator ID(s). This list
+	// includes operator(s) across all OperatorSource object(s). Each ID
 	// returned can be used to retrieve the manifest associated with the
 	// operator from underlying datastore.
 	GetPackageIDs() string
```
galletti94: why not make a list of strings instead?

pkg/operatorsource/handler.go
```diff
@@ -59,26 +59,47 @@ func (h *operatorsourcehandler) Handle(ctx context.Context, in *v1alpha1.Operato
...
 	})
 
-	reconciler, err := h.factory.GetPhaseReconciler(logger, in)
+	prePhaseReconciler, err := h.factory.GetPrePhaseReconciler(logger, in)
 	if err != nil {
 		return err
 	}
 
-	out, status, err := reconciler.Reconcile(ctx, in)
+	out, status, err := prePhaseReconciler.Reconcile(ctx, in)
```
galletti94: potentially re-declaring of err variable here no?

pkg/operatorsource/handler.go
```diff
@@ -59,26 +59,47 @@ func (h *operatorsourcehandler) Handle(ctx context.Context, in *v1alpha1.Operato
...
 
+	// The pre-phase reconciler has handled the event. So we should transition
+	// into the new phase and return.
+	if status != nil {
+		return h.transition(ctx, logger, out, status, err)
+	}
+
+	// The pre-phase reconciler has not handled the event, so let's use the
+	// regular phase reconciler to handle this event.
+	phaseReconciler, err := h.factory.GetPhaseReconciler(logger, in)
```
galletti94: potentially re-declaring of err variable here no?

pkg/datastore/datastore.go
tkashem: Good question, right now `CatalogSourceConfig.Spec.packages` is a string type. We need to change it to `[]string`. Could be a candidate for 4.1, I believe.

pkg/operatorsource/handler.go
tkashem: `a short variable declaration may redeclare variables provided they were originally declared earlier in the same block. Redeclaration does not introduce a new variable; it just assigns a new value to the original.`

More about it - https://golang.org/ref/spec#Short_variable_declarations

pkg/operatorsource/prephase.go
```diff
@@ -0,0 +1,85 @@
...
+// NewPrePhaseReconciler returns a pre-phase Reconciler.
+func NewPrePhaseReconciler(logger *log.Entry, datastore datastore.Writer, client client.Client) Reconciler {
+	return &prePhaseReconciler{
+		logger:    logger,
+		datastore: datastore,
+		client:    client,
+	}
+}
+
+// prePhaseReconciler implements Reconciler interface.
```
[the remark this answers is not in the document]
tkashem: Does `OutOfSyncReconciler` make it better, since it detects if an `OperatorSource` is not in sync any longer? Feel free to suggest another name based on its primary role. :)

pkg/operatorsource/purging_test.go
```diff
@@ -51,27 +50,26 @@ func TestReconcile_SpecHasChanged_ReconciliationTriggered(t *testing.T) {
...
 }
 
-// Use Case: The associated CatalogSourceConfig object is not found while purging.
-// Expected Result: NotFound error is ignored and the next phase should be set
-// to "Validating" so that reconciliation is triggered.
-func TestReconcile_CatalogSourceConfigNotFound_ErrorExpected(t *testing.T) {
+// In the event the associated CatalogSourceConfig object is not found while
+// purging is in progress, we expect NotFound error to be ignored and the next
+// phase set to "Initial" so that reconciliation can start anew.
+func TestReconcileWithPurgingWithCatalogSourceConfigNotFound(t *testing.T) {
```
SamiSousa: Can we establish a naming convention for tests? Here tests are in pure CamelCase, but in other tests we use a mix of CamelCase and underscores.

pkg/operatorsource/configuring.go
```diff
@@ -58,7 +58,7 @@ type configuringReconciler struct {
 // If the corresponding CatalogSourceConfig object already exists
 // then no further action is taken.
 func (r *configuringReconciler) Reconcile(ctx context.Context, in *v1alpha1.OperatorSource) (out *v1alpha1.OperatorSource, nextPhase *v1alpha1.Phase, err error) {
-	if in.Status.CurrentPhase.Name != phase.Configuring {
+	if in.GetCurrentPhaseName() != phase.Configuring {
```
aravindhp: You should introduce GetCurrentPhaseName() in this commit. So this commit will come before `[opsrc] Introduce a new phase: "Purging"`.

pkg/apis/marketplace/v1alpha1/operatorsource_types.go
```diff
@@ -50,6 +50,11 @@ type OperatorSourceSpec struct {
 type OperatorSourceStatus struct {
 	// Current phase of the OperatorSource object
 	CurrentPhase ObjectPhase `json:"currentPhase,omitempty"`
+
+	// Packages is a comma separated list of package(s) each of which has been
+	// downloaded and processed by Marketplace operator from the specified
+	// endpoint.
+	Packages string `json:"packages,omitempty"`
```
aravindhp: Is this required to keep the datastore cache in sync? If not, this warrants its own PR as it is introducing a new status field. If it does then it needs to be explained in the commit message as to how it is connected. 

My suggestion is that this becomes a separate PR so that it is clearly called out. You can cherry pick this into another branch and drop it from this PR.

tkashem:
@aravindhp @kevinrizza @galletti94 addressed the comments, please review.

PR #47, opened 2018-12-07
Update upstream Dockerfile

kevinrizza:
Root Dockerfile currently uses ocp builder image
Use openshift/release image instead


awgreene:
/lgtm

Thanks for the quick response on this item.

PR #48, opened 2018-12-07
[deploy] Add liveness and readiness probes

awgreene:
Updated main.go to include an http endpoint that returns a 200 status when pinged for use in readiness and liveness probes.
Updated the operator.yaml to include a readiness and liveness probe.

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -19,9 +19,19 @@ spec:
           ports:
           - containerPort: 60000
             name: metrics
+          - containerPort: 8080
+            name: healthz
           command:
           - marketplace-operator
           imagePullPolicy: Always
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: 8080
+          readinessProbe:
+            httpGet:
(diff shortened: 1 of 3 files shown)


deploy/operator.yaml
```diff
@@ -19,9 +19,19 @@ spec:
           ports:
           - containerPort: 60000
             name: metrics
+          - containerPort: 8080
+            name: healthz
           command:
           - marketplace-operator
           imagePullPolicy: Always
+          livenessProbe:
```
aravindhp: Please add this to the CSV also.

awgreene: Added to the CSV.

awgreene:
Relies on changes introduced in PR #49

kevinrizza:
Probably outside the scope of this pr, but I think we should start to consider doing something more than have a single long main() method. Right now we have framework logic around creating the manager and the controllers in this method, which seems somewhat orthogonal to adding an http endpoint to the process.

/lgtm

PR #49, opened 2018-12-07
[bug] Update csv to use the expected service account and namespace

awgreene:
This bug came as a result of the changes that were introduced when the project was rebuilt using version 0.1.1 of the operator sdk.
In this update, the service account changed from default to marketplace-operator. The csv has been updated to use the marketplace-operator service account.

### DIFF
--- a/deploy/marketplace.csv.yaml
+++ b/deploy/marketplace.csv.yaml
@@ -2,7 +2,7 @@ apiVersion: operators.coreos.com/v1alpha1
 kind: ClusterServiceVersion
 metadata:
   name: marketplace-operator.v0.0.1
-  namespace: marketplace
+  namespace: openshift-marketplace
 spec:
   displayName: marketplace
   description: |-
@@ -27,7 +27,7 @@ spec:
     strategy: deployment
     spec:
       clusterPermissions:
-      - serviceAccountName: default
+      - serviceAccountName: marketplace-operator
         rules:
         - apiGroups:
           - marketplace.redhat.com
@@ -61,6 +61,7 @@ spec:
               labels:
                 name: marketplace-operator
             spec:
+              serviceAccountName: marketplace-operator
               containers:
                 - name: marketplace-operator
                   image: quay.io/openshift/origin-operator-marketplace


deploy/role_binding.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 subjects:
 - kind: ServiceAccount
   name: marketplace-operator
-  namespace: openshift-marketplace
+  namespace: marketplace
```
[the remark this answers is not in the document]
awgreene: How does this introduce a bug? I recognize that this should be done in a separate PR but I don’t see the bug. Are we supposed to deploy to the openshift-marketplace namespace? I only ask because oc fails when running oc new-project openshift-marketplace.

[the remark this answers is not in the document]
aravindhp: Because you now make it inconsistent with some of the other yamls like the OperatorSource example that QE is using. We need to create `openshift-marketplace` using the installer for 4.0. In the interim, we can use `kubectl` to create that namespace.

deploy/marketplace.csv.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: operators.coreos.com/v1alpha1
 kind: ClusterServiceVersion
 metadata:
   name: marketplace-operator.v0.0.1
-  namespace: marketplace
+  namespace: openshift-marketplace
```
tkashem: Should this still be in the same PR?

awgreene: Good question, the role_binding is created in `openshift-marketplace` whereas the csv was created in `marketplace`. If this isn't updated, the marketplace operator won't be deployed because of unmet dependencies.

awgreene: What are your thoughts @tkashem

aravindhp: Your commit header and message should then reflect this.

awgreene:
https://github.com/operator-framework/operator-marketplace/pull/49#discussion_r240252016

@aravindhp Commit header and body have been updated.

PR #50, opened 2018-12-10
[deploy] Add namespace yaml

aravindhp:
- Add a namespace yaml for creating the `openshift-marketplace` namespace
- Update the README's deploying section to reflect this

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,8 @@ It is important to note that the order in which you apply the deployment files m
 
 #### Deploying the Marketplace Operator
 ```bash
+$ oc apply -f deploy/marketplace.ns.yaml
+$ oc project openshift-marketplace
 $ oc apply -f deploy/catalogsourceconfig.crd.yaml
 $ oc apply -f deploy/operatorsource.crd.yaml
 $ oc apply -f deploy/service_account.yaml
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -33,6 +33,15 @@ The operator will then create a CatalogSourceConfig CR which will, for the time
 ### Deploying the Marketplace Operator with OKD
 It is important to note that the order in which you apply the deployment files matters, do not execute the `oc apply` commands featured in this section out of order.
 
+#### Creating the Marketplace Namespace
+Create the `openshift-marketplace` namespace using `kubectl` as `oc` will not allow a user to create namespaces with the 
+`openshift` prefix. This is a stopgap measure for OKD until Marketplace is installed by default in a cluster. For
+upstream Kubernetes this step will always be required.
+
+```bash
+kubectl create -f deploy/marketplace.ns.yaml
```
awgreene: ~~Is there a reason we you've opted to use `kubectl` over `oc` to create the namespace here? The rest of the readme has been using the `oc` command.~~

awgreene: This isn't entirely correct - the `oc` command can be used here since you are creating a namespace and not a project. So while `oc new-project marketplace-operator` will fail `oc apply -f deploy/marketplace.ns.yaml` will not.

aravindhp: Good point. I have fixed it. Please review.

PR #51, opened 2018-12-10
Fix unit test job and pin golang/mock to the latest release

tkashem:
Right now we 'go get' the latest version of golang/mock to build mockgen binary. On the other hand, The gomock package used in test is imported from the vendor folder. These two sources of golang/mock
can easily be out of sync. This is exactly why the unit test job in CI is failing.

* Ensure `mockgen` is the same version as `gomock`.
* Pin `golang/mock` to the latest release `v1.2.0`


tkashem:
@awgreene It's primarily because I could have one without the other. The first commit is to make sure both mockgen and gomock are tied to the same version and the other upgrades to a new version. The upgrade ideally should have been a separate PR because you could make the argument that fixing the unit test job does not require an upgrade.

PR #54, opened 2018-12-11
Updated OperaorSource unit tests using fake client

SamiSousa:
Use the `controller-runtime`'s fake client in unit tests for `operatorsource` package.

`operator-sdk` should include an up-to-date fake client with custom scheme support built-in by this week.
I've applied the minimum changes needed to get fake client to work with Marketplace's custom CRDs in the meantime.

**UPDATE**
With PR #69 merged, fake client should be good to go.

### DIFF
--- a/pkg/operatorsource/configuring_test.go
+++ b/pkg/operatorsource/configuring_test.go
@@ -5,16 +5,17 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	gomock "github.com/golang/mock/gomock"
+	"github.com/operator-framework/operator-marketplace/pkg/apis"
 	"github.com/operator-framework/operator-marketplace/pkg/apis/marketplace/v1alpha1"
 	mocks "github.com/operator-framework/operator-marketplace/pkg/mocks/operatorsource_mocks"
 	"github.com/operator-framework/operator-marketplace/pkg/operatorsource"
 	"github.com/operator-framework/operator-marketplace/pkg/phase"
-	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
 // Use Case: Not configured, CatalogSourceConfig object has not been created yet.
@@ -24,18 +25,22 @@ func TestReconcile_NotConfigured_NewCatalogConfigSourceObjectCreated(t *testing.
 	controller := gomock.NewController(t)
 	defer controller.Finish()
 
+	namespace, name := "marketplace", "foo"
 	nextPhaseWant := &v1alpha1.Phase{
 		Name:    phase.Succeeded,
 		Message: phase.GetMessage(phase.Succeeded),
 	}
 
-	datastore := mocks.NewDatastoreWriter(controller)
-	kubeclient := mocks.NewKubeClient(controller)
+	scheme := runtime.NewScheme()
+	apis.AddToScheme(scheme)
 
-	reconciler := operatorsource.NewConfiguringReconciler(helperGetContextLogger(), datastore, kubeclient)
+	datastore := mocks.NewDatastoreWriter(controller)
+	fakeclient := fake.NewFakeClientWithScheme(scheme)
+
+	reconciler := operatorsource.NewConfiguringReconciler(helperGetContextLogger(), datastore, fakeclient)
 
 	ctx := context.TODO()
-	opsrcIn := helperNewOperatorSourceWithPhase("marketplace", "foo", phase.Configuring)
+	opsrcIn := helperNewOperatorSourceWithPhase(namespace, name, phase.Configuring)
 
 	labelsWant := map[string]string{
 		"opsrc-group": "Community",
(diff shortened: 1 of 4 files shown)


pkg/operatorsource/configuring_test.go
```diff
@@ -34,19 +35,20 @@ func TestReconcile_NotConfigured_NewCatalogConfigSourceObjectCreated(t *testing.
...
+	reconciler := operatorsource.NewConfiguringReconciler(helperGetContextLogger(), datastore, fakeclient)
 
 	ctx := context.TODO()
 	opsrcIn := helperNewOperatorSourceWithPhase("marketplace", "foo", phase.Configuring)
 	namespacedName := types.NamespacedName{Name: "opsrc-foo", Namespace: "marketplace"}
 
 	// We expect that the given CatalogConfigSource object does not exist.
 	cscGet := helperNewCatalogSourceConfig(opsrcIn.Namespace, getExpectedCatalogSourceConfigName(opsrcIn.Name))
-	kubeClientErr := k8s_errors.NewNotFound(schema.GroupResource{}, "CatalogSourceConfig not found")
-	kubeclient.EXPECT().Get(context.TODO(), namespacedName, cscGet).Return(kubeClientErr)
```
tkashem: Question: Because the fresh client at this point does not have any record in cache it will return a `NotFound` error for any `get` call. There is no way to ensure that the function being tested passes the expected `namespace`, `name` and `metadata` to the `client.get` function, correct?

pkg/operatorsource/configuring_test.go
```diff
@@ -66,13 +68,15 @@ func TestReconcile_NotConfigured_NewCatalogConfigSourceObjectCreated(t *testing.
...
 	}
-	kubeclient.EXPECT().Create(context.TODO(), cscWant).Return(nil)
 
 	opsrcGot, nextPhaseGot, errGot := reconciler.Reconcile(ctx, opsrcIn)
 
 	assert.NoError(t, errGot)
 	assert.Equal(t, opsrcIn, opsrcGot)
 	assert.Equal(t, nextPhaseWant, nextPhaseGot)
+
+	errClient := fakeclient.Get(ctx, namespacedName, cscWant)
```
tkashem: Here we want to verify that the function being tested passes the expected `CatalogSourceConfig` object to the `fakeClient`. I suggest the following:
```
cscGot := &v1alpha1.CatalogSourceConfig{}
errClient := fakeclient.Get(ctx, namespacedName, cscGot)
assert.NoError(t, errClient)
assert.Equal(t, cscWant, cscGot)
```
This will ensure that the expected value of `CatalogSourceConfig` was passed to `fakeClient.Create`

pkg/operatorsource/configuring_test.go
```diff
@@ -34,19 +35,20 @@ func TestReconcile_NotConfigured_NewCatalogConfigSourceObjectCreated(t *testing.
...
 	opsrcIn := helperNewOperatorSourceWithPhase("marketplace", "foo", phase.Configuring)
 	namespacedName := types.NamespacedName{Name: "opsrc-foo", Namespace: "marketplace"}
 
 	// We expect that the given CatalogConfigSource object does not exist.
 	cscGet := helperNewCatalogSourceConfig(opsrcIn.Namespace, getExpectedCatalogSourceConfigName(opsrcIn.Name))
-	kubeClientErr := k8s_errors.NewNotFound(schema.GroupResource{}, "CatalogSourceConfig not found")
-	kubeclient.EXPECT().Get(context.TODO(), namespacedName, cscGet).Return(kubeClientErr)
 
 	packages := "a,b,c"
 	datastore.EXPECT().GetPackageIDs().Return(packages)
```
tkashem: The mix of both `fakeClient` and `gomocked` interfaces ( like `datastore` above ) in the same test introduces inconsistency.

pkg/operatorsource/configuring_test.go
```diff
@@ -85,19 +89,19 @@ func TestReconcile_AlreadyConfigured_NoActionTaken(t *testing.T) {
...
+	marketplace.AddToScheme(scheme)
 
 	datastore := mocks.NewDatastoreWriter(controller)
-	kubeclient := mocks.NewKubeClient(controller)
-
-	reconciler := operatorsource.NewConfiguringReconciler(helperGetContextLogger(), datastore, kubeclient)
 
-	ctx := context.TODO()
+	// We expect that the given CatalogConfigSource object already exists.
 	opsrcIn := helperNewOperatorSourceWithPhase("marketplace", "foo", phase.Configuring)
```
tkashem: If we expect the given `CatalogConfigSource` object to already exist then we should call `fakeClient.Create` before the function being tested is called, right?

SamiSousa: You can also initialize the fake client with runtime objects, which I have done on line 100.

pkg/operatorsource/configuring_test.go
SamiSousa: So the only change here is to add an `assert.Equal` and move the `csc` declaration closer to the `fakeclient` call, correct?

pkg/operatorsource/configuring_test.go
SamiSousa: I think the best approach would be to have all `gomocked` interface calls in one section and all `fakeClient` calls in another. Based on how we're using `gomocked` interfaces and how others have used `fakeClient` in tests, I suggest that all `gomocked` tests should be called before the actual function being tested, and all `fakeClient` calls should be after.

pkg/operatorsource/configuring_test.go
SamiSousa: Correct. I think that the `fakeClient` can only test that the client has the expected state after the function is called.

pkg/operatorsource/configuring_test.go
SamiSousa: I've addressed these changes following your suggestion.

pkg/operatorsource/configuring_test.go
```diff
@@ -6,16 +6,17 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	gomock "github.com/golang/mock/gomock"
+	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis"
```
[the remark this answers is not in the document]
SamiSousa: Good point, I'll remove the alias. Thanks!

SamiSousa:
@kevinrizza @tkashem Addressed your comments. The test changes are in the second commit, pertaining to the `operatorsource_test` package.

kevinrizza:
@SamiSousa You will probably want to wait until your sdk upgrade pr is merged, then rebase this branch so that there aren't any conflicts (if anything changes on that pr) and this pr is easier to read.

SamiSousa:
@tkashem @kevinrizza I've pushed the fake client tests for the `operatorsource` package. My plan is to follow up with tests for the other packages in marketplace, namely `catalogsourceconfig`. Let me know what you think. :)

kevinrizza:
@SamiSousa I can take a look at what's here now, but think that CatalogSourceConfig unit tests can be a separate pr.

SamiSousa:
Removed WIP label as per @kevinrizza's comment. Unit tests for `catalogsourceconfig` package will be created in a separate PR.

SamiSousa:
@aravindhp Good point, since we're no longer using mocks for the kubeclient I'll update `generate-mocks` to reflect this.

pkg/operatorsource/handler_test.go
```diff
@@ -32,13 +39,16 @@ func TestHandle_PhaseHasChanged_UpdateExpected(t *testing.T) {
...
-	handler := operatorsource.NewHandlerWithParams(kubeclient, writer, factory, transitioner, newCacheReconcilerFunc)
+	handler := operatorsource.NewHandlerWithParams(fakeclient, writer, factory, transitioner, newCacheReconcilerFunc)
 
 	ctx := context.TODO()
 
 	// Making two OperatorSource objects that are not equal to simulate a change.
 	opsrcIn, opsrcOut := helperNewOperatorSourceWithEndpoint("marketplace", "foo", "remote"), helperNewOperatorSourceWithEndpoint("marketplace", "foo", "local")
 
+	// Add OperatorSource to fakeclient
+	assert.NoError(t, fakeclient.Create(ctx, opsrcIn));
```
kevinrizza: Why do an assert here? This isn't actually testing our code, right?

pkg/operatorsource/handler_test.go
```diff
@@ -55,12 +65,16 @@ func TestHandle_PhaseHasChanged_UpdateExpected(t *testing.T) {
...
-
 	errGot := handler.Handle(ctx, opsrcIn)
 
 	assert.NoError(t, errGot)
+
+	// We expect the object to be updated successfully.
+	namespacedName := types.NamespacedName{Name: "foo", Namespace: "marketplace"}
+	opsrcGot := &v1alpha1.OperatorSource{}
+
+	assert.NoError(t, fakeclient.Get(ctx, namespacedName, opsrcGot))
```
kevinrizza: Same question as above. What's the point of asserting there isn't an error for fakeclient.Get()? Isn't that implied in the next assertion? What you are trying to say is "The operatorsource we are looking for really does exist", but I think we should only assert about errors on our own code. If this throws an error, does that necessarily mean that there was nothing found for that `namespacedName`?

pkg/operatorsource/handler_test.go
```diff
@@ -102,6 +122,13 @@ func TestHandle_PhaseHasNotChanged_NoUpdateExpected(t *testing.T) {
 	errGot := handler.Handle(ctx, opsrcIn)
 
 	assert.NoError(t, errGot)
+
+	// We expect no changes to the object
+	namespacedName := types.NamespacedName{Name: "foo", Namespace: "namespace"}
+	opsrcGot := &v1alpha1.OperatorSource{}
+
+	assert.NoError(t, fakeclient.Get(ctx, namespacedName, opsrcGot))
```
kevinrizza: same comment as above.

pkg/operatorsource/handler_test.go
SamiSousa: That's true. `fakeclient.Create` returns an error object, so I figured I might as well. Since it shouldn't throw an error and it's part of the test setup, I can call `Create` without the assert.

pkg/operatorsource/handler_test.go
SamiSousa: That's an interesting point. So what you're saying is, if `fakeclient.Get` does throw an error for some reason but `opsrcGot` matches `opsrcOut` then the test still passes since we're testing `handler` not the fake client. Is that right?

SamiSousa:
@kevinrizza Addressed your comments. I removed the asserts for calls to the fake client since they aren't part of the unit test.

pkg/operatorsource/configuring_test.go
```diff
@@ -122,48 +129,18 @@ func TestReconcile_CatalogSourceConfigAlreadyExists_Updated(t *testing.T) {
...
 	opsrcGot, nextPhaseGot, errGot := reconciler.Reconcile(ctx, opsrcIn)
 
 	assert.NoError(t, errGot)
 	assert.Equal(t, opsrcIn, opsrcGot)
 	assert.Equal(t, nextPhaseWant, nextPhaseGot)
-}
 
-// Use Case: Update of existing CatalogSourceConfig object fails.
-// Expected Result: The object is moved to "Failed" phase.
-func TestReconcile_UpdateError_MovedToFailedPhase(t *testing.T) {
```
SamiSousa: @tkashem I have looked at this test again, and I don't see much added value by having it. The unique code path is within an error check that sets the `nextPhase` to `Failed` while other error checks aren't given their own tests. Let me know if you have any objections.

SamiSousa:
Closing in favor of #147

PR #55, opened 2018-12-11
[opsrc] Add downloaded manifest list to Status

tkashem:
Add a new field called 'Packages' to OperatorSourceStatus. This is a
comma separated list of operator manifest(s) downloaded by
marketplace operator from the registry specified by the OperatorSource
object.

When an OperatorSource object is reconciled this field allows us to see
the list of manifest(s) processed.

### DIFF
--- a/pkg/apis/marketplace/v1alpha1/operatorsource_types.go
+++ b/pkg/apis/marketplace/v1alpha1/operatorsource_types.go
@@ -20,6 +20,7 @@ type OperatorSourceList struct {
 	Items           []OperatorSource `json:"items"`
 }
 
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // OperatorSource is the Schema for the operatorsources API
 // +k8s:openapi-gen=true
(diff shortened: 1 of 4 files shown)


pkg/apis/marketplace/v1alpha1/operatorsource_types.go
```diff
@@ -20,6 +20,8 @@ type OperatorSourceList struct {
 	Items           []OperatorSource `json:"items"`
 }
 
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
```
aravindhp: Did you actually mean to move this?

tkashem: Yes, I had to move it. Otherwise if I do `operator-sdk generate k8s` it removes `deepcopy` functions.

tkashem:
@kevinrizza @aravindhp Can you thumbs up. I force pushed thinking I don't need to move the `k8s:deepcopy-gen` only to realize later that I do.

PR #59, opened 2018-12-12
[csc+opsrc] Introduce datastore label

aravindhp:
- Update the CatalogSourceConfigBuilder` to add a datastore label.
- The CatalogSourceConfig's configuring reconciler will check for this datastore label. If present it will add the _olm-visibility: hidden_ and _openshift-marketplace: true_ labels to the CatalogSource. 
_olm-visibility: hidden_ will cause the CatalogSource to be hidden in the OLM Package Manifest UI. 
_openshift-marketplace: true_ will be used to filter out the global CatalogSources in the Marketplace UI.
- Add the datastore label to the ObjectMeta in helperNewCatalogSourceConfig()

### DIFF
--- a/pkg/catalogsourceconfig/catalogsourcebuilder.go
+++ b/pkg/catalogsourceconfig/catalogsourcebuilder.go
@@ -28,10 +28,23 @@ func (b *CatalogSourceBuilder) WithTypeMeta() *CatalogSourceBuilder {
 // WithMeta sets basic TypeMeta and ObjectMeta.
 func (b *CatalogSourceBuilder) WithMeta(name, namespace string) *CatalogSourceBuilder {
 	b.WithTypeMeta()
-	b.cs.ObjectMeta = metav1.ObjectMeta{
-		Name:      name,
-		Namespace: namespace,
+	objectMeta := b.cs.GetObjectMeta()
+	if objectMeta == nil {
+		b.cs.ObjectMeta = metav1.ObjectMeta{}
 	}
+	b.cs.SetName(name)
+	b.cs.SetNamespace(namespace)
+	return b
+}
+
+// WithOLMLabels adds "olm-visibility" and "openshift-marketplace" labels to ObjectMeta.
+func (b *CatalogSourceBuilder) WithOLMLabels() *CatalogSourceBuilder {
+	b.WithTypeMeta()
+	objectMeta := b.cs.GetObjectMeta()
+	if objectMeta == nil {
+		b.cs.ObjectMeta = metav1.ObjectMeta{}
+	}
+	b.cs.SetLabels(map[string]string{"olm-visibility": "hidden", "openshift-marketplace": "true"})
 	return b
 }
 
--- a/pkg/catalogsourceconfig/configuring.go
+++ b/pkg/catalogsourceconfig/configuring.go
@@ -3,6 +3,7 @@ package catalogsourceconfig
 import (
 	"context"
 	"fmt"
+	"github.com/operator-framework/operator-marketplace/pkg/operatorsource"
 	"strings"
 
 	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
@@ -204,10 +205,21 @@ func newConfigMap(csc *v1alpha1.CatalogSourceConfig, data map[string]string) *co
 
 // newCatalogSource returns a CatalogSource object.
 func newCatalogSource(csc *v1alpha1.CatalogSourceConfig, configMapName string) *olm.CatalogSource {
-	return new(CatalogSourceBuilder).
-		WithMeta(v1alpha1.CatalogSourcePrefix+csc.Name, csc.Spec.TargetNamespace).
+	builder := new(CatalogSourceBuilder).
 		WithOwner(csc).
+		WithMeta(v1alpha1.CatalogSourcePrefix+csc.Name, csc.Spec.TargetNamespace).
 		// TBD: where do we get display name and publisher from?
-		WithSpec("internal", configMapName, csc.Name, csc.Name).
-		CatalogSource()
+		WithSpec("internal", configMapName, csc.Name, csc.Name)
+
+	// Check if the operatorsource.DatastoreLabel is "true" which indicates that
+	// the CatalogSource is the datastore for an OperatorSource. This is a hint
+	// for us to set the "olm-visibility" label in the CatalogSource so that it
+	// is not visible in the OLM Packages UI. In addition we will set the
+	// "openshift-marketplace" label which will be used by the Marketplace UI
+	// to filter out global CatalogSources.
+	datastoreLabel, found := csc.ObjectMeta.GetLabels()[operatorsource.DatastoreLabel]
+	if found && strings.ToLower(datastoreLabel) == "true" {
+		builder.WithOLMLabels()
+	}
+	return builder.CatalogSource()
 }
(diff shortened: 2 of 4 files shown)


pkg/catalogsourceconfig/catalogsourcebuilder.go
```diff
@@ -28,13 +28,29 @@ func (b *CatalogSourceBuilder) WithTypeMeta() *CatalogSourceBuilder {
...
 	}
 	return b
 }
 
+// WithVisibilityMeta sets basic TypeMeta and ObjectMeta with the "olm-visibility" label.
+func (b *CatalogSourceBuilder) WithVisibilityMeta(name, namespace string) *CatalogSourceBuilder {
+	b.WithTypeMeta()
+	b.WithMeta(name, namespace)
+	// Hint for OLM Packages UI to hide this CatalogSource
+	b.cs.ObjectMeta.Labels = map[string]string{"olm-visibility": "hidden"}
```
SamiSousa: Is this separate from the Datastore label? I thought they were the same until I saw the label string was different.

pkg/catalogsourceconfig/configuring.go
```diff
@@ -204,10 +205,21 @@ func newConfigMap(csc *v1alpha1.CatalogSourceConfig, data map[string]string) *co
...
-		WithSpec("internal", configMapName, csc.Name, csc.Name).
-		CatalogSource()
+		WithSpec("internal", configMapName, csc.Name, csc.Name)
+
+	// Check if the operatorsource.DatastoreLabel is "true" which indicates that
+	// the CatalogSource is the datastore for an OperatorSource. This is a hint
+	// for us to set the OLM visibility label in the CatalogSource so that it is
+	// not visible in the OLM Packages UI.
+	labels := csc.ObjectMeta.GetLabels()
+	datastore, found := labels[operatorsource.DatastoreLabel]
```
SamiSousa: Where is the Datastore label being set? Is this going to be added to our example `OperatorSource` CR?

pkg/catalogsourceconfig/catalogsourcebuilder.go
aravindhp: Yes this is different from the datastore label.

pkg/catalogsourceconfig/configuring.go
aravindhp: No, this will be added in the `OperatorSource` `Configuring` reconciler.

aravindhp:
> I think the difference between the Datastore label and the olm-visibility label is unclear.
Does [this comment](https://github.com/operator-framework/operator-marketplace/pull/59/files/77abb2c658f61b177186865dea0096d9dbe9cc51#diff-b19a565d7e89eb9ee947f1ee06d23d06R213) not explain it well enough? Let me know and I will try to reword.

SamiSousa:
@aravindhp If I understand the comment, the `operatorsource.DatastoreLabel` is the name of the label in the `CatalogSourceConfig` object. If that's true, then it's a bit confusing that the label is defined in the `operatorsource` package rather than the `catalogsourceconfig` package.

aravindhp:
>  If I understand the comment, the operatorsource.DatastoreLabel is the name of the label in the CatalogSourceConfig object.

That is why I added it to the `CatalogSourceConfigBuilder` in `catalogsourceconfigbuilder.go`. If others find it confusing too then I am open to moving it to where it makes more sense.

kevinrizza:
My preference is to leave `operatorsource.DatastoreLabel` where it is as the label is going to be applied when building a catalogsourceconfig.

awgreene:
I agree with @kevinrizza on this one.

pkg/catalogsourceconfig/catalogsourcebuilder.go
```diff
@@ -28,13 +28,29 @@ func (b *CatalogSourceBuilder) WithTypeMeta() *CatalogSourceBuilder {
...
+		b.cs.SetName(name)
+		b.cs.SetNamespace(namespace)
 	}
 	return b
 }
 
+// WithVisibilityMeta sets basic TypeMeta and ObjectMeta with the "olm-visibility" label.
+func (b *CatalogSourceBuilder) WithVisibilityMeta(name, namespace string) *CatalogSourceBuilder {
+	b.WithTypeMeta()
+	b.WithMeta(name, namespace)
```
kevinrizza: Is it preferable to define this function call here like this? Wouldn't it be better to do something similar to what's done in the `WithMeta()` method? i.e.

`if objectMeta == nil {
  b.cs.ObjectMeta{
    Labels: map[string]string{"olm-visibility": "hidden"}
  }
} else {
  b.cs.ObjectMeta.Labels = map[string]string{"olm-visibility": "hidden"}
}`

I know that the reason you are doing this here is that you are right now guaranteeing that some fields will be set, but if those methods are ever changed then it may be better to only modify the fields that the function is described to do.

pkg/catalogsourceconfig/catalogsourcebuilder.go
```diff
@@ -28,13 +28,29 @@ func (b *CatalogSourceBuilder) WithTypeMeta() *CatalogSourceBuilder {
...
-	b.cs.ObjectMeta = metav1.ObjectMeta{
-		Name:      name,
-		Namespace: namespace,
+	objectMeta := b.cs.GetObjectMeta()
+	if objectMeta == nil {
+		b.cs.ObjectMeta = metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		}
+	} else {
```
awgreene: :+1:

pkg/catalogsourceconfig/catalogsourcebuilder.go
aravindhp: So you want me to remove the `b.WithMeta()` call and only define this function as `WithVisibilityLabel()`? I am fine doing that if that is what you are asking. @tkashem are you in agreement too?

kevinrizza: Yeah, that is what I mean. I think it makes sense that the function only does one thing.

tkashem: I think that's a good idea to add `WithVisibilityLabel()` that adds the label.

pkg/catalogsourceconfig/catalogsourcebuilder.go
```diff
@@ -28,13 +28,29 @@ func (b *CatalogSourceBuilder) WithTypeMeta() *CatalogSourceBuilder {
 // WithMeta sets basic TypeMeta and ObjectMeta.
 func (b *CatalogSourceBuilder) WithMeta(name, namespace string) *CatalogSourceBuilder {
 	b.WithTypeMeta()
-	b.cs.ObjectMeta = metav1.ObjectMeta{
-		Name:      name,
-		Namespace: namespace,
+	objectMeta := b.cs.GetObjectMeta()
+	if objectMeta == nil {
```
tkashem: I am not clear why you need to check for nil though, or may be this be removed once you addressed @kevinrizza's comments?

aravindhp: Because if the object meta was already initialized using the `WithMeta()` call, I don't want to overwrite or lose any information. This check will be moved to `WithVisibilityLabel()`.

tkashem: nit: maybe we can make it a bit more concise?
```
if objectMeta == nil {
  b.cs.ObjectMeta = metav1.ObjectMeta{}
}

b.cs.SetName(name)
b.cs.SetNamespace(namespace)
```

aravindhp:
@kevinrizza @tkashem I have made the changes you requested. Please review.

PR #60, opened 2018-12-12
Update .gitignore to include Jetbrain IDEs

aravindhp:
Update .gitignore based on output from https://www.gitignore.io/api/go,vim,emacs,visualstudiocode,jetbrains

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,8 @@ pkg/mocks
 # Generated yaml files for operator-sdk test framework
 deploy/test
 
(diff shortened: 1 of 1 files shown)


awgreene:
@aravindhp Why does the `gitignore` file start with a newline?

aravindhp:
> Why does the gitignore file start with a newline?

`.gitignore` was seeded by the SDK, so I am guessing it added the newline. Do you want me to remove it?

awgreene:
> .gitignore was seeded by the SDK, so I am guessing it added the newline. Do you want me to remove it?
@aravindhp No need, I was just curious if it was by design.

PR #61, opened 2018-12-13
[WIP] [docs] Creating operator artifacts from source

SamiSousa:
- Adds operator resources organized by operator
- Provides helm templates for creating operator-artifacts
- Includes brief documentation for creating additional artifacts

Since we're creating operator artifacts for individual operators, I've reworked some of the helm templates that OLM uses to help automate that process.


SamiSousa:
@aravindhp I've addressed your comments and reorganized the operators to be directly in the `operators` directory. The generated artifacts are located in `operators/operator-artifacts/templates/` as created by the `helm template` command.

aravindhp:
>The generated artifacts are located in `operators/operator-artifacts/templates/` as created by the helm template command.

Do we need to check in these generated artifacts? Can't we just recreate them by running the helm command every time one of the operators change?

operators/how-to-create-artifacts.md
```diff
@@ -0,0 +1,29 @@
...
+
+```bash
+helm template -f values.yaml . --output-dir .
+```
+
+Your operator artifact(s) will be created in the `./operator-artifacts/templates/` directory.
+
+## Pushing your Operator Artifact to an App Registry
+
+Follow the steps [here](../docs/how-to-upload-artifact.md) for how to push your Operator Artifact to Quay's App Registry and add your operator offering to an OpenShift cluster.
```
kevinrizza: Maybe this doc should have some mention of prereqs? (i.e. helm needs to be installed?)

kevinrizza:
@aravindhp I think, based on looking at what's checked into OLM, the intention is that the templates are used to build specific releases which are checked in. i.e. if we wanted to cut a new release of marketplace that we would generate a new version of these files.

IMO that doesn't necessarily fit into our release model, so maybe we should reevaluate? This seems like it is tied to how we would build a pipeline for these artifact operators, which I don't think we have considered yet. Right now we are in the "just make something available" mode of delivery.

kevinrizza:
Another thought, I assume we are just using helm to generate these manifests because that is what OLM was doing, so we are copying it out of convenience? I think that might be fine for now, but we might want to use an alternative method of generating these files in the future. Because of that, it might just make more sense for us to check in the generated files for now rather than use this templating system in our code at all?

tkashem:
Great progress, @SamiSousa. 
Question: if we want to rebuild these from source then all we need to do is manually copy the individual files from OLM to their respective destination folder in `operators` and run the tool?

tkashem:
> > The generated artifacts are located in `operators/operator-artifacts/templates/` as created by the helm template command.
> 
> Do we need to check in these generated artifacts? Can't we just recreate them by running the helm command every time one of the operators change?

I think, what goes in to the appregistry repositories should be checked in so that we can track changes to the artifact.

tkashem:
Question: Once the artifacts are generated into the `templates` folder, how are we going to push them to `quay.io`, each operator has its own push script?

aravindhp:
> @aravindhp I think, based on looking at what's checked into OLM, the intention is that the templates are used to build specific releases which are checked in. i.e. if we wanted to cut a new release of marketplace that we would generate a new version of these files.
> 
> IMO that doesn't necessarily fit into our release model, so maybe we should reevaluate? This seems like it is tied to how we would build a pipeline for these artifact operators, which I don't think we have considered yet. Right now we are in the "just make something available" mode of delivery.

These blobs are not going to be connected to any Marketplace release. They are orthogonal to it. They just happen to live in the repo as a stop gap measure until we figure out the pipeline for uploading operator artifacts into Quay's app-registry. As you said, today we are in "just make something available" mode of delivery.

aravindhp:
> > > The generated artifacts are located in `operators/operator-artifacts/templates/` as created by the helm template command.
> > 
> > 
> > Do we need to check in these generated artifacts? Can't we just recreate them by running the helm command every time one of the operators change?
> 
> I think, what goes in to the appregistry repositories should be checked in so that we can track changes to the artifact.

Why are the templates being pushed into appregistry?

aravindhp:
> Great progress, @SamiSousa.
> Question: if we want to rebuild these from source then all we need to do is manually copy the individual files from OLM to their respective destination folder in `operators` and run the tool?

Once these operators are removed from rh-operators OLM won't be the source of truth.

aravindhp:
> Question: Once the artifacts are generated into the `templates` folder, how are we going to push them to `quay.io`, each operator has its own push script?

I would prefer one script that takes an operator as input which then gets pushed to quay.

SamiSousa:
> Another thought, I assume we are just using helm to generate these manifests because that is what OLM was doing, so we are copying it out of convenience?

It's out of convenience. In the future it would make sense to have some scripts which better automate the process of generating operator artifacts.

SamiSousa:
> > Question: Once the artifacts are generated into the `templates` folder, how are we going to push them to `quay.io`, each operator has its own push script?
> 
> I would prefer one script that takes an operator as input which then gets pushed to quay.

Right now the `scripts/push-to-quay` script is sufficient to push each operator one at a time.

tkashem:
> at takes an operator as input which then gets pushed to quay
Yes, agreed. We have 15 to 20 operators to start with, I would like to avoid a scenario where an operator ends up in a mismatched repository because we fat finger operator and repository name.  
If we haven't done it already, I recommend we lock the operator folder name to match the `repository` name and then change the script to accept only a folder/operator name. 

Another concern I have is if we fat finger a version number that is older than the current one. I believe quay.io/cnr will accept it as long as it is a semver. I will double check.

tkashem:
> > Great progress, @SamiSousa.
> > Question: if we want to rebuild these from source then all we need to do is manually copy the individual files from OLM to their respective destination folder in `operators` and run the tool?
> 
> Once these operators are removed from rh-operators OLM won't be the source of truth.

Agreed, but we are not the source of truth yet. We should do another copy when they actually remove them or when OLM stops accepting any change(s) on these operators. Just to make sure that we have the latest copy.

SamiSousa:
Closing. Operators that appear in marketplace will be stored in [github.com/operator-framework/community-operators](github.com/operator-framework/community-operators)

PR #65, opened 2018-12-14
Improve naming of CRs

aravindhp:
- Remove the prefixes of the created CatalogSourceConfigs, CatalogSources and ConfigMaps. This has been done to make the naming more consistent.
- Hard code the display name to "Marketplace Operators" and publisher to "Red Hat" in the CatalogSource. This has been done to make the OLM package UI display look nice.

### DIFF
--- a/pkg/catalogsourceconfig/configuring.go
+++ b/pkg/catalogsourceconfig/configuring.go
@@ -110,13 +110,13 @@ func (r *configuringReconciler) reconcileCatalogSource(csc *v1alpha1.CatalogSour
 	// Check if the CatalogSource already exists
 	catalogSourceGet := new(CatalogSourceBuilder).WithTypeMeta().CatalogSource()
 	key := client.ObjectKey{
-		Name:      v1alpha1.CatalogSourcePrefix + csc.Name,
+		Name:      csc.Name,
 		Namespace: csc.Spec.TargetNamespace,
 	}
 	err = r.client.Get(context.TODO(), key, catalogSourceGet)
 
 	// Update the CatalogSource if it exists else create one.
-	configMapName := v1alpha1.ConfigMapPrefix + csc.Name
+	configMapName := csc.Name
 	if err == nil {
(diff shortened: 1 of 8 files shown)


pkg/operatorsource/configuring.go
```diff
@@ -65,7 +58,7 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *v1alpha1.Oper
 
 	out = in
 
-	cscName := getCatalogSourceConfigName(in.Name)
+	cscName := in.Name
```
tkashem: This is not required for the UI, but we are removing all prefixes, right?

pkg/catalogsourceconfig/configuring.go
```diff
@@ -207,9 +207,8 @@ func newConfigMap(csc *v1alpha1.CatalogSourceConfig, data map[string]string) *co
 func newCatalogSource(csc *v1alpha1.CatalogSourceConfig, configMapName string) *olm.CatalogSource {
 	builder := new(CatalogSourceBuilder).
 		WithOwner(csc).
-		WithMeta(v1alpha1.CatalogSourcePrefix+csc.Name, csc.Spec.TargetNamespace).
-		// TBD: where do we get display name and publisher from?
```
kevinrizza: Maybe this comment should remain? I know we are hardcoding them for now, but the problem still exists if we want to make it come from somewhere dynamic (real datastore?)

pkg/operatorsource/configuring.go
aravindhp: Correct. This is for more consistent naming when you look at the details or the command line.

aravindhp: I will also add this to the commit message.

pkg/catalogsourceconfig/configuring.go
aravindhp: Done.

aravindhp:
@tkashem @kevinrizza I have addressed your comments

PR #66, opened 2018-12-17
Make datastore access concurrent

tkashem:
* Make `datastore` access concurrent since there will be concurrent reader(s) and writer(s) of the underlying datastore.
* Store registry metadata that includes release information associated with each repository in remote registry so that we can determine if there is a new release.
* Add a new function 'OperatorSourceHasUpdate' that returns true if the operator source in remote registry has any update(s).

The above feature(s) are going to pave the way for `sync in cluster datastore cache with remote registry update(s)`.

### DIFF
--- a/pkg/datastore/datastore.go
+++ b/pkg/datastore/datastore.go
@@ -84,69 +84,6 @@ type Writer interface {
 	GetOperatorSource(opsrcUID types.UID) (spec *v1alpha1.OperatorSourceSpec, ok bool)
 }
 
-// operatorSourceRow is what gets stored in datastore after an OperatorSource CR
-// is reconciled.
-//
-// Every reconciled OperatorSource object has a corresponding operatorSourceRow
-// in datastore. The Writer interface accepts a raw operator manifest and
-// marshals it into this type before writing it to the underlying storage.
-type operatorSourceRow struct {
-	// We store the Spec associated with a given OperatorSource object. This is
-	// so that we can determine whether Spec for an existing operator source
-	// has been updated.
-	//
-	// We compare the Spec of the received OperatorSource object to the one
-	// in datastore.
-	Spec *v1alpha1.OperatorSourceSpec
-
-	// Operators is the collection of all single-operator manifest(s) associated
-	// with the underlying operator source.
-	// The package name is used to uniquely identify the operator manifest(s).
-	Operators map[string]*SingleOperatorManifest
-}
-
-// GetPackages returns the list of available package(s) associated with an
-// operator source.
-// An empty list is returned if there are no package(s).
-func (r *operatorSourceRow) GetPackages() []string {
-	packages := make([]string, 0)
-	for packageID, _ := range r.Operators {
-		packages = append(packages, packageID)
-	}
-
-	return packages
-}
-
-// operatorSourceRowMap is a map that holds a collection of operator source(s)
-// represented by operatorSourceRow.
-// The UID of an OperatorSource object is used as the key to uniquely identify
-// an operator source.
-type operatorSourceRowMap map[types.UID]*operatorSourceRow
-
-// GetAllPackages return a list of all package(s) available across all
-// operator source(s).
-func (m operatorSourceRowMap) GetAllPackages() []string {
-	packages := make([]string, 0)
-	for _, row := range m {
-		packages = append(packages, row.GetPackages()...)
-	}
-
-	return packages
-}
-
-// GetAllPackagesMap returns a collection of all available package(s) across all
-// operator sources in a map. Package name is used as the key to this map.
-func (m operatorSourceRowMap) GetAllPackagesMap() map[string]*SingleOperatorManifest {
-	packages := map[string]*SingleOperatorManifest{}
-	for _, row := range m {
-		for packageID, manifest := range row.Operators {
-			packages[packageID] = manifest
-		}
-	}
-
-	return packages
-}
-
 // memoryDatastore is an in-memory implementation of operator manifest datastore.
 // TODO: In future, it will be replaced by an indexable persistent datastore.
 type memoryDatastore struct {
--- a/pkg/datastore/datastore.go
+++ b/pkg/datastore/datastore.go
@@ -22,7 +22,7 @@ func init() {
 // New returns an instance of memoryDatastore.
 func New() *memoryDatastore {
 	return &memoryDatastore{
-		rows:    operatorSourceRowMap{},
+		rows:    newOperatorSourceRowMap(),
 		parser:  &manifestYAMLParser{},
 		walker:  &walker{},
 		bundler: &bundler{},
@@ -111,7 +111,7 @@ type Writer interface {
 // memoryDatastore is an in-memory implementation of operator manifest datastore.
 // TODO: In future, it will be replaced by an indexable persistent datastore.
 type memoryDatastore struct {
-	rows    operatorSourceRowMap
+	rows    *operatorSourceRowMap
 	parser  ManifestYAMLParser
 	walker  ManifestWalker
 	bundler Bundler
@@ -158,19 +158,7 @@ func (ds *memoryDatastore) Write(opsrc *v1alpha1.OperatorSource, rawManifests []
 		metadata[rawManifest.RegistryMetadata.Repository] = &rawManifest.RegistryMetadata
 	}
 
-	row := &operatorSourceRow{
-		OperatorSourceKey: OperatorSourceKey{
-			Name: types.NamespacedName{
-				Namespace: opsrc.GetNamespace(),
-				Name:      opsrc.GetName(),
-			},
-			Spec: &opsrc.Spec,
-		},
-		Operators: operators,
-		Metadata:  metadata,
-	}
-
-	ds.rows[opsrc.GetUID()] = row
+	ds.rows.Add(opsrc, metadata, operators)
 
 	return nil
 }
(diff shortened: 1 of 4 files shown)


pkg/datastore/row.go
```diff
@@ -0,0 +1,154 @@
...
+	return &operatorSourceRowMap{
+		Sources: map[types.UID]*operatorSourceRow{},
+	}
+}
+
+// OperatorSourceKey is what datastore uses to relate to an OperatorSource
+// object.
+type OperatorSourceKey struct {
+	// UID is the UID associated with the OperatorSource object.
+	UID types.UID
```
kevinrizza: Why do we save this field in the struct? If we are creating a map that uses this value as a key, isn't it enough that the map itself contains the key? Why store it here and then never reference the field?

tkashem: Good point, I think my next PR uses it. If you want I can remove it now and then add it back in the next PR.

kevinrizza: Yeah, I think it would make more sense that way.

aravindhp: I also don't understand why we need the `UID` and `NamespacedName`. Given that both are unique shouldn't only one of them be required?

pkg/datastore/datastore.go
```diff
@@ -166,13 +206,49 @@ func (ds *memoryDatastore) RemoveOperatorSource(uid types.UID) {
...
+func (ds *memoryDatastore) OperatorSourceHasUpdate(opsrcUID types.UID, metadata []*RegistryMetadata) (bool, error) {
+	// TODO: Return fine grained information that describes repository that
+	// was removed, added or has a new release.
+	row, exists := ds.rows[opsrcUID]
+	if !exists {
+		return false, fmt.Errorf("datastore has no record of the specified OperatorSource [%s]", opsrcUID)
+	}
+
+	// Right now we consider remote and local operator source to be same only
+	// when the following is true:
```
aravindhp: `are true`

aravindhp: On another note, it will be more useful to say what is the criteria for an `OperatorSource` to have an update.

pkg/datastore/datastore.go
```diff
@@ -166,13 +206,49 @@ func (ds *memoryDatastore) RemoveOperatorSource(uid types.UID) {
...
+	//  b. Each repository in remote has a corresponding local repository with
+	//     exactly the same release.
+
+	if len(row.Metadata) != len(metadata) {
+		return true, nil
+	}
+
+	for _, remote := range metadata {
+		if remote.Release == "" {
+			return false, fmt.Errorf("Release not specified for repository [%s]", remote.ID())
```
aravindhp: - Should this be `false`? Won't it be safer to update the `OperatorSource` as not having a version signifies a change in the version? 
- What about the scenario where the repository never had a version in the first place?

pkg/datastore/row.go
```diff
@@ -59,13 +67,70 @@ func (r *operatorSourceRow) GetPackages() []string {
 // represented by operatorSourceRow.
 // The UID of an OperatorSource object is used as the key to uniquely identify
 // an operator source.
-type operatorSourceRowMap map[types.UID]*operatorSourceRow
+type operatorSourceRowMap struct {
+	lock sync.RWMutex
```
aravindhp: Should we consider using a thread safe cache like `go-cache` at this point?

pkg/datastore/row.go
tkashem: datastore uses the UID in its methods and we need the namespaced name to query the object via the client. datastore can be changed to use the Namespaced name only, but I would leave it to a future refactor. 
@kevinrizza removed the UID from this PR, I will add it back the next.

pkg/datastore/row.go
tkashem: My use case does not go beyond that of a standard `map`, `go-cache` seems like an overkill.

pkg/datastore/datastore.go
tkashem: We should always have a release, I don't think we can push without specifying a release. I put it here as a defensive guard. `Not having a version` is treated as an error and hence I can't use it to indicate a change in version.

pkg/datastore/datastore.go
tkashem:
```
	//   New repositories added to the operator source.
	//   Repositories that have been removed from the operator source.
	//   Repositories that have a new version.
```
I already have it in the doc for `OperatorSourceHasUpdate` function. Is this what you were looking for?

pkg/datastore/datastore.go
```diff
@@ -81,7 +81,28 @@ type Writer interface {
...
 	// a Spec already exists. If no Spec is found then the function
 	// returns (nil, false).
-	GetOperatorSource(opsrcUID types.UID) (spec *v1alpha1.OperatorSourceSpec, ok bool)
+	GetOperatorSource(opsrcUID types.UID) (key *OperatorSourceKey, ok bool)
+
+	// OperatorSourceHasUpdate returns true if the operator source in remote
+	// registry specified in metadata has update(s) that need to be pulled.
+	//
+	// It's a rather rudimentary implementation, the function returns true if
+	// the remote registry has any update(s). It does not provide specific
```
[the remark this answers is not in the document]
tkashem: I clarified the doc.

pkg/datastore/datastore.go
```diff
@@ -81,7 +81,28 @@ type Writer interface {
 	// datastore uses the UID of the given OperatorSource object to check if
 	// a Spec already exists. If no Spec is found then the function
 	// returns (nil, false).
-	GetOperatorSource(opsrcUID types.UID) (spec *v1alpha1.OperatorSourceSpec, ok bool)
+	GetOperatorSource(opsrcUID types.UID) (key *OperatorSourceKey, ok bool)
+
+	// OperatorSourceHasUpdate returns true if the operator source in remote
+	// registry specified in metadata has update(s) that need to be pulled.
+	//
+	// It's a rather rudimentary implementation, the function returns true if
```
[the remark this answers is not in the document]
tkashem: I would like to leave it as is since I am trying to be consistent with the signature of a get from map (value, ok).

tkashem:
@aravindhp @kevinrizza addressed your feedback, take a look when you have time.

pkg/datastore/datastore.go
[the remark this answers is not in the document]
aravindhp: But the meaning of this in the context is different. It just so happens that you are using a map. That doesn't need to be exposed to the caller.

pkg/datastore/datastore.go
aravindhp: If it is there in the doc for that function, it can be repeated here. No reason of the reverse documentation.

pkg/datastore/row.go
aravindhp: For the record, `go-cache` is pretty light weight. There are other alternatives too. I would rather we not have our own cache implementation, so please add a story to replace this with an off the shelf library.

tkashem: Right now I don't see any use case(s) that goes beyond standard `map`. In the future if there are new use case(s) that warrant a cache library then we will look into it.

pkg/datastore/datastore.go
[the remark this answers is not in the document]
kevinrizza: I personally prefer the use of the `ok` variable name since it is very clear and easily readable as a standard.

pkg/datastore/row.go
kevinrizza: I tend to agree here. What value at the moment do we get from `go-cache` that we don't get here? I think a simple lock should suffice in this case

pkg/datastore/datastore.go
tkashem: The comments I put inside the function is to explain the `for loop`. I prefer not to repeat the entire doc of the function but if you think repeating it is useful then I can copy it.

aravindhp: I don't want it to be repeated. It just seems that you are restating the conditions in a negated fashion. Let me go through your next set of changes and see how it reads. I will comment if it needs any change.

pkg/datastore/row.go
aravindhp: At a minimum this should be a concurrent data structure. Otherwise we risk introducing code that directly accesses the cache without grabbing the lock. But I won't argue further about this.

pkg/datastore/datastore.go
```diff
@@ -166,13 +207,49 @@ func (ds *memoryDatastore) RemoveOperatorSource(uid types.UID) {
...
+
+func (ds *memoryDatastore) OperatorSourceHasUpdate(opsrcUID types.UID, metadata []*RegistryMetadata) (bool, error) {
+	// TODO: Return fine grained information that describes repository that
+	// was removed, added or has a new release.
+	row, exists := ds.rows[opsrcUID]
+	if !exists {
+		return false, fmt.Errorf("datastore has no record of the specified OperatorSource [%s]", opsrcUID)
+	}
+
+	// Right now we consider remote and local operator source to be same only
```
aravindhp: Not sure how useful this is. It is a repeat of what you have in the function description: `following event(s) indicate that a remote registry has been updated.`, whereas here you have it as when a remote registry has **not** been updated

pkg/datastore/row.go
tkashem: This PR addresses the issue of concurrency, we are adding locks to make it so.

aravindhp: And hence introducing a chance of accessing the datastore without grabbing the lock which won't happen if we use a thread safe cache or inherently concurrent data structure library. Like I said, I won't argue further about this.

pkg/datastore/datastore.go
tkashem: I removed it.

tkashem:
@kevinrizza @aravindhp I believe with this commit, all feedback has been addressed. Please take a look.

PR #70, opened 2018-12-18
[opsrc] Handle OperatorSource deletion

kevinrizza:
-Add deletedReconciler to opsrc reconcilers
-Use finalizer to reconcile before deleting opsrc
-Delete manifests from cache when deleting

### DIFF
--- a/pkg/operatorsource/deleted.go
+++ b/pkg/operatorsource/deleted.go
@@ -1,0 +1,64 @@
+package operatorsource
+
+import (
+	"context"
+
+	"github.com/operator-framework/operator-marketplace/pkg/apis/marketplace/v1alpha1"
+	"github.com/operator-framework/operator-marketplace/pkg/datastore"
+	log "github.com/sirupsen/logrus"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// NewDeletedReconciler returns a Reconciler that reconciles
+// an OperatorSource that has been marked for deletion.
+func NewDeletedReconciler(logger *log.Entry, datastore datastore.Writer, client client.Client) Reconciler {
+	return &deletedReconciler{
+		logger:    logger,
+		datastore: datastore,
+		client:    client,
+	}
+}
+
+// deletedReconciler is an implementation of Reconciler interface that
+// reconciles an OperatorSource object that has been marked for deletion.
+type deletedReconciler struct {
+	logger    *log.Entry
+	datastore datastore.Writer
+	client    client.Client
+}
+
+// Reconcile reconciles an OperatorSource object that is marked for deletion.
+// In the generic case, this is called when the OperatorSource has been marked
+// for deletion. It removes all data related to this OperatorSource from the
+// datastore, and it removes the OperatorSource finalizer from the object so
+// that it can be cleaned up by the garbage collector.
+//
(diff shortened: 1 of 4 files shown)


pkg/operatorsource/deleted.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// now delete the operator source manifests
+	r.datastore.RemoveOperatorSource(out.UID)
+
+	return out, nil, nil
+}
+
+func removeFinalizer(opsrc *v1alpha1.OperatorSource) error {
+	outFinalizers := make([]string, 0)
+	for _, finalizer := range opsrc.ObjectMeta.Finalizers {
```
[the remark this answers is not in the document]
tkashem: After giving it some thoughts, let's leave it as is.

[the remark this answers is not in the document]
kevinrizza: Yeah I think my thought process here is that there is no guarantee that this is the only finalizer that will ever be attached to this object.

kevinrizza:
@tkashem Thanks for the review! I made changes all related to your comments, please take a look when you have a chance.

pkg/operatorsource/deleted.go
```diff
@@ -0,0 +1,58 @@
...
+//
+// nextPhase represents the next desired phase for the given OperatorSource
+// object. If nil is returned, it implies that no phase transition is expected.
+func (r *deletedReconciler) Reconcile(ctx context.Context, in *v1alpha1.OperatorSource) (out *v1alpha1.OperatorSource, nextPhase *v1alpha1.Phase, err error) {
+	out = in
+
+	// Remove the opsrc finalizer from the object
+	out.RemoveFinalizer()
+
+	err = r.client.Update(context.TODO(), out)
```
aravindhp: Is this `Update()` required given we call it in the main handler?

pkg/operatorsource/deleted.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	r.logger.Info("Finalizer removed, now garbage collector will clean it up.")
+
+	// now delete the operator source manifests
+	r.datastore.RemoveOperatorSource(out.UID)
```
tkashem: I think if `r.datastore.RemoveOperatorSource(out.UID)` precedes the call to `update` then datastore is in desired state even though `update` fails. It takes us a tiny bit closer to desired state(s) before the next reconciliation attempt is made. :) You don't need to make the change, I was just thinking out loud in terms of the order of these calls.

pkg/operatorsource/deleted.go
kevinrizza: We are not setting `nextPhase` because we aren't transitioning into a phase after using the delete reconciler. So either we would need to create a phase after delete reconciler which exists purely to follow this pattern and be a no-op, or we would have to use the client to update the operator source outside of the transition function (like it is currently written)

tkashem: It might be required, since you are not changing the phase the transitioner will deem as if nothing has changed. We don't do a `out.DeepEquals(in)` yet.

aravindhp: Please add a comment. This is not obvious to the reader.

PR #71, opened 2018-12-18
Keep datastore in sync with registry update(s)

tkashem:
Keep in-memory OperatorSource cache of datastore in sync with remote registry update(s). Rebuild the local cache when any of the following happens:
- A new repository has been added to the given namespace in remote registry associated with the OperatorSource object.
- An existing repository has been removed from the given namespace in remote registry associate with the OperatorSource object.
- A new release has been pushed for an existing repository in remote registry associate with the OperatorSource object.

Implement the following to accomplish the above goal:
* Add a new function 'ListPackages' in appregistry that returns latest metadata (with release information) for all repositories under the given namespace.
* Add 'Poller' that polls for changes to the underlying namespace associated with each OperatorSource object that datastore is aware of.
* Compare local metadata with remote and determine if there is any update. If an OperatorSource has update(s) available in remote registry then trigger a rebuild of datastore cache for the given OperatorSource object.
* Implement a 'RegistrySyncer' loop that checks every N minute(s) and kicks off the 'Polling' operation.
* Kick off the 'RegistrySyncer.Sync' method in a dedicated goroutine when the operator process starts.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -5,10 +5,12 @@ import (
 	"log"
 	"net/http"
 	"runtime"
+	"time"
 
 	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/operator-framework/operator-marketplace/pkg/apis"
 	"github.com/operator-framework/operator-marketplace/pkg/controller"
+	"github.com/operator-framework/operator-marketplace/pkg/operatorsource"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
@@ -68,7 +70,15 @@ func main() {
 	go http.ListenAndServe(":8080", nil)
 
 	log.Print("Starting the Cmd.")
+	stopCh := signals.SetupSignalHandler()
+
+	// TODO: resyncInterval is hardcoded to 1 hour now, it would have to be
+	// configurable on a per OperatorSource level.
+	initialWait, resyncInterval := time.Duration(10), time.Duration(60)
+	syncer := operatorsource.NewRegistrySyncer(mgr.GetClient(), initialWait, resyncInterval)
+
+	go syncer.Sync(stopCh)
 
 	// Start the Cmd
-	log.Fatal(mgr.Start(signals.SetupSignalHandler()))
+	log.Fatal(mgr.Start(stopCh))
 }
--- a/pkg/appregistry/client.go
+++ b/pkg/appregistry/client.go
@@ -17,6 +17,10 @@ type Client interface {
 
 	// RetrieveOne retrieves a given package from the source
 	RetrieveOne(name, release string) (*datastore.OperatorMetadata, error)
+
+	// ListPackages returns metadata associated with each package in the
+	// specified namespace.
+	ListPackages(namespace string) ([]*datastore.RegistryMetadata, error)
 }
 
 type client struct {
@@ -38,6 +42,36 @@ func (c *client) RetrieveAll(namespace string) ([]*datastore.OperatorMetadata, e
 		}
 
 		list[i] = manifest
+	}
+
+	return list, nil
+}
+
+func (c *client) ListPackages(namespace string) ([]*datastore.RegistryMetadata, error) {
+	packages, err := c.adapter.ListPackages(namespace)
+	if err != nil {
+		return nil, err
+	}
+
+	list := make([]*datastore.RegistryMetadata, len(packages))
+	for i, pkg := range packages {
+		namespace, repository, err := split(pkg.Name)
+		if err != nil {
+			return nil, err
+		}
+
+		metadata := &datastore.RegistryMetadata{
+			Namespace:  namespace,
+			Repository: repository,
+
+			// 'Default' points to the latest release pushed.
+			Release: pkg.Default,
+
+			// Getting 'Digest' would require an additional call to the app
+			// registry, so it is being defaulted.
+		}
+
+		list[i] = metadata
 	}
 
 	return list, nil
--- a/pkg/operatorsource/poller.go
+++ b/pkg/operatorsource/poller.go
@@ -1,0 +1,81 @@
+package operatorsource
+
+import (
+	"fmt"
+
+	"github.com/operator-framework/operator-marketplace/pkg/appregistry"
+	"github.com/operator-framework/operator-marketplace/pkg/datastore"
+	"github.com/operator-framework/operator-marketplace/pkg/phase"
+	log "github.com/sirupsen/logrus"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// NewPoller returns a new instance of Poller interface.
+func NewPoller(client client.Client) Poller {
+	poller := &poller{
+		datastore: datastore.Cache,
+		helper: &pollHelper{
+			factory:      appregistry.NewClientFactory(),
+			datastore:    datastore.Cache,
+			client:       client,
+			transitioner: phase.NewTransitioner(),
+		},
+	}
+
+	return poller
+}
+
+// Poller is an interface that wraps the Poll method.
+//
+// Poll iterates through all available operator source(s) that are in the
+// underlying datastore and performs the following action(s):
+//   a) It polls the remote registry namespace to check if there are any
+//      update(s) available.
+//
+//   b) If there is an update available then it triggers a purge and rebuild
+//      operation for the specified OperatorSource object.
+//
+// On any error during each iteration it logs the error encountered and moves
+// on to the next OperatorSource object.
+type Poller interface {
+	Poll()
+}
+
+// poller implements the Poller interface.
+type poller struct {
+	helper    PollHelper
+	datastore datastore.Writer
+}
+
+func (p *poller) Poll() {
+	sources := p.datastore.GetAllOperatorSources()
+
+	for _, source := range sources {
+		if err := p.pollSource(source); err != nil {
+			log.Errorf("%v", err)
+		}
+	}
+}
+
+func (p *poller) pollSource(source *datastore.OperatorSourceKey) error {
+	updated, err := p.helper.HasUpdate(source)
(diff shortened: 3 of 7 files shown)


kevinrizza:
Pretty much exactly what we discussed, looks good except for a few small comments.

pkg/operatorsource/updater.go
```diff
@@ -0,0 +1,113 @@
...
+
+		// Otherwise, it is an error.
+		updateErr = err
+		return
+	}
+
+	// Needed because sdk does not get the gvk.
+	instance.EnsureGVK()
+
+	if instance.GetCurrentPhaseName() == phase.OperatorSourceDownloading {
```
aravindhp: Is checking for `OperatorSourceDownloading` sufficient? Shouldn't any sort of sync happen only if the status is `Succeeded` or `Failed`?

pkg/operatorsource/updater.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/operator-framework/operator-marketplace/pkg/appregistry"
+	"github.com/operator-framework/operator-marketplace/pkg/datastore"
+	"github.com/operator-framework/operator-marketplace/pkg/phase"
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Updater is an interface that can be used to check whether a remote registry
+// has any update(s) and trigger a rebuild of in cluster OperatorSource cache.
+type Updater interface {
```
aravindhp: Given we talk about CR update reconciliation, can we come up with another name for this interface and struct to signify that you are updating the underlying datastore and not the CR itself?

cmd/manager/main.go
```diff
@@ -68,7 +70,15 @@ func main() {
...
 
 	log.Print("Starting the Cmd.")
+	stopCh := signals.SetupSignalHandler()
+
+	// TODO: Registry resync period is hardcoded to 1 minute now, make it
+	// configurable via command line.
+	initialWait, resyncWait := time.Duration(5), time.Duration(1)
+	syncer := operatorsource.NewRegistrySyncer(mgr.GetClient(), initialWait, resyncWait)
+
+	go syncer.Sync(stopCh)
```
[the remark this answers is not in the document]
tkashem: yes, correct.

cmd/manager/main.go
```diff
@@ -68,7 +70,15 @@ func main() {
...
 
 	// Start the Cmd
-	log.Fatal(mgr.Start(signals.SetupSignalHandler()))
+	log.Fatal(mgr.Start(stopCh))
```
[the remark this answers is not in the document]
tkashem: I agree, we have to find a way to layer it appropriately, maybe in a separate PR?

pkg/appregistry/client.go
```diff
@@ -43,6 +47,36 @@ func (c *client) RetrieveAll(namespace string) ([]*datastore.OperatorMetadata, e
 	return list, nil
 }
 
+func (c *client) ListPackages(namespace string) ([]*datastore.RegistryMetadata, error) {
+	packages, err := c.adapter.ListPackages(namespace)
```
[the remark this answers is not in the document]
tkashem: It should not, even if it does iterating a nil slice is equivalent to iterating an empty slice, I believe.

pkg/operatorsource/poller.go
```diff
@@ -0,0 +1,81 @@
...
+// Poll iterates through all available operator source(s) that are in the
+// underlying datastore and performs the following action(s):
+//   a) It polls the remote registry namespace to check if there are any
+//      update(s) available.
+//
+//   b) If there is an update available then it triggers a purge andrebuild
+//      operation for the specified OperatorSource object.
+//
+// On ay error during each iteration it logs the error encountered and moves
+// on ti the next OperatorSource object.
```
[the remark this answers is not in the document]
tkashem: interesting, not sure why the spell checker did not catch any of these typos, they are fixed.

pkg/operatorsource/updater.go
tkashem: `RegistryUpdater` ok?

pkg/operatorsource/updater.go
tkashem: Ideally we don't want to rely on phase transition logic. It introduces a coupling I think. If the object is stuck at a particular phase then we would not pull update(s).  
But the downside is on rare occasion we might pull update(s) immediately after a fresh download.
In reality, if there are too many updates of an object within a short amount time, k8s will discard the older version(s) and will inform of the latest version only.

cmd/manager/main.go
```diff
@@ -68,7 +70,15 @@ func main() {
 	go http.ListenAndServe(":8080", nil)
 
 	log.Print("Starting the Cmd.")
+	stopCh := signals.SetupSignalHandler()
+
+	// TODO: Registry resync period is hardcoded to 1 minute now, make it
+	// configurable via command line.
+	initialWait, resyncWait := time.Duration(5), time.Duration(1)
```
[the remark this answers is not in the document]
tkashem: So if we hard code the interval to 1-hour for now, we are going to force qe to wait an hour to run their test :)

tkashem:
@kevinrizza @aravindhp addressed the comments, can you take a look when you have time?

pkg/operatorsource/updater.go
aravindhp: >Ideally we don't want to rely on phase transition logic. It introduces a coupling I think. 

That is what I was hoping for but I thought you wanted that coupling.

>If the object is stuck at a particular phase then we would not pull update(s).

Why should we pull updates if it is stuck in a state? Are you thinking of using this thread as a kicker to make it keep moving as a side effect? I am not a big fan of that side effect.

>In reality, if there are too many updates of an object within a short amount time, k8s will discard the older version(s) and will inform of the latest version only.

I don't see how this relates to my original question.

cmd/manager/main.go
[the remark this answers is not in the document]
aravindhp: :man_shrugging: They can multi-task :smile:

pkg/operatorsource/updater.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/operator-framework/operator-marketplace/pkg/appregistry"
+	"github.com/operator-framework/operator-marketplace/pkg/datastore"
+	"github.com/operator-framework/operator-marketplace/pkg/phase"
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// RegistryUpdater is an interface that can be used to check whether a remote registry
+// has any update(s) and trigger a rebuild of in cluster OperatorSource cache.
+type RegistryUpdater interface {
```
aravindhp: FYI, the filename is still `updater.go`. Shouldn't that be `registry_updater.go`?

pkg/operatorsource/updater.go
tkashem: * I am happy to remove the check for `Downloading` phase. I added this based on your question `what if a download is in progress while we are kicking off an datastore update`.

* I believe this is not a side effect, if you have new update(s) it's time to reconcile the OperatorSource CR irrespective of what phase it is in. 

* I mention this because k8s internally optimizes and informs of the latest version only in between two reconciliation window.

aravindhp: I am not asking for the check to be removed. My point is that if the `OperatorSource` is in any phase other than `Succeeded` or `Failed` we should not pull down updates. Instead we back off and grab the updates during the next sync period. So the check should reflect that.

>I believe this is not a side effect, if you have new update(s) it's time to reconcile the OperatorSource CR irrespective of what phase it is in.

If a particular phase (not `Succeeded` or `Failed`) is having an issue preventing it from moving to the next, there is a potential for it to keep going back and forth between the erroneous phase and `Purging` every sync cycle. It is better at that point for the admin to intervene.

pkg/operatorsource/updater.go
aravindhp: `RegistryUpdater` seems to imply that you are updating the registry. How about `CacheUpdater`?

tkashem:
@aravindhp are you referring to the `updater` rename? The link is not working for me for some reason.

pkg/operatorsource/updater.go
tkashem: > there is a potential for it to keep going back and forth between the erroneous phase and `Purging` every sync cycle. It is better at that point for the admin to intervene.

* A sync cycle is not designed to cause `back and forth`. Can you give me a scenario?
* Even if there is potential for `back and forth`, this function is not the place to address it. This function does one thing - if there are update(s) for an `OperatorSource`, request for reconciliation to start anew. That's what I meant when I referred to less coupling, this function should not care what phase the object is in. It's intention is to request a reconciliation.
* If there are update(s), sync cycle should be able to trigger reconciliation, without need of any human intervention. It's safe not to assume that the `OperatorSource` object will be in certain phase(s) in an asynchronous world.

I summarized my points, if you still want the change, I will add the check.

aravindhp: >A sync cycle is not designed to cause back and forth. Can you give me a scenario?

I gave you a generic scenario in my previous response. There could be an error during validation causing it go to failed. The sync cycle will set it back to purging and it will again fail in validation. This will keep happening on very sync cycle unless the root cause of the validation failure is not fixed by updating the CR spec. Yes, there is no harm in this happening but do we need this churn?

>That's what I meant when I referred to less coupling, this function should not care what phase the object is in. It's intention is to request a reconciliation.

If that is your desired effect then there should be no check.

>If there are update(s), sync cycle should be able to trigger reconciliation, without need of any human intervention. It's safe not to assume that the OperatorSource object will be in certain phase(s) in an asynchronous world.

I am not saying a sync cycle should require human intervention nor am I saying that we should assume the object should be in a particular phase. I am just calling out that the sync cycle can cause an object to keep going back and forth between two phases without it ever succeeding. I hope my above example helped clarify this. 

Any how I think it is best we table this discussion for now and address it separately if needed. I will approve this as it stands or with an updated check or without a check. I leave that to you.

tkashem:
@aravindhp @kevinrizza addressed the feedback, can you take a look when you have time.

PR #72, opened 2018-12-18
[opsrc] Handle faulty operator manifest(s)

tkashem:
Handle faulty operator manifest(s) in remote registry associated
with an OperatorSource object in a graceful manner:

- Log error(s) encountered for faulty manifest(s).
- If a faulty package manifest is encountered move on to the next
  repository in the namespace as opposed to aborting the entire
  operation.

### DIFF
--- a/pkg/operatorsource/downloading.go
+++ b/pkg/operatorsource/downloading.go
@@ -77,15 +77,22 @@ func (r *downloadingReconciler) Reconcile(ctx context.Context, in *v1alpha1.Oper
 
 	r.logger.Infof("Downloaded %d manifest(s) from the operator source endpoint", len(manifests))
 
-	err = r.datastore.Write(in, manifests)
+	count, err := r.datastore.Write(in, manifests)
 	if err != nil {
-		nextPhase = phase.GetNextWithMessage(phase.Failed, err.Error())
-		return
+		if count == 0 {
+			// No operator manifest was written, move to Failed phase.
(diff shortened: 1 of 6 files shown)


pkg/operatorsource/downloading.go
```diff
@@ -77,15 +77,22 @@ func (r *downloadingReconciler) Reconcile(ctx context.Context, in *v1alpha1.Oper
...
+		}
+
+		r.logger.Infof("There were some faulty operator manifest(s), errors - %v", err)
+		err = nil
 	}
 
 	packages := r.datastore.GetPackageIDsByOperatorSource(out.GetUID())
 	out.Status.Packages = packages
 
+	r.logger.Infof("Successfully stored %d operator manifest(s)", count)
```
aravindhp: Nit: It will nice to show how many were successfully stored vs how many were downloaded. For example `Stored 9/10 operator manifest(s)`.

tkashem: When we have downloaded, what we know is `the number of repositories` associated with the operator source. Unfortunately we don't have the total number of operator(s) since one repository can have multiple operator(s).

I can add a `Status.FaultyRepositories` attribute to specify the faulty repositories so that the admin is aware and can take action. (I will do it in a separate PR)

aravindhp: It's just a nice to have. No need for it at the moment.

PR #77, opened 2018-12-20
Update the cache before reconciling

kevinrizza:
-When configuring a csc, update the cache upfront


kevinrizza:
@aravindhp @tkashem Improved commit message and code comments

PR #79, opened 2018-12-20
[opsrc] refactor configuring reconciler

tkashem:
- client.Get needs an empty instance, not populated.

( My PR to follow will need this refactor )

### DIFF
--- a/pkg/operatorsource/configuring_test.go
+++ b/pkg/operatorsource/configuring_test.go
@@ -39,7 +39,7 @@ func TestReconcile_NotConfigured_NewCatalogConfigSourceObjectCreated(t *testing.
 	namespacedName := types.NamespacedName{Name: "foo", Namespace: "marketplace"}
 
 	// We expect that the given CatalogConfigSource object does not exist.
-	cscGet := helperNewCatalogSourceConfig(opsrcIn.Namespace, opsrcIn.Name)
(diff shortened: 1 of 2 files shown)


tkashem:
@aravindhp @kevinrizza can you check now?

aravindhp:
@tkashem your commit message still doesn't say why you are making the change. I see `( My PR to follow will need this refactor )` in the Pr comment. But even that doesn't explain why the refactor is needed. Ex: A follow up PR need this refactor because _______

pkg/operatorsource/configuring_test.go
```diff
@@ -39,15 +39,15 @@ func TestReconcile_NotConfigured_NewCatalogConfigSourceObjectCreated(t *testing.
 	namespacedName := types.NamespacedName{Name: "foo", Namespace: "marketplace"}
 
 	// We expect that the given CatalogConfigSource object does not exist.
-	cscGet := helperNewCatalogSourceConfig(opsrcIn.Namespace, opsrcIn.Name)
+	cscGet := &v1alpha1.CatalogSourceConfig{}
```
SamiSousa: This makes sense to me; so a call to `client.Get` populates the provided reference, correct? So populating it with the helper isn't necessary since it will be overwritten.

aravindhp: This is the sort of explanation that should go into the commit message.

PR #80, opened 2018-12-21
[opsrc] Add support to filter operator offering(s)

tkashem:
Allow the admin to filter operator offering(s) by Operator Type.
With multiple operator source(s) an admin needs the ability to filter
operator offering(s) by group. For example, we want to list operator
offering(s) from "Community" or "ISV" etc.

Make the following change(s) to achieve this goal:
- Apply all label(s) associated with an OperatorSource to its
  corresponding CatalogSourceConfig object.

This will ensure that the corresponding CatalogSource object
has all the label(s) from the OperatorSource object.

This will cause Package API server to return the same label(s) for
all PackageManifest object(s) related to the CatalogSource
in question.

We can use a specific label `opsrc-group` that reflects the group
name of an operator source. The UI can utilize this label to
filter operator offering(s) from multiple source(s).

### DIFF
--- a/pkg/operatorsource/catalogsourceconfigbuilder.go
+++ b/pkg/operatorsource/catalogsourceconfigbuilder.go
@@ -22,19 +22,38 @@ func (b *CatalogSourceConfigBuilder) CatalogSourceConfig() *v1alpha1.CatalogSour
 	return &b.object
 }
 
-// WithMeta sets TypeMeta and ObjectMeta accordingly.
-func (b *CatalogSourceConfigBuilder) WithMeta(namespace, name string) *CatalogSourceConfigBuilder {
+// WithTypeMeta sets TypeMeta of the CatalogSourceConfig object.
+func (b *CatalogSourceConfigBuilder) WithTypeMeta() *CatalogSourceConfigBuilder {
 	b.object.TypeMeta = metav1.TypeMeta{
 		APIVersion: fmt.Sprintf("%s/%s",
 			v1alpha1.SchemeGroupVersion.Group, v1alpha1.SchemeGroupVersion.Version),
 		Kind: v1alpha1.CatalogSourceConfigKind,
 	}
 
-	b.object.ObjectMeta = metav1.ObjectMeta{
-		Name:      name,
-		Namespace: namespace,
-		Labels:    map[string]string{DatastoreLabel: "true"},
+	return b
+}
+
+// WithNamespacedName sets name and namespace of the CatalogSourceConfig object.
+func (b *CatalogSourceConfigBuilder) WithNamespacedName(namespace, name string) *CatalogSourceConfigBuilder {
+	b.object.SetNamespace(namespace)
+	b.object.SetName(name)
+
+	return b
+}
+
+// WithLabels sets appropriate labels for the CatalogSourceConfig object. It
+// applies all labels associated with an OperatorSource object specified in
+// opsrcLabels.
+func (b *CatalogSourceConfigBuilder) WithLabels(opsrcLabels map[string]string) *CatalogSourceConfigBuilder {
+	labels := map[string]string{
+		DatastoreLabel: "true",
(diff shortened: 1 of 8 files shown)


pkg/apis/marketplace/v1alpha1/operatorsource_types.go
```diff
@@ -49,6 +49,12 @@ type OperatorSourceSpec struct {
 	// manifests under this namespace will be visible.
 	// Please note that this is not a k8s namespace.
 	RegistryNamespace string `json:"registryNamespace,omitempty"`
+
+	// GroupName specifies the group name of the operator offering(s) associated
+	// with given OperatorSource object. For example, an admin can specify
+	// "Community" or "ISV" as the group name.
+	// This is an optional attribute.
+	GroupName string `json:"group,omitempty"`
```
tkashem: The original jira story says by `Operator Type` which is not clear. Please suggest a better name.

pkg/operatorsource/catalogsourceconfigbuilder.go
```diff
@@ -7,10 +7,16 @@ import (
...
-const DatastoreLabel string = "opsrc-datastore"
+const (
+	// DatastoreLabel is the label used in a CatalogSourceConfig to indicate that
+	// the resulting CatalogSource acts as the datastore for the OperatorSource
+	// if it is set to "true".
+	DatastoreLabel string = "opsrc-datastore"
+
+	// GroupLabel is the key of the label that is used to represent the group of
+	// the associated OperatorSource object.
+	GroupLabel string = "opsrc-group"
```
tkashem: Same here, looking for a better name.

pkg/apis/marketplace/v1alpha1/operatorsource_types.go
kevinrizza: Shouldn't it be based on whatever the field looks like on the UI? I assume that field will be visible there.

pkg/operatorsource/catalogsourceconfigbuilder.go
```diff
@@ -22,20 +28,37 @@ func (b *CatalogSourceConfigBuilder) CatalogSourceConfig() *v1alpha1.CatalogSour
...
 
-	b.object.ObjectMeta = metav1.ObjectMeta{
-		Name:      name,
-		Namespace: namespace,
-		Labels:    map[string]string{DatastoreLabel: "true"},
+	return b
+}
+
+// WithNamespacedName sets name and namespace of the CatalogSourceConfig object.
+func (b *CatalogSourceConfigBuilder) WithNamespacedName(namespace, name string) *CatalogSourceConfigBuilder {
```
kevinrizza: I'm not opposed to this refactor, but is it necessary for this pr?

tkashem: Previously, I had `WithMeta` that added the label as well. I did not want to populate an object with a label to pass it to `client.Get`. So some refactor was necessary, I chose to break it down by typemeta, namespacedName and label.

pkg/apis/marketplace/v1alpha1/operatorsource_types.go
aravindhp: Do we really need to expose this to the end user? Can't we just add labels to the `OperatorSource` CR and then pass them down to the `CatalogSourceConfig` and `CatalogSource`? This way if the UI filtering requirements keep growing it can be handled without making any operator changes.  

Moreover I think these labels will be used only in the default CRs that we are adding to the cluster. I don't believe we want expose this to the cluster admin. Yes they could add labels to their CRs but at least we are not advertising it at the `OpetatorSource` API level. We can then revisit UI requirements in the 4.1 time-frame and introduce this if needed at that point.

tkashem:
@aravindhp @kevinrizza can you check when you have time?

PR #81, opened 2018-12-21
[opsrc] Make configuring reconciler idempotent

tkashem:
Make configuring reconciler for OperatorSource object idempotent.
Right now, if the given CatalogSourceConfig object already exists
then the reconciler assumes that it is already properly configured
and moves on. This may not be true, due to unexpected error we might
have stale CatalogSourceConfig object. Making the configuring
reconciler idempotent helps us work around these issue(s).

Do the following to make it idempotent:
- Call Create to cretae the CatalogSourceConfig object.
- If we get an AlreadyExists error, then get the latest version
  of the object.
- Set the fields appropriately and then Call Update.

### DIFF
--- a/pkg/operatorsource/configuring.go
+++ b/pkg/operatorsource/configuring.go
@@ -59,38 +59,53 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *v1alpha1.Oper
 
 	out = in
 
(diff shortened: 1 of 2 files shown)


tkashem:
@aravindhp you can look at it now.

PR #83, opened 2019-01-05
[install] Add support for installation by CVO

aravindhp:
- Move to using "openshift-marketplace" namespace instead of "openshift-operators"
- Add a "manifests" folder consisting of an ordered set of resources required by the operator. CVO will apply them in the given order. Note that the "imagePullPolicy" is set to "IfNotPresent", which differs from the deployment in the CSV. This is because the image will be replaced by the CVO as per image-references.
- Update the Dockerfiles to add the "manifests" folder and io.openshift.release.operator label
- Update the README to reflect that Marketplace will be running by default in the "openshift-marketplace" namespace


tkashem:
Are you going to add the default `OperatorSource` CR in a separate PR? We also need to remove `OperatorSource` CRD and default CR from OLM.

aravindhp:
Yes I will add the default CR after it's removed from OLM in a separate PR. I want this merged and tested from the installer first.

aravindhp:
>Remove creation of resources from the e2e tests as they will be part
of the cluster by default. Update the SDK local run configuration to not
create any resources.

I am not sure if I should remove this in this PR or a follow up PR. I will debug the e2e tests and decide on that.

tkashem:
Thinking out loud - immediately after deployment, reconciliation of the default `OperatorSource` CR(s) might fail if `CatalogSource` CRD is not registered (this might happen if CVO hasn't got to OLM yet in release payload). Although, with future retries it should  eventually get reconciled. 

As an option, we could deploy Marketplace operator immediately after OLM by getting a `CVO run level` higher than OLM. I personally prefer to keep them order agnostic which is the way it is now.

kevinrizza:
@aravindhp Don't we want to move these files from the /deploy folder so we don't have duplicates?

aravindhp:
> Thinking out loud - immediately after deployment, reconciliation of the default `OperatorSource` CR(s) might fail if `CatalogSource` CRD is not registered (this might happen if CVO hasn't got to OLM yet in release payload). Although, with future retries it should eventually get reconciled.

Correct, CVO will keep retrying until it succeeds. Moreover the `CatalogSource` CRD gets registered before the `OperatorSource` as per the current ordering in OLM. Once we move that CR into Marketplace, we will fine as we ordering them correctly.

> As an option, we could deploy Marketplace operator immediately after OLM by getting a `CVO run level` higher than OLM. I personally prefer to keep them order agnostic which is the way it is now.

I don't see a reason for us to run immediately after OLM as long as we run at some point after OLM which is the case as our runlevel is 99 and OLM's is 30.

aravindhp:
> @aravindhp Don't we want to move these files from the /deploy folder so we don't have duplicates?

Won't that cause issues with the SDK for local development?

aravindhp:
@tkashem please add your lgtm if you are fine with latest set of changes.

galletti94:
FYI the UI changes required as a result of this PR are here https://github.com/openshift/console/pull/1027

PR #84, opened 2019-01-07
[opsrc] Add retry logic to opsrc reconciliation

tkashem:
On any reconciliation error we set the current phase to 'Failed'. The reconciliation of a Failed OperatorSource object is currently a no op. In order to retry a failed reconciliation, do the following:

- On error, make sure the object retains its current phase so that the next retry can resume from the current phase.
- Set 'CurrentPhase.Message' to the last error encountered so that  the object reflects the most recent error encountered.


tkashem:
@aravindhp enabled retry for purging. Retry won't help if the operator source has no manifest(s). Manual intervention is required for this situation to resolve - once the admin pushes new manifest(s) then reconciliation will be triggered by the registry sync loop.

kevinrizza:
Just had a thought about testing. Aren't there going to be test cases where we want to know the difference between a genuine failure and getting stuck in a retrying state? Maybe we should consider adding a shim that lets us force the operator into the failed state under certain conditions?

/lgtm

tkashem:
@aravindhp comment added

PR #85, opened 2019-01-07
[csc] Add retry logic to csc reconciliation

tkashem:
On any reconciliation error we set the current phase to 'Failed'. The reconciliation of a Failed CatalogSourceConfig object is currently a no op. In order to retry a failed reconciliation, do the following:

- On error, make sure the object retains its current phase so that the next retry can resume from the current phase.
- Set 'CurrentPhase.Message' to the last error encountered so that the object reflects the most recent error encountered.


tkashem:
@aravindhp yeah, addressed update.

PR #86, opened 2019-01-07
Allow specifying DisplayName and Provider

kevinrizza:
-Update spec to include displayname and publisher as optional fields
-Set default values when the fields aren't specified
-Update example cr to have the fields set

### DIFF
--- a/deploy/crds/catalogsourceconfig.crd.yaml
+++ b/deploy/crds/catalogsourceconfig.crd.yaml
@@ -29,6 +29,14 @@ spec:
     type: string
     description: Current status of the CatalogSourceConfig
     JSONPath: .status.currentPhase.phase.name
+  - name: DisplayName
+    type: string
+    description: Display Name for resulting CatalogSource
+    JSONPath: .spec.csDisplayName
+  - name: Publisher
+    type: string
+    description: Publisher for resulting CatalogSource
+    JSONPath: .spec.csPublisher
   - name: Message
     type: string
     description: Message associated with the current status
--- a/deploy/crds/operatorsource.crd.yaml
+++ b/deploy/crds/operatorsource.crd.yaml
@@ -29,6 +29,14 @@ spec:
     type: string
     description: App registry namespace
     JSONPath: .spec.registryNamespace
+  - name: DisplayName
+    type: string
+    description: Display (pretty) name to indicate the OperatorSource's name
+    JSONPath: .spec.displayName
+  - name: Publisher
+    type: string
+    description: Publisher for resulting to CatalogSource
+    JSONPath: .spec.csPublisher
   - name: Status
     type: string
     description: Current status of the OperatorSource
--- a/pkg/apis/marketplace/v1alpha1/catalogsourceconfig_types.go
+++ b/pkg/apis/marketplace/v1alpha1/catalogsourceconfig_types.go
@@ -35,6 +35,14 @@ type CatalogSourceConfig struct {
 type CatalogSourceConfigSpec struct {
 	TargetNamespace string `json:"targetNamespace"`
 	Packages        string `json:"packages"`
+
+	// DisplayName is passed along to the CatalogSource to be used
+	// as a pretty name.
+	DisplayName string `json:"csDisplayName,omitempty"`
+
+	// Publisher is passed along to the CatalogSource to be used
+	// to define what entity published the artifacts from the OperatorSource.
+	Publisher string `json:"csPublisher,omitempty"`
 }
 
 // CatalogSourceConfigStatus defines the observed state of CatalogSourceConfig
--- a/pkg/catalogsourceconfig/configuring.go
+++ b/pkg/catalogsourceconfig/configuring.go
@@ -207,11 +207,22 @@ func newConfigMap(csc *v1alpha1.CatalogSourceConfig, data map[string]string) *co
 
 // newCatalogSource returns a CatalogSource object.
 func newCatalogSource(csc *v1alpha1.CatalogSourceConfig, configMapName string) *olm.CatalogSource {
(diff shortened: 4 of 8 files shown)


kevinrizza:
Please note this pr will have to change once we get integrated with CVO (and also can't be easily tested until then, because of code that needs to be removed from OLM).

pkg/catalogsourceconfig/configuring.go
```diff
@@ -207,11 +207,22 @@ func newConfigMap(csc *v1alpha1.CatalogSourceConfig, data map[string]string) *co
 
 // newCatalogSource returns a CatalogSource object.
 func newCatalogSource(csc *v1alpha1.CatalogSourceConfig, configMapName string) *olm.CatalogSource {
+	// Set default display name when one is not set to Custom.
+	displayName := "Custom"
+	if csc.Spec.DisplayName != "" {
```
galletti94: Would this throw an error if there is no DisplayName field in the csc spec?

pkg/apis/marketplace/v1alpha1/catalogsourceconfig_types.go
```diff
@@ -35,6 +35,14 @@ type CatalogSourceConfig struct {
 type CatalogSourceConfigSpec struct {
 	TargetNamespace string `json:"targetNamespace"`
 	Packages        string `json:"packages"`
+
+	// DisplayName is passed along to the CatalogSource to be used
+	// to be used as a pretty name.
```
SamiSousa: Duplicate `to be used` here.

deploy/crds/catalogsourceconfig.crd.yaml
```diff
@@ -29,6 +29,14 @@ spec:
     type: string
     description: Current status of the CatalogSourceConfig
     JSONPath: .status.currentPhase.phase.name
+  - name: DisplayName
+    type: string
+    description: Display Name for resulting CatalogSource
+    JSONPath: .spec.displayName
+  - name: Publisher
+    type: string
+    description: Publisher for source propagated to CatalogSource
```
aravindhp: `Publisher for resulting CatalogSource`

deploy/crds/operatorsource.crd.yaml
```diff
@@ -29,6 +29,14 @@ spec:
     type: string
     description: App registry namespace
     JSONPath: .spec.registryNamespace
+  - name: DisplayName
+    type: string
+    description: Display Name for resulting CatalogSource
+    JSONPath: .spec.displayName
+  - name: Publisher
+    type: string
+    description: Publisher for source propagated to CatalogSource
```
aravindhp: `Publisher for resulting CatalogSource`

deploy/crds/catalogsourceconfig.crd.yaml
```diff
@@ -29,6 +29,14 @@ spec:
     type: string
     description: Current status of the CatalogSourceConfig
     JSONPath: .status.currentPhase.phase.name
+  - name: DisplayName
+    type: string
+    description: Display Name for resulting CatalogSource
+    JSONPath: .spec.displayName
```
aravindhp: Should be prefix it with `cs` i.e. `.spec.csDisplayName` to call out it is for the `CatalogSource`?

deploy/crds/catalogsourceconfig.crd.yaml
```diff
@@ -29,6 +29,14 @@ spec:
...
+  - name: Publisher
+    type: string
+    description: Publisher for source propagated to CatalogSource
+    JSONPath: .spec.publisher
```
aravindhp: Should be prefix it with `cs` i.e. `.spec.csPublisher` to call out it is for the `CatalogSource`?

deploy/crds/operatorsource.crd.yaml
```diff
@@ -29,6 +29,14 @@ spec:
     type: string
     description: App registry namespace
     JSONPath: .spec.registryNamespace
+  - name: DisplayName
+    type: string
+    description: Display Name for resulting CatalogSource
```
aravindhp: We could make this more generic as this can be used even if stopped using `CatalogSource` as our datastore.

kevinrizza: Does it make sense for `OperatorSource` to ever have a display name? I didn't think there was a 1:1 mapping between `OperatorSource` and anything in the ui. Would that ever change? It seems like an arbitrary abstraction from that perspective. Personally I think your argument to call all of these fields `csDisplayName` and `csPublisherName` behind the scenes since they only exist right now to pass along a value makes sense. I'm not as sold that we would need a real `DisplayName` for `OperatorSource`. Where would that be used?

aravindhp: You are correct in that they wouldn't be used anywhere today. I just wanted to future proof this once `OperatorSource` makes it's way into the UI. I also think @jeff-phillips-18 is working on a UI that allows adding of `OperatorSources`. Not sure if that is for 4.0 or the future.

jeff-phillips-18: Yes, we are adding a filter for the operator source type in the Operator Hub in console. Currently I am hard coding the display text for the operator source types.

tkashem: @jeff-phillips-18 are you referring to the `opsrc-provider` label the UI is using to filter `Community`, `ISV` and such?

kevinrizza:
This pr isn't necessary since the ui is filtering based off a tag, rather than using the displayname

aravindhp:
This PR was mainly to ensure that the display name on the `CatalogSources` maps to the `OperatorSource`. So this would still be required.

aravindhp:
Superseded by #90

PR #87, opened 2019-01-08
Add default operatorsource

kevinrizza:
-Include default operatorsource managed by CVO on cluster install
-Add default operatorsource yaml to manifests folder
-Reorder manifests to install default cr before deployment


kevinrizza:
WIP until this cr is removed from OLM

kevinrizza:
@aravindhp and I spoke offline. Basically I made that change because my opinion is that state should be completely created before the operator itself starts. That way if the operator comes up there isn't the case that not all of our default state exists yet in the cluster.

PR #88, opened 2019-01-08
[e2e] Stop creating Marketplace resources

aravindhp:
Now that we are managed by CVO, all the required Marketplace resources will be present in a cluster by default. This patch removes code that was creating these resources. Also removed downloading kubectl as that is not needed anymore.

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -1,21 +1,8 @@
 #!/bin/bash
 set -e
 
-TEST_NAMESPACE="openshift-operators"
-MANIFEST_FOLDER="./test/e2e/environment/"
-NAMESPACED_MANIFEST="./${MANIFEST_FOLDER}/namespaced-manifest.yaml"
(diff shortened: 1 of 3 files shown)


aravindhp:
@kevinrizza is `kubectl` required for the e2e tests? If not, I should remove installing that from `run-e2e-jobs.sh`.

tkashem:
Question: My understanding of the pipeline is limited. So the installer deploys marketplace off of the image built from latest master branch. And the test should be pointing to a deployment off of an image built from the PR branch. For my clarity, where is this happening?

aravindhp:
> Question: My understanding of the pipeline is limited. So the installer deploys marketplace off of the image built from latest master branch. And the test should be pointing to a deployment off of an image built from the PR branch. For my clarity, where is this happening?

The cluster used in the e2e uses a marketplace image that includes the bits from this PR and not from master. This is done as part of the CI pipeline.

PR #89, opened 2019-01-10
Update catalog when a new version is available

tkashem:
When a new version of an operator is pushed to quay.io, we want to
make the new version available to existing CatalogSource object(s)
so that user can upgrade to the new version.

Make the following change(s) to achieve this goal:

- Extend datastore to maintain a list of package(s) associated with
  each repository so that when a repository has a new version we
  can relate to the package(s)involved.
- Extend datastore to return more fine grained information on update:
    -- Return a list of repositories that have new versions.
    -- Return a list of repositories that have been removed
       from the registry namespace.
- Implement a sync loop that waits on a channel for package update
  notification. Once activated, it lists all existing
  CatalogSourceConfig object(s) [label: opsrc-datastore!=true]. for
  each CatalogSourceConfig listed it executes the following:
    --  Determine if any package specified in Spec.Packages has a
        new version.
    -- Determine if any package specified in Spec.Packages has been
       removed from registry.
    -- If any of the above is true, update the CatalogSourceConfig
       object accordingly to trigger a new reconciliation.
  Once reconciliation kicks off the corresponding CatalogSource and
  ConfigMap object(s) will have the latest package information.
  - Extend OperatorSource registry sync to notify the above sync loop
    with the new package update information.

TODO#
This approach has the following limitations:
- Once an update is detected OperatorSource and CatalogSourceConfig
  reconciliation take place concurrently. This may cause CatalogSourceConfig
  to ask for a package for which OperatorSource has not finished downloading
  and processing the new version yet. So CatalogSourceConfig will get the
  older version of the package and call it a day. Right now, we have a
  hard coded wait time before we before we send update notification to
  CatalogSourceConfig sync loop.
- If the update operation of CatalogSourceConfig object that needs to get
  the new version of a package fails, we don't retry and hence the object
  is left stale.

One way we can resolve this is by keeping track of the version of the
repository in spec of CatalogSourceConfig object. This way it
can eventually get the new version from datastore when it is
available by retrying.

-- Add e2e test(s).

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -9,6 +9,7 @@ import (
 
 	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/operator-framework/operator-marketplace/pkg/apis"
+	"github.com/operator-framework/operator-marketplace/pkg/catalogsourceconfig"
 	"github.com/operator-framework/operator-marketplace/pkg/controller"
 	"github.com/operator-framework/operator-marketplace/pkg/operatorsource"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
@@ -17,6 +18,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
+)
+
+const (
+	// TODO: resyncInterval is hardcoded to 1 hour now, it would have to be
+	// configurable on a per OperatorSource level.
+	resyncInterval = time.Duration(60) * time.Minute
+
+	initialWait                = time.Duration(10) * time.Minute
+	updateNotificationSendWait = time.Duration(10) * time.Minute
 )
 
 func printVersion() {
@@ -72,12 +82,11 @@ func main() {
 	log.Print("Starting the Cmd.")
 	stopCh := signals.SetupSignalHandler()
 
-	// TODO: resyncInterval is hardcoded to 1 hour now, it would have to be
-	// configurable on a per OperatorSource level.
-	initialWait, resyncInterval := time.Duration(10), time.Duration(60)
-	syncer := operatorsource.NewRegistrySyncer(mgr.GetClient(), initialWait, resyncInterval)
+	catalogSyncer := catalogsourceconfig.NewCatalogSyncer(mgr.GetClient(), initialWait)
+	registrySyncer := operatorsource.NewRegistrySyncer(mgr.GetClient(), initialWait, resyncInterval, updateNotificationSendWait, catalogSyncer)
 
-	go syncer.Sync(stopCh)
+	go registrySyncer.Sync(stopCh)
+	go catalogSyncer.Sync(stopCh)
 
 	// Start the Cmd
 	log.Fatal(mgr.Start(stopCh))
--- a/pkg/catalogsourceconfig/syncer.go
+++ b/pkg/catalogsourceconfig/syncer.go
@@ -1,0 +1,71 @@
+package catalogsourceconfig
+
+import (
+	"time"
+
+	"github.com/operator-framework/operator-marketplace/pkg/datastore"
+	log "github.com/sirupsen/logrus"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// NewCatalogSyncer returns a new instance of CatalogSyncer interface.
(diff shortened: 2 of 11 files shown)


tkashem:
@aravindhp @kevinrizza you can start looking at it now.

pkg/operatorsource/syncer.go
```diff
@@ -3,19 +3,29 @@ package operatorsource
...
+func NewRegistrySyncer(client client.Client, initialWait time.Duration, resyncInterval time.Duration, sender PackageUpdateNotificationSender) RegistrySyncer {
 	return &registrySyncer{
 		initialWait:    initialWait,
 		resyncInterval: resyncInterval,
-		poller:         NewPoller(client),
+		poller:         NewPoller(client, sender),
 	}
 }
 
+// PackageUpdateNotificationSender is an interface that wraps the Send methid.
```
tkashem: `method`

cmd/manager/main.go
```diff
@@ -75,9 +76,12 @@ func main() {
 	// TODO: resyncInterval is hardcoded to 1 hour now, it would have to be
 	// configurable on a per OperatorSource level.
 	initialWait, resyncInterval := time.Duration(10), time.Duration(60)
-	syncer := operatorsource.NewRegistrySyncer(mgr.GetClient(), initialWait, resyncInterval)
 
-	go syncer.Sync(stopCh)
+	catalogSyncer := catalogsourceconfig.NewCatalogSyncer(mgr.GetClient(), initialWait)
+	registrySyncer := operatorsource.NewRegistrySyncer(mgr.GetClient(), initialWait, resyncInterval, catalogSyncer)
```
kevinrizza: Do we want both of these to have the exact same initialWait? Wouldn't it be better to stagger them slightly? I think in the current implementation it shouldn't matter, but there could be some timing issues if these get more complex.

pkg/operatorsource/syncer.go
```diff
@@ -3,19 +3,29 @@ package operatorsource
...
+//
+// Send sends package update notification to the underlying channel that
+// CatalogSourceConfig is waiting on. This methid is expected to be a non
```
kevinrizza: Same comment as above, this is also misspelled and should be `method`

pkg/operatorsource/poller.go
```diff
@@ -45,28 +47,48 @@ type Poller interface {
...
+	if !aggregator.IsUpdatedOrRemoved() {
+		return
 	}
 
-	if !updated {
-		return nil
-	}
+	// TODO: This is a stop gap measure. We should not need this any longer when
+	// CatalogSourceConfig has the version stored.
+	<-time.After(time.Minute * 10)
```
kevinrizza: Can we make this 10 minute value a `const` somewhere else?

cmd/manager/main.go
tkashem: catalogSyncer waits for an event from registrySyncer, so it already is staggered in that regard. But the larger timing issue as we discussed earlier still remains with this approach.

pkg/operatorsource/poller.go
tkashem: moved to `main.go` so that all the duration consts are in one place.

tkashem:
@aravindhp @kevinrizza addressed the comments, can you give it a look?

pkg/catalogsourceconfig/syncer.go
```diff
@@ -0,0 +1,70 @@
...
+	// reconciling existing CR(s). Let's give the process a grace period to
+	// reconcile and rebuild the local cache from existing CR(s).
+	<-time.After(s.initialWait * time.Minute)
+
+	log.Info("[sync] CatalogSourceConfig sync loop has started")
+	for {
+		select {
+		case notification := <-s.notificationCh:
+			if notification == nil {
+				break
```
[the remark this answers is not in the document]
aravindhp: `cannot`. Shouldn't it be an error log?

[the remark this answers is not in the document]
tkashem: @aravindhp fixed.

tkashem:
@kevinrizza can you give a lgtm if it looks good.

cmd/manager/main.go
kevinrizza: fair

kevinrizza:
whoops, thought I already did.

/lgtm

tkashem:
@aravindhp @kevinrizza rebase done, can you apply `lgtm` labels when you have time?

tkashem:
@aravindhp @kevinrizza hold off on the `lgtm` label. I spoke too soon.

tkashem:
@aravindhp @kevinrizza all good, you can look now. I made sure unit tests and e2e tests pass locally.

PR #91, opened 2019-01-10
[deploy] Deploy operator to master node

aravindhp:
As per CVO guidelines, all the operators it manages must be deployed to the master node. The toleration is added to allow the pod onto any node with any taint.


tkashem:
How did you test it? :)
/lgtm

aravindhp:
>How did you test it?

I first deployed the operator to another namespace and then figured out the node it got deployed to using `oc get pods -o wide`. I then crossed checked it with `oc get machine -o wide --all-namespaces -o=custom-columns=NAME:.metadata.name,IP:.status.addresses | grep <node-ipaddress>` to confirm it was in a master node.

PR #94, opened 2019-01-15
[deploy] Fix image-references for OCP build

aravindhp:
Change tag name to "marketplace-operator" to match what the release build expects.


aravindhp:
The correct fix was in openshift/ocp-build-data#33

PR #100, opened 2019-01-31
Add reporting of operator health 

aravindhp:
- Add the ClusterOperator CR to the manifests
- Add support for reporting the health of the operator through the ClusterOperator CR
- Update dependencies

This addresses Jira issue [MKTPLC-222](https://jira.coreos.com/browse/MKTPLC-222)


pkg/status/status.go
```diff
@@ -0,0 +1,155 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const clusterOperatorName = "marketplace-operator"
+
+type status struct {
+	client          *configclient.ConfigV1Client
+	coAPINotPresent bool
+	namespace       string
+	co              *configv1.ClusterOperator
```
kevinrizza: Can we have a better name for this? I think `coAPINotPresent` is fine for the purpose of brevity, but I think it would help readability if this variable was called something like `clusterOperator`.

pkg/status/status.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// setStatus handles setting all the required fields for the given
+// ClusterStatusConditionType
+func (s *status) setStatus(condition configv1.ClusterStatusConditionType, message string) {
+	if s.coAPINotPresent {
+		return
+	}
+	log.Infof("Setting ClusterOperator condition: %s message: %s", condition, message)
+	s.ensureClusterOperator()
```
kevinrizza: Why run `ensureClusterOperator()` here every time we set the status, rather than once on first startup? It seems a little like unnecessary overhead. I can see the value in some cases, but in ours I would imagine that we just have a single status object that persists and shouldn't lose this reference.

pkg/status/status.go
```diff
@@ -0,0 +1,155 @@
...
+		availableStatus = configv1.ConditionTrue
+		availableMessage = message
+
+	case configv1.OperatorFailing:
+		failingStatus = configv1.ConditionTrue
+		failingMessage = message
+	}
+
+	time := v1.Now()
+	// As per the docs, all three StatusConditionTypes needs to be set with only
```
kevinrizza: This comment may be hard to understand for some. What docs are you referring to in this case? Can we link to them?

pkg/status/status.go
aravindhp: I am not sure if CVO modifies the `ClusterOperator` CR. So to be on the safe side I am getting every time we set the status. I see this pattern being followed in the other operators too.

pkg/status/status.go
aravindhp: Done

pkg/status/status.go
aravindhp: Done. I have to say that even the docs don't clearly call this out. I found this out from a slack channel :-)

pkg/status/status.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// New returns an initialized Status
+func New(client *configclient.ConfigV1Client, namespace string) Status {
+	// Check if the ClusterOperator API is present on the cluster. This is so
+	// that we can continue to work with Kubernetes clusters that don't have
+	// this API.
+	_, err := client.ClusterOperators().Get(clusterOperatorName, v1.GetOptions{})
+	coAPINotPresent := false
+	if meta.IsNoMatchError(err) {
```
aravindhp: @kevinrizza @tkashem I discovered that the error returned when `ClusterOperator` CRD is not present is a generic not found error. This won't help us on vanilla kube. So instead I am going to push a change where I look for the `ClusterOperator` CRD itself. The downside is that I need to add the CRD scheme to the manager. Let me know what you think after I push.

kevinrizza:
@aravindhp Are we saying that we always need to install the ClusterOperator even on Openshift install? I thought it was just needed for vanilla kube? Are other openshift components following this pattern?

aravindhp:
I might be misunderstanding the context of your question. I thought it was understood that this is needed only for OpenShift and not for vanilla kube. `ClusterOperator` is a CRD present only on OpenShift clusters.

kevinrizza:
@aravindhp Sorry maybe I was unclear. I was under the impression that the latest change needed to be made specifically because of that (i.e. ClusterOperator is only available on openshift clusters, so we need to add some bits so that we don't start thowing errors when running on vanilla kubernetes. Is that not true?

aravindhp:
> @aravindhp Sorry maybe I was unclear. I was under the impression that the latest change needed to be made specifically because of that (i.e. ClusterOperator is only available on openshift clusters, so we need to add some bits so that we don't start thowing errors when running on vanilla kubernetes. Is that not true?

That is correct. And it is just not throwing errors but to be able to work on an vanilla kube cluster. However your statement in your original comment confused me:

>Are we saying that we always need to install the ClusterOperator even on Openshift install?

kevinrizza:
Ah, yeah I realize that's unclear. Sorry. What I mean is, we have updated the manifest folder with specific references to the `ClusterOperator`. Right now, the only way to install marketplace is to apply that manifest folder. If we add those bits, that means that both the Openshift and vanilla kube will include those. Is that what we want?

pkg/status/status.go
```diff
@@ -0,0 +1,183 @@
...
+func New(cfg *rest.Config, mgr manager.Manager, namespace string) Status {
+	// The default client serves read requests from the cache which contains
+	// objects only from the namespace the operator is watching. Given we need
+	// to query CRDs which are cluster wide, we add the CRDs to the manager's
+	// scheme and create our own client.
+	if err := v1beta1.AddToScheme(mgr.GetScheme()); err != nil {
+		log.Fatal(err)
+	}
+
+	controllerClient, err := client.New(cfg, client.Options{Scheme: mgr.GetScheme()})
```
tkashem: If you use an openshift client `configclient.NewForConfig(cfg)` you don't see the expected `IsNoMatchError`?

aravindhp:
The current `manifest` folder is meant only for OpenShift. As part of adding artifacts for upstream kubernetes, we would create a directory with upstream specific artifacts. We would also rename the namespaces in those files.

aravindhp:
@kevinrizza @tkashem I figured out that the cluster role was not updated to read CRDs. I have fixed that issue and updated some logs. Please review.

PR #101, opened 2019-01-31
Service Catalog: aggregated roles & service

jboyd01:
these roles and service resources can not be processed as part of the CSV.  Until we have support to put these within Service Catalog manifests and created when the CSV is installed, these need to be created when Marketplace is installed.

### DIFF
--- a/manifests/12_servicecatalog.yaml
+++ b/manifests/12_servicecatalog.yaml
@@ -1,0 +1,130 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: system:service-catalog:aggregate-to-admin
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+rules:
+- apiGroups:
+  - "servicecatalog.k8s.io"
+  attributeRestrictions: null
+  resources:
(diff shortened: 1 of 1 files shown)


jboyd01:
Need to hold until we have explicit approval for Service Catalog being approved for 4.0
/hold

manifests/12_servicecatalog.yaml
```diff
@@ -0,0 +1,130 @@
...
+metadata:
+  name: system:service-catalog:aggregate-to-edit
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+rules:
+- apiGroups:
+  - "servicecatalog.k8s.io"
+  attributeRestrictions: null
+  resources:
+  - servicebrokers
```
djzager: do you not need `clusterservicebrokers` or does `servicebrokers` catch both?

jboyd01: Right, non cluster admin uses don't need access to _clusterservicebrokers_ and this role is excluding clusterservicebrokers on purpose.  Glad to have the additional eyes on this though - thanks!

jboyd01:
no longer relevant, Service Catalog is installed by cluster operator now

PR #103, opened 2019-02-05
[opsrc] Fix OperatorSource reconciliation error

tkashem:
OperatorSource reconciler fails in 'Configuring' phase if the
following conditions are true:
 - The CatalogSourceConfig object already exists.
 - It is already owned by the OperatorSource object.

The reconciler appends the OperatorSource object to the
OwnerReference list of the CatalogSourceConfig without checking
whether it is already owned by the same OperatorSource. This causes
the Update API to throw an error.

Solution: Remove the given OperatorSource object from the
OwnerReference list to ensure it is not specified more than once.

Bugzilla link: https://bugzilla.redhat.com/show_bug.cgi?id=1672701.

### DIFF
--- a/pkg/operatorsource/configuring.go
+++ b/pkg/operatorsource/configuring.go
@@ -95,6 +95,12 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *v1alpha1.Oper
 	}
 
 	cscExisting.EnsureGVK()
+
+	// The existing CatalogSourceConfig might already be owned by this
+	// OperatorSource object. Let's remove the owner reference, otherwise we
+	// will be adding it twice.
+	cscExisting.RemoveOwner(in.GetUID())
+
 	builder := CatalogSourceConfigBuilder{object: cscExisting}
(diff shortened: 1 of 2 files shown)


tkashem:
@kevinrizza @aravindhp take a look when you have time.

pkg/operatorsource/configuring.go
```diff
@@ -95,6 +95,8 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *v1alpha1.Oper
 	}
 
 	cscExisting.EnsureGVK()
+	cscExisting.RemoveOwner(in.GetUID())
```
aravindhp: Please add a comment mentioning why you are removing the owner. Looks good to me otherwise.

tkashem: @aravindhp done

tkashem:
@aravindhp @kevinrizza updated the commit with bugzilla link, can I have `lgtm`?

PR #104, opened 2019-02-06
Add upstream deployment

kevinrizza:


deploy/upstream/0.1.0/02_catalogsourceconfig.crd.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
tkashem: Nit: Since the user on vanilla k8s will also install `OLM`, should we follow the same directory layout? Right now, OLM has this layout `deploy/upstream/manifests/{version}`?

aravindhp: I think we can follow our own layout as long as we document it. 

For upstream, what we should we be doing is creating a `CatalogSource` with Marketplace in it. Add it to the OLM repo along with a `Subscription` so that we get installed automatically along with OLM. That way it is a one shot process. I am not saying this needs to be done now but can be done post 4.0.

deploy/upstream/0.1.0/07_operator.yaml
```diff
@@ -0,0 +1,42 @@
...
+      name: marketplace-operator
+  template:
+    metadata:
+      labels:
+        name: marketplace-operator
+    spec:
+      serviceAccountName: marketplace-operator
+      containers:
+        - name: marketplace-operator
+          image: quay.io/openshift/origin-operator-marketplace:latest
```
aravindhp: If you are saying that the upstream version is `0.1.0` then this would need to be tagged. So we should either remove the version from the directory layout or figure out a way to tag the image.

kevinrizza: If we don't care specifically about versioning, lets just remove it.

aravindhp: I am fine with that for now. But I am guessing we would have to introduce versioning at some point for upstream.

kevinrizza: Let's cross that bridge when we get there.

kevinrizza:
@aravindhp @tkashem Removed versioning, please take a look again. Thanks

PR #106, opened 2019-02-12
Generate typed clientset for marketplace type(s)

tkashem:
- Generate typed clientset for marketplace type(s) so that other
  components can access marketplace APIs.
- We will use the typed client to query OperatorSource
  object(s) from operator-registry pod.

https://jira.coreos.com/browse/MKTPLC-108
https://jira.coreos.com/browse/MKTPLC-244

TODO:
- Add a make goal `make gen-clientset` that will generate the
  clientset using client-gen.


pkg/apis/marketplace/v1alpha1/catalogsourceconfig_types.go
```diff
@@ -13,15 +13,8 @@ const (
 	CSCFinalizer = "finalizer.catalogsourceconfigs.marketplace.redhat.com"
 )
 
-// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
tkashem: I had to move this for `client-gen` to work. I ran `operator-sdk generate k8s` after this and it did not throw any error or produce any changes to the typeset.

PR #107, opened 2019-02-12
Use grpc CatalogSource

aravindhp:
"internal" CatalogSources uses ConfigMaps to store operator artifacts. This has the downside of running into Etcd space limits of 1MB. To overcome that we are moving to using "grpc" CatalogSources that are backed by a sqlite database that is exposed using a grpc interface from a pod. To achieve this the following changes have been made:

- Refactor the OperatorSource datastore to only download the meta-data for the packages stored in Quay app-registries
- Add a registry builder to CatalogSourceConfig that creates all the necessary resources for the grpc pod.

This should be merged only after https://github.com/operator-framework/operator-registry/pull/27 and https://github.com/operator-framework/operator-lifecycle-manager/pull/684

TODO before merging:

- [ ] Update vendor commit after operator-framework/operator-lifecycle-manager/pull/684 has merged
- [x] Remove `ConfigMap` from registry `Role`
- [x] Replace the test image with the actual image


pkg/catalogsourceconfig/clusterrolebindingbuilder.go
```diff
@@ -0,0 +1,66 @@
...
+	b.crb.SetName(name)
+	return b
+}
+
+// WithOwner sets the owner of the ClusterRoleBinding object to the given owner.
+func (b *ClusterRoleBindingBuilder) WithOwner(owner *v1alpha1.CatalogSourceConfig) *ClusterRoleBindingBuilder {
+	b.crb.SetOwnerReferences(append(b.crb.GetOwnerReferences(),
+		[]meta.OwnerReference{
+			*meta.NewControllerRef(owner, owner.GroupVersionKind()),
+		}[0]))
```
kevinrizza: nit: Why make a slice of meta.OwnerReference and then immediately reference the first element of it? Why not just initialize the NewControllerRef directly?

pkg/catalogsourceconfig/clusterrolebuilder.go
```diff
@@ -0,0 +1,60 @@
...
+	b.clusterRole.SetName(name)
+	return b
+}
+
+// WithOwner sets the owner of the ClusterRole object to the given owner.
+func (b *ClusterRoleBuilder) WithOwner(owner *v1alpha1.CatalogSourceConfig) *ClusterRoleBuilder {
+	b.clusterRole.SetOwnerReferences(append(b.clusterRole.GetOwnerReferences(),
+		[]meta.OwnerReference{
+			*meta.NewControllerRef(owner, owner.GroupVersionKind()),
+		}[0]))
```
kevinrizza: same as above.

pkg/catalogsourceconfig/configuring.go
```diff
@@ -21,11 +20,16 @@ import (
 // TBD: Vendor in the folder once we require more than just these constants from
 // the OLM registry code.
```
kevinrizza: nit: This comment seems to be disingenuous since we are adding to the const list. We don't need to vendor, but maybe the comment should be updated?

deploy/operator.yaml
```diff
@@ -28,7 +28,9 @@ spec:
             name: healthz
           command:
           - marketplace-operator
-          imagePullPolicy: Always
+          args:
+          - -registryServerImage=quay.io/openshift/origin-operator-registry
+          imagePullPolicy: IfNotPresent
```
kevinrizza: This change to image pull policy will be removed once we update the image right?

pkg/catalogsourceconfig/clusterrolebindingbuilder.go
aravindhp: `ClusterRoles` and `ClusterRoleBindings` are no longer needed. I forgot to remove the files. I will do so now.

deploy/operator.yaml
aravindhp: I don't follow. This is the correct image. Why would would we change the pull policy?

pkg/catalogsourceconfig/configuring.go
aravindhp: Actually the constants need to be removed :-)

deploy/operator.yaml
kevinrizza: Sorry didn't look at the uri. I had thought that we were using a temp image until Abu's pr was pushed built and available in quay. Looks like that happened.

So what I'm not sure about here is why are we changing the imagepullpolicy? Don't we want to always get the latest when this gets applied?

aravindhp: You are right. I will fix that here. I also realize I need to update the upstream manifests. :-(

aravindhp:
@kevinrizza I have addressed your concerns. Please review.

pkg/catalogsourceconfig/configuring.go
```diff
@@ -148,45 +124,20 @@ func (r *configuringReconciler) reconcileCatalogSource(csc *v1alpha1.CatalogSour
...
-		Name:      configMapName,
-		Namespace: csc.Spec.TargetNamespace,
+// checkPackages returns an error if there no valid packages present in the
+// datastore.
+func (r *configuringReconciler) checkPackages(csc *v1alpha1.CatalogSourceConfig) error {
+	atLeastOneFound := false
+	packageIDs := GetPackageIDs(csc.Spec.Packages)
+	for _, packageID := range packageIDs {
+		if _, err := r.reader.Read(packageID); err != nil {
+			continue
```
tkashem: Suggestion: It would be useful to log the error(s) we run into. `NewAggregate` function from   `k8s.io/apimachinery/pkg/util/errors` can be used to return aggregated error(s), if that works for you.

test/e2e/marketplace_test.go
```diff
@@ -132,27 +134,31 @@ func defaultCreateTest(t *testing.T, f *test.Framework, ctx *test.TestCtx) error
...
-		)
+	// Then check that the deployment was created.
+	resultDeployment := &apps.Deployment{}
+	err = WaitForResult(t, f, resultDeployment, namespace, deploymentName)
+	if err != nil {
+		return err
+	}
+
+	// Now check that the deployment is ready.
+	err = WaitForSuccessfulDeployment(t, f, *resultDeployment)
```
tkashem: In an asynchronous world, we would return as soon as we successfully create the Deployment object. We can watch the `pods` and other relevant resources with appropriate labels to get to the desired state. We can address this in the next iteration.

pkg/datastore/datastore.go
```diff
@@ -125,70 +113,63 @@ type Writer interface {
...
+		return
 	}
 
-	ds.rows.Add(opsrc, repositories, operators)
-
-	count = len(operators)
-	err = utilerrors.NewAggregate(allErrors)
+	for _, repo := range repositories {
+		if repo.Package == pkg {
+			repository = repo
```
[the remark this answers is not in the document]
tkashem: @kevinrizza ^

pkg/catalogsourceconfig/deploymentbuilder.go
```diff
@@ -0,0 +1,59 @@
...
+		Kind:       "Deployment",
+		APIVersion: "apps/v1",
+	}
+	return b
+}
+
+// WithMeta sets basic TypeMeta and ObjectMeta.
+func (b *DeploymentBuilder) WithMeta(name, namespace string) *DeploymentBuilder {
+	b.WithTypeMeta()
+	if b.deployment.GetObjectMeta() == nil {
```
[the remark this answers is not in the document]
aravindhp: I ran into an issue with `CatalogSource` builder that caused me to add this. I can't recollect the issue now. We can look into this further post beta 2 and remove it.

pkg/catalogsourceconfig/registry.go
```diff
@@ -0,0 +1,462 @@
...
+		r.log.Infof("Created ServiceAccount %s", serviceAccount.GetName())
+	} else {
+		r.log.Infof("ServiceAccount %s is present", serviceAccount.GetName())
+	}
+	return nil
+}
+
+// getCommand returns the command used to launch the registry server
+func (r *registry) getCommand(packages string, sources string) []string {
+	return []string{"appregistry-server", "debug", "-t", "termination.log", "-s", sources, "-o", packages}
```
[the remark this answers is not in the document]
aravindhp: I have removed this already

pkg/datastore/datastore.go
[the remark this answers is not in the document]
kevinrizza: Yeah it makes sense to add a break there.

pkg/catalogsourceconfig/registry.go
```diff
@@ -0,0 +1,369 @@
...
+// ensureService ensure that the Service for the registry deployment is present.
+func (r *registry) ensureService() error {
+	service := new(ServiceBuilder).WithTypeMeta().Service()
+	// Delete the Service so that we get a new ClusterIP
+	if err := r.client.Get(context.TODO(), r.csc.key(), service); err == nil {
+		r.log.Infof("Service %s is present", service.GetName())
+		err := r.client.Delete(context.TODO(), service)
+		if err != nil {
+			r.log.Errorf("Failed to delete Service %s", service.GetName())
+			// Make a best effort to create the service
```
tkashem: If it can't delete, then the logs will say `failed to delete` and then say `deleted the object`.

pkg/catalogsourceconfig/rolebuilder.go
```diff
@@ -0,0 +1,62 @@
...
+	return b
+}
+
+// NewRule returns PolicyRule objects
+func NewRule(verbs, apiGroups, resources, resourceNames []string) rbac.PolicyRule {
+	return rbac.PolicyRule{
+		Verbs:         verbs,
+		APIGroups:     apiGroups,
+		Resources:     resources,
+		ResourceNames: resourceNames,
```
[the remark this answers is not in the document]
aravindhp: I am just being extra careful and narrowing the scope.

pkg/catalogsourceconfig/registry.go
aravindhp: I will fix this

PR #109, opened 2019-02-13
Use secret in appregistry

kevinrizza:
-Update operatorsource spec to include secret ref
-Update appregistry to pass token in auth header
-Update downloading reconciler to pass secret to appregistry
-Update poller to pass secret to appregistry
-Query for secret when setting up appregistry instance


pkg/apis/marketplace/v1alpha1/operatorsource_types.go
```diff
@@ -13,19 +13,6 @@ const (
...
-
-// OperatorSource is the Schema for the operatorsources API
-// +k8s:openapi-gen=true
-type OperatorSource struct {
-	metav1.TypeMeta   `json:",inline"`
-	metav1.ObjectMeta `json:"metadata,omitempty"`
-	Spec              OperatorSourceSpec   `json:"spec,omitempty"`
-	Status            OperatorSourceStatus `json:"status,omitempty"`
-}
-
```
tkashem: Same question: did you have to move it to get `operatorsdk k8s generate` to work? I don't see the `// +genclient` tag that was added.

pkg/operatorsource/registryhelper.go
```diff
@@ -0,0 +1,37 @@
...
+		key := k8sclient.ObjectKey{
+			Name:      auth.SecretName,
+			Namespace: auth.SecretNamespace,
+		}
+		err := client.Get(context.TODO(), key, &secret)
+		if err != nil {
+			return options, err
+		}
+
+		options.AuthToken = string(secret.Data["token"])
```
tkashem: The auth token is already opaque, wondering whether we need to base64 encode it before we store it in the Secret object.

pkg/apis/marketplace/v1alpha1/operatorsource_types.go
kevinrizza: @tkashem I didn't use the sdk to generate it, I just pulled your commit down. I can undo this and reapply it using the sdk if needed.

pkg/operatorsource/registryhelper.go
kevinrizza: @tkashem Is there a way to store it as a simple string without the encoding? I'm not too familiar with kube secrets, but my understanding is that the only real user data supported type is Opaque. Are you saying we should use the 'StringData' field to save the token in plain text? Won't that be harder to define later?

kevinrizza:
This was replaced by changes in another pr. Closing as obsolete.

PR #114, opened 2019-02-21
Fix update sync with Quay

aravindhp:
When there are update present in the Quay namespace but no new packages are being added or removed, the updates are not getting pulled down. This was due to the deployment spec for the registry pod being updated only when the list of packages changes. This PR fixes that issue.

https://bugzilla.redhat.com/show_bug.cgi?id=1678906

### DIFF
--- a/pkg/catalogsourceconfig/registry.go
+++ b/pkg/catalogsourceconfig/registry.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/operator-framework/operator-marketplace/pkg/apis/marketplace/v1alpha1"
 	"github.com/operator-framework/operator-marketplace/pkg/datastore"
@@ -14,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -91,8 +93,8 @@ func (r *registry) GetAddress() string {
 	return r.address
 }
 
-// ensureDeployment ensures that registry Deployment is present that for serving
-// a database consisting of the packages from the given operatorSources
+// ensureDeployment ensures that registry Deployment is present for serving
+// the the grpc interface for the packages from the given operatorSources
 func (r *registry) ensureDeployment(operatorSources string) error {
 	registryCommand := getCommand(r.csc.Spec.Packages, operatorSources)
 	deployment := new(DeploymentBuilder).WithTypeMeta().Deployment()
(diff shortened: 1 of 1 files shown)


pkg/catalogsourceconfig/registry.go
```diff
@@ -104,15 +106,51 @@ func (r *registry) ensureDeployment(operatorSources string) error {
...
+		// There should only be one pod in the Deployment. If we find more
+		// than one we will delete the pod whose name starts with the
+		// CatalogSourceConfig's name
+		if len(pods.Items) > 1 {
+			r.log.Warnf("Found %d in Deployment %s. Will delete most likely match!",
+				len(pods.Items), deployment.GetName())
+		}
+
+		for _, pod := range pods.Items {
+			if !strings.HasPrefix(pod.GetName(), r.csc.GetName()) {
```
kevinrizza: Is this convention guaranteed when creating pods? Isn't it possible that the naming could change? Maybe it would be safer to use some unique identifier, like a label.

pkg/catalogsourceconfig/registry.go
```diff
@@ -104,15 +106,51 @@ func (r *registry) ensureDeployment(operatorSources string) error {
...
+				r.log.Warnf("Found %s in Deployment %s that did not match", pod.GetName(), deployment.GetName())
+				continue
+			}
+			if err = r.client.Delete(context.TODO(), &pod); err != nil {
+				msg := fmt.Sprintf("Failed to delete Pod %s: %v", pod.GetName(), err)
+				r.log.Errorf(msg)
+				r.log.Errorf("%s will not be updated", r.csc.GetName())
```
kevinrizza: Can we combine these into one error message?

pkg/catalogsourceconfig/registry.go
```diff
@@ -369,3 +413,16 @@ func getRules(operatorSources []string) []rbac.PolicyRule {
 		NewRule([]string{"get"}, []string{""}, []string{"secrets"}, nil),
 	}
 }
+
+// hasCommandChanged returns true if the new command is different from the old
+func hasCommandChanged(old, new []string) bool {
+	if len(old) != len(new) {
```
kevinrizza: Couldn't this result in a false positive? For example if one operator was switched out with another of the same length?

pkg/catalogsourceconfig/registry.go
aravindhp: Yeah there is no guarantee. I am using a label selector otherwise deployments won't work. I am just being paranoid that there could be another pod with the same label not created by us. Maybe I am being paranoid. Let me look into a better approach. Maybe scale down and scale up the deployment. It is just a pain to do that given we are not using `client-go`.

pkg/catalogsourceconfig/registry.go
aravindhp: Sure, it would be become very verbose and hard to read though.

pkg/catalogsourceconfig/registry.go
aravindhp: I am checking for `!=`. If I was using `==` and returning false then what you say is true.

aravindhp:
@kevinrizza @tkashem I have updated the PR. Please review.

pkg/catalogsourceconfig/registry.go
```diff
@@ -105,10 +105,27 @@ func (r *registry) ensureDeployment(operatorSources string) error {
...
+			r.log.Errorf("Failed to get updated Deployment %s : %v", deployment.GetName(), err)
+			return err
+		}
+
+		replicas = 1
+		deployment.Spec.Replicas = &replicas
 		deployment.Spec.Template = r.newPodTemplateSpec(registryCommand)
-		err = r.client.Update(context.TODO(), deployment)
-		if err != nil {
+		if err = r.client.Update(context.TODO(), deployment); err != nil {
```
tkashem: Because we are making two updates, one happening immediately after the other. There is a chance that the controller-manager will be informed of the latest version ( which has `Replicas` set to 1, the original value ) and may not result in pod recreation. This is my understanding based on how level based system works.  I could be wrong, but just wanted to bring this up.

aravindhp: I initially wrote this code with a wait for the deployment to scale down. I can go back to doing that assuming that's what you're asking for.

tkashem: Ideally we can watch pods with specific labels and do things asynchronously but we don't have time for this now, we can take a holistic approach in the future. 
For now, what you are suggesting looks to be a good stop-gap measure.

aravindhp: Not sure how what we are trying to do here can be achieved by watching the pods. I would watch the deployment. Anyhow, I have https://jira.coreos.com/browse/MKTPLC-266 to track this.

aravindhp: @tkashem @kevinrizza I have addressed this. Please review.

pkg/catalogsourceconfig/registry.go
```diff
@@ -356,6 +375,34 @@ func (r *registry) newServiceSpec() core.ServiceSpec {
...
 
+// waitForDeploymentScaleDown waits for the deployment to scale down to zero within the timeout duration.
+func (r *registry) waitForDeploymentScaleDown(retryInterval, timeout time.Duration) (*apps.Deployment, error) {
+	deployment := apps.Deployment{}
+	err := wait.Poll(retryInterval, timeout, func() (done bool, err error) {
+		err = r.client.Get(context.TODO(), r.csc.key(), &deployment)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				r.log.Infof("Waiting for availability of %s Deployment", deployment.GetName())
+				return false, nil
```
tkashem: Question: If the deployment object is not found is it worth waiting?

aravindhp: This is from the SDK code. I assume this won't happen in our case. But I see no harm in leaving this in.

aravindhp: But I am open to removing it too.

aravindhp: Fixed.

PR #115, opened 2019-02-21
[csc] Handle invalid packages

kevinrizza:
-Throw error when csc spec references any package missing from datastore
-Set custom error on status and retry
-Fix for https://bugzilla.redhat.com/show_bug.cgi?id=1679405


kevinrizza:
> Please add the bugzilla link to the PR and commit message. LGTM otherwise.

@aravindhp Updated.

PR #117, opened 2019-02-25
[opsrc] Do not delete csc during purge

tkashem:
On restart, marketplace operator deletes and recreates the
CatalogSourceConfig object(s) associated with each OperatorSource.

These are CatalogSourceConfig object(s) having the following label:
      opsrc-datastore: "true"

Resolution:
- Do not delete CatalogSourceConfig object(s) during purge, instead
let the CatalogSourceConfig(s) be updated during the reconciliation
process  ( in configuring phase ).

- Make the log a bit clearer by mentioning whether it is creating
  or updating a CatalogSource object in 'Configuring' phase.

https://jira.coreos.com/browse/MKTPLC-236
https://bugzilla.redhat.com/show_bug.cgi?id=1682928

### DIFF
--- a/pkg/operatorsource/purging.go
+++ b/pkg/operatorsource/purging.go
@@ -7,7 +7,6 @@ import (
 	"github.com/operator-framework/operator-marketplace/pkg/datastore"
 	"github.com/operator-framework/operator-marketplace/pkg/phase"
 	log "github.com/sirupsen/logrus"
-	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/operatorsource/purging.go
```diff
@@ -56,16 +55,6 @@ func (r *purgingReconciler) Reconcile(ctx context.Context, in *v1alpha1.Operator
 
 	r.datastore.RemoveOperatorSource(in.GetUID())
 
-	builder := &CatalogSourceConfigBuilder{}
-	csc := builder.WithTypeMeta().
-		WithNamespacedName(in.Namespace, in.Name).
-		CatalogSourceConfig()
-
-	if err = r.client.Delete(ctx, csc); err != nil && !k8s_errors.IsNotFound(err) {
```
aravindhp: >- Do not delete CatalogSourceConfig object(s) during purge, instead let the CatalogSourceConfig(s) be updated during the reconciliation process.

I don't follow how the `CatalogSourceConfig` will get updated during the reconciliation process given it's phase will remain at `Succeeded`. Shouldn't you force an update so that we get the current state from Quay? Or are you saying that it will get updated during the next sync loop?

tkashem: The `CatalogSourceConfig` object will get updated during the 'Configuring' phase.

kevinrizza: Right this makes sense. Basically the Purging phase, rather than deleting the object, just resets the status and then sets the next phase to Initial. That way the configuring reconciler just updates the already existing object and it should work.

pkg/operatorsource/purging.go
```diff
@@ -56,16 +55,6 @@ func (r *purgingReconciler) Reconcile(ctx context.Context, in *v1alpha1.Operator
...
-		nextPhase = phase.GetNextWithMessage(phase.OperatorSourcePurging, err.Error())
-		return
-	}
```
kevinrizza: Can you update the comments in this file to reflect that we are no longer actually "purging" the operator source object itself?

pkg/operatorsource/purging.go
aravindhp: Ah yes, sorry I missed the bit of existing code below. Make sense now.

pkg/operatorsource/purging.go
aravindhp: This should also be reflected in log statement which says `Purging all resource(s)`

tkashem: Updated the log statement to clarify what has been purged.

PR #121, opened 2019-02-28
Bug 1680070: Allow specifying spaces in between packages

aravindhp:
A CatalogSourceConfig spec should allow a valid set of packages to be specified with spaces in between them. For example: foo, bar, jazz


aravindhp:
Will merge after beta2 to be on the safe side

PR #123, opened 2019-03-01
Bug 1683422: [csc] Make CLI output less verbose

### LINKED ISSUE #116: CatalogSourceConfig `kubectl` Output Difficult to Read
### Description

Having [seven items in the `additionalPrinterColumns`](https://github.com/operator-framework/operator-marketplace/blob/master/deploy/crds/catalogsourceconfig.crd.yaml#L19) for the `CatalogSourceConfig` CRD makes the rendered output from `kubectl get` difficult to read on even wide terminals (see screenshots). Consider reducing the number of fields being displayed. Changing `spec.packages` from being a comma-separated string to a proper array would likely help as well.

### Screenshots

**Viewing `CatalogSourceConfigs`:**
screenshot_20190225_135813

**Viewing `Pods`:**

screenshot_20190225_135919

anik120:
- Fixes #116
- Removed `TARGETNAMESPACE`, `PACKAGES`, `DISPLAYNAME` and `PUBLISHER` fields from output
- Modified catalogsourceconfig.crd to remove fields from `additionalPrinterColumns`

### DIFF
--- a/deploy/crds/catalogsourceconfig.crd.yaml
+++ b/deploy/crds/catalogsourceconfig.crd.yaml
@@ -17,26 +17,10 @@ spec:
   scope: Namespaced
   version: v1alpha1
   additionalPrinterColumns:
-  - name: TargetNamespace
-    type: string
-    description: The namespace where the operators will be enabled
-    JSONPath: .spec.targetNamespace
-  - name: Packages
-    type: string
-    description: List of operator(s) which will be enabled in the target namespace
-    JSONPath: .spec.packages
   - name: Status
     type: string
     description: Current status of the CatalogSourceConfig
     JSONPath: .status.currentPhase.phase.name
-  - name: DisplayName
-    type: string
-    description: Display Name for resulting CatalogSource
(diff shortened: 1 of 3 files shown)


kevinrizza:
Also please update the commit message to include what actually changed (i.e. a reference or description of what changed and why).

kevinrizza:
@aravindhp I think since the purpose of those fields isn't purposefully well documented we can remove them. But in general I think that identifiers like a `DisplayName` would be useful to leave. If we revisit the Display Name as something more robust and for something other than just OpenShift ui, we should add it back in.

aravindhp:
Sounds good @kevinrizza 

@anik120 please remove it for now

anik120:
@aravindhp @kevinrizza just to confirm, we also want the `DISPLAYNAME` and `PUBLISHER` fields removed.

kevinrizza:
@anik120 Yes that is correct.

anik120:
@aravindhp @kevinrizza PTAL now.

PR #125, opened 2019-03-05
Bug 1685066: Fix OperatorSource sync

aravindhp:
**Problem**
Prior to f9d37cc029, as part of syncing `OperatorSources`, the associated `CatalogSourceConfigs` were being deleted and recreated. This always ensured that we got the latest state from Quay. With that change we moved to updating `CatalogSourceConfigs`. Now in the scenario where a Quay namespace has changed without app-registry repositories being added or removed but with existing repositories being updated, these updates will not be reflected on the cluster.

**Solution**
The right fix for this would be for the `CatalogSourceConfig` cache to become app-registry version aware and to detect that updates are required not just based on spec changes. In the short term an easy fix is to drop the `CatalogSourceConfig` status before updating.

This fixes [bug 1685066](https://bugzilla.redhat.com/show_bug.cgi?id=1685066)

### DIFF
--- a/pkg/operatorsource/configuring.go
+++ b/pkg/operatorsource/configuring.go
@@ -102,6 +102,12 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *v1alpha1.Oper
 		WithOwnerLabel(in).
 		CatalogSourceConfig()
 
+	// Drop the status to force a CatalogSourceConfig update. This is to account
+	// for the the scenario where a Quay namespace has changed without
+	// app-registry repositories being added or removed but with existing
+	// repositories being updated.
+	cscUpdate.Status = v1alpha1.CatalogSourceConfigStatus{}
+
 	err = r.client.Update(ctx, cscUpdate)
 	if err != nil {
 		r.logger.Errorf("Unexpected error while updating CatalogSourceConfig: %s", err.Error())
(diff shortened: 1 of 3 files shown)


pkg/operatorsource/configuring.go
```diff
@@ -102,6 +102,12 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *v1alpha1.Oper
 		WithOwnerLabel(in).
 		CatalogSourceConfig()
 
+	// Drop the status to force a CatalogSourceConfig update. This is to account
+	// for the the scenario where a Quay namespace has changed without
+	// app-registry repositories being added or removed but with existing
+	// repositories being updated.
+	cscUpdate.Status = v1alpha1.CatalogSourceConfigStatus{}
```
kevinrizza: If we are going to do this, why not just remove the invalid cache check?

Just for a short term fix?

aravindhp: Do you mean to remove the `IsEntryStale()` check? Won't that cause the `CatalogSourceConfig` to get updated on every kube resync interval?

kevinrizza: No, I mean that we should do something to the csc that makes `IsEntryStale` return true in this case.

This is a nit either way..
 Let's figure it out later to get something working for now.

PR #126, opened 2019-03-05
Bug 1685331: Update priority for operator

aravindhp:
- Specify `system-cluster-critical` priority for operator
- Blocks passing smoke tests for ensuring control plane pods always schedule

See: openshift/origin#22217
Fixes [bug 1685331](https://bugzilla.redhat.com/show_bug.cgi?id=1685331)

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -16,6 +16,7 @@ spec:
       serviceAccountName: marketplace-operator
       nodeSelector:
         node-role.kubernetes.io/master: ""
+      priorityClassName: "system-cluster-critical"
       tolerations:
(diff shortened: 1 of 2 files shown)


aravindhp:
> Shouldn't we also update this in the `deploy/` and `deploy/upstream/` folders?

I figured not because this is an OpenShift specific requirement.

kevinrizza:
@aravindhp Fair point for upstream folder, but the `deploy` folder is not specifically aimed at upstream. I think it should be included.

aravindhp:
@kevinrizza, I added it to the deployment in the `deploy` folder too. Please review.

PR #128, opened 2019-03-06
Bug 1686121: Update Marketplace management of clusterOperator conditions

awgreene:
* Version is accepted as an environment variable within the CSV.
* Improved information provided by clusterOperator Conditions.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -2,8 +2,10 @@ package main
 
 import (
 	"flag"
+	"fmt"
 	"log"
 	"net/http"
+	"os"
 	"runtime"
 	"time"
 
@@ -61,23 +63,24 @@ func main() {
 		log.Fatal(err)
 	}
 
-	status := status.New(cfg, mgr, namespace)
+	operatorStatus := status.New(cfg, mgr, namespace, os.Getenv("RELEASE_VERSION"))
+	operatorStatus.SetProgressing(fmt.Sprintf("Version %s of the operator is being deployed", operatorStatus.GetVersion()))
 
 	log.Print("Registering Components.")
 
 	// Setup Scheme for all defined resources
 	if err := apis.AddToScheme(mgr.GetScheme()); err != nil {
-		fatal(err, status)
+		fatal(err, operatorStatus)
 	}
 
 	// Add external resource to scheme
 	if err := olm.AddToScheme(mgr.GetScheme()); err != nil {
-		fatal(err, status)
+		fatal(err, operatorStatus)
 	}
 
 	// Setup all Controllers
 	if err := controller.AddToManager(mgr); err != nil {
-		fatal(err, status)
+		fatal(err, operatorStatus)
 	}
 
 	// Serve a health check.
@@ -95,12 +98,13 @@ func main() {
 	go registrySyncer.Sync(stopCh)
 	go catalogSyncer.Sync(stopCh)
 
-	status.SetAvailable("Operator running")
+	operatorStatus.SetAvailable(fmt.Sprintf("Version %s of the operator is available", operatorStatus.GetVersion()))
+
 	// Start the Cmd
 	log.Fatal(mgr.Start(stopCh))
 }
 
-func fatal(err error, status status.Status) {
-	status.SetFailing(err.Error())
+func fatal(err error, operatorStatus status.Status) {
+	operatorStatus.SetFailing(err.Error())
 	log.Fatal(err)
 }
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -47,3 +47,6 @@ spec:
                   fieldPath: metadata.namespace
             - name: OPERATOR_NAME
               value: "marketplace-operator"
+            - name: "RELEASE_VERSION"
+              # The string "0.0.1-snapshot" is substituted by the CVO with the version of the payload
+              value: "0.0.1-snapshot"
--- a/manifests/10_operator.yaml
+++ b/manifests/10_operator.yaml
@@ -47,3 +47,6 @@ spec:
                   fieldPath: metadata.namespace
             - name: OPERATOR_NAME
               value: "marketplace-operator"
+            - name: "RELEASE_VERSION"
+              # The string "0.0.1-snapshot" is substituted by the CVO with the version of the payload
+              value: "0.0.1-snapshot"
--- a/manifests/11_clusteroperator.yaml
+++ b/manifests/11_clusteroperator.yaml
@@ -3,3 +3,8 @@ kind: ClusterOperator
 metadata:
   name: marketplace-operator
   namespace: openshift-marketplace
+status:
+  versions:
+    - name: operator
+      # The string "0.0.1-snapshot" is substituted by the CVO with the version of the payload
+      version: "0.0.1-snapshot"
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -7,7 +7,6 @@ import (
 	configclient "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	cohelpers "github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
 	operatorhelpers "github.com/openshift/library-go/pkg/operator/v1helpers"
-	"github.com/operator-framework/operator-marketplace/version"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -24,16 +23,23 @@ type status struct {
 	coAPINotPresent bool
 	namespace       string
 	clusterOperator *configv1.ClusterOperator
+	version         string
 }
 
 // Status is the interface to report the health of the operator to CVO
 type Status interface {
 	SetFailing(message string)
 	SetAvailable(message string)
+	SetProgressing(message string)
+	GetVersion() string
+}
+
+func (s *status) GetVersion() string {
+	return s.version
 }
 
 // New returns an initialized Status
-func New(cfg *rest.Config, mgr manager.Manager, namespace string) Status {
+func New(cfg *rest.Config, mgr manager.Manager, namespace string, version string) Status {
 	// The default client serves read requests from the cache which contains
 	// objects only from the namespace the operator is watching. Given we need
 	// to query CRDs which are cluster wide, we add the CRDs to the manager's
@@ -70,6 +76,7 @@ func New(cfg *rest.Config, mgr manager.Manager, namespace string) Status {
 		configClient:    configClient,
 		coAPINotPresent: coAPINotPresent,
 		namespace:       namespace,
+		version:         version,
 	}
 }
 
@@ -81,6 +88,11 @@ func (s *status) SetFailing(message string) {
 // SetAvailable reports that the operator is available to process events
 func (s *status) SetAvailable(message string) {
 	s.setStatus(configv1.OperatorAvailable, message)
+}
+
+// SetProgressing reports that the operator is being deployed
+func (s *status) SetProgressing(message string) {
+	s.setStatus(configv1.OperatorProgressing, message)
 }
 
 // ensureClusterOperator ensures that a ClusterOperator CR is present on the
@@ -127,7 +139,7 @@ func (s *status) setOperandVersion() {
 	// Report the operator version
 	operatorVersion := configv1.OperandVersion{
 		Name:    "operator",
-		Version: version.Version,
+		Version: s.version,
 	}
 	operatorhelpers.SetOperandVersion(&s.clusterOperator.Status.Versions, operatorVersion)
 }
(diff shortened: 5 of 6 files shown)


awgreene:
@aravindhp I had looked over [this document](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#what-should-be-the-contents-of-clusteroperator-custom-resource-in-manifests)  and saw that CVO requires that the `clusteroperator` has the `status.version` field. It does not look like our current file in the manifest has this field nor does it look like it can be set with the API. Is there anyone you recommend speaking to on the CVO team to learn more?

deploy/operator.yaml
```diff
@@ -46,3 +46,5 @@ spec:
                   fieldPath: metadata.namespace
             - name: OPERATOR_NAME
               value: "marketplace-operator"
+            - name: RELEASE_VERSION
+              value: "0.0.1-snapshot"
```
kevinrizza: Is this the version we want to hardcode? What is our plan for incrementing `RELEASE_VERSION` in the future? We probably want to scope that out and define it in a doc somewhere.

pkg/status/status.go
```diff
@@ -127,50 +165,23 @@ func (s *status) setOperandVersion() {
 	// Report the operator version
 	operatorVersion := configv1.OperandVersion{
 		Name:    "operator",
-		Version: version.Version,
+		Version: os.Getenv("RELEASE_VERSION"),
```
kevinrizza: I'm hesitant about using `os.Getenv` inside this method. It will make testing more complicated, among other things. Can we do the `os.Getenv` in the `main()` method and compose the status struct to include it as a variable?

deploy/operator.yaml
awgreene: The string "0.0.1-snapshot" is substituted by the CVO with the version of the payload
OpenShift operators are expected to report the payload version as their own and developers will have map from payload to code version.

manifests/11_clusteroperator.yaml
```diff
@@ -3,3 +3,8 @@ kind: ClusterOperator
 metadata:
   name: marketplace-operator
   namespace: openshift-marketplace
+spec: {}
+status:
+  versions:
+    - name: operator
+      version: "0.0.1-snapshot"
```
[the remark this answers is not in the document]
awgreene: Will add new line, but see description in comment above.

pkg/status/status.go
```diff
@@ -28,8 +28,52 @@ type status struct {
...
+
+// Condition is a struct used to configure operator status conditions
+type Condition struct {
+	conditionType   configv1.ClusterStatusConditionType
+	conditionStatus configv1.ConditionStatus
+	message         string
+}
+
+// NewAvailableCondition return an available condition based on the status and message provided
+func NewAvailableCondition(condititionStatus bool, message string) Condition {
```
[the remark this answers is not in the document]
awgreene: :+1:

pkg/status/status.go
```diff
@@ -28,8 +28,52 @@ type status struct {
...
+	if condititionStatus {
+		return createCondition(configv1.OperatorAvailable, configv1.ConditionTrue, message)
+	} else {
+		return createCondition(configv1.OperatorAvailable, configv1.ConditionFalse, message)
+	}
+}
+
+// NewProgressingCondition return a progressing condition based on the status and message provided
+func NewProgressingCondition(condititionStatus bool, message string) Condition {
```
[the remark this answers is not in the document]
awgreene: :+1:

pkg/status/status.go
```diff
@@ -28,8 +28,52 @@ type status struct {
...
+	if condititionStatus {
+		return createCondition(configv1.OperatorProgressing, configv1.ConditionTrue, message)
+	} else {
+		return createCondition(configv1.OperatorProgressing, configv1.ConditionFalse, message)
+	}
+}
+
+// NewFailingCondition return a failing condition based on the status and message provided
+func NewFailingCondition(condititionStatus bool, message string) Condition {
```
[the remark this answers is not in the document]
awgreene: :+1:

pkg/status/status.go
```diff
@@ -28,8 +28,52 @@ type status struct {
...
+	if condititionStatus {
+		return createCondition(configv1.OperatorFailing, configv1.ConditionTrue, message)
+	} else {
+		return createCondition(configv1.OperatorFailing, configv1.ConditionFalse, message)
+	}
+}
```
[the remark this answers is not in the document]
awgreene: I don't foresee a difference outside of the `configv1.ConditionStatus` parameter. I had avoided converging these methods into a single method because I wanted the calls to be clear, concise, and not require `configv1` to be imported. I do see the benefit of generalizing the method, we could delete the `NewAvailableCondition/NewProgressingCondition/NewFailingCondition` methods and update existing calls to reference the `createCondition`  method.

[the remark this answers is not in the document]
aravindhp: I agree with @kevinrizza. This seems like a lot of repeated code. Not sure how much we are gaining here vs the initial implementation which was simpler.

scripts/e2e-tests.sh
```diff
@@ -2,6 +2,7 @@
 set -e
 
 TEST_NAMESPACE="openshift-marketplace"
+RELEASE_VERSION="0.0.1-snapshot"
```
aravindhp: Please add a comment as to why this is needed here.

deploy/operator.yaml
kevinrizza: +1

pkg/status/status.go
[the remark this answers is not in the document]
awgreene: Understood, I will remove the `NewAvailableCondition/NewProgressingCondition/NewFailingCondition` methods

> Not sure how much we are gaining here vs the initial implementation which was simpler.

While the original implementation was simpler, we could not update individual condition statuses while leave others conditions as they were .

cmd/manager/main.go
```diff
@@ -61,23 +63,28 @@ func main() {
 		log.Fatal(err)
 	}
 
-	status := status.New(cfg, mgr, namespace)
+	operatorStatus := status.New(cfg, mgr, namespace)
+	operatorStatus.SetStatus([]status.Condition{
+		status.NewAvailableCondition(false, "The Operator is not yet available"),
```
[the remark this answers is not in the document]
awgreene: :+1:

cmd/manager/main.go
```diff
@@ -95,12 +102,19 @@ func main() {
 	go registrySyncer.Sync(stopCh)
 	go catalogSyncer.Sync(stopCh)
 
-	status.SetAvailable("Operator running")
+	operatorStatus.SetStatus([]status.Condition{
+		status.NewAvailableCondition(true, fmt.Sprintf("Version %s of the operator is available", os.Getenv("RELEASE_VERSION"))),
```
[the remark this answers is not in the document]
awgreene: The fields only need to be updated if they are changed.

pkg/status/status.go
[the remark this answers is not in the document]
aravindhp: >While the original implementation was simpler, we could not update individual condition statuses while leave others conditions as they were .
Isn't there a risk of them being changed in an non-atomic manner? Moreover do we have that use case.

pkg/status/status.go
awgreene: Updated accordingly.

pkg/status/status.go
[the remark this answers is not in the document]
awgreene: > Isn't there a risk of them being changed in an non-atomic manner? 

The only method that changes the status is `status.UpdateStatus` which is called after updating all the conditions - I don't see a point where these statuses could be changed individually..

> Moreover do we have that use case?

Marketplace doesn't take advantage of setting individual conditions yet so I could move this into a method similar to the one we had earlier if you don't see this being a usecase in the future.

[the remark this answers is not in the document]
aravindhp: >Marketplace doesn't take advantage of setting individual conditions yet so I could move this into a method similar to the one we had earlier if you don't see this being a usecase in the future.


I prefer introducing this level of changes when it is actually needed. Plus we are close to GA, so less code changes the better.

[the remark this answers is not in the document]
awgreene: Unerstood, I will update the code accordingly.

awgreene:
In the latest e2e-aws-operator test, the version featured in the clusteroperator for marketplace was updated from the `0.0.1-snapshot` placeholder value:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/operator-framework_operator-marketplace/128/pull-ci-operator-framework-operator-marketplace-master-e2e-aws-operator/308/artifacts/e2e-aws-operator/clusteroperators.json
```
{
            "apiVersion": "config.openshift.io/v1",
            "kind": "ClusterOperator",
            "metadata": {
                "creationTimestamp": "2019-03-07T15:58:52Z",
                "generation": 1,
                "name": "marketplace-operator",
                "resourceVersion": "20596",
                "selfLink": "/apis/config.openshift.io/v1/clusteroperators/marketplace-operator",
                "uid": "e74604b2-40f1-11e9-bec5-12e395f32ee8"
            },
            "spec": {},
            "status": {
                "conditions": [
                    {
                        "lastTransitionTime": "2019-03-07T16:06:20Z",
                        "status": "False",
                        "type": "Progressing"
                    },
                    {
                        "lastTransitionTime": "2019-03-07T16:06:20Z",
                        "message": "Version 0.0.1-2019-03-07-153050 of the operator is available",
                        "status": "True",
                        "type": "Available"
                    },
                    {
                        "lastTransitionTime": "2019-03-07T15:58:52Z",
                        "status": "False",
                        "type": "Failing"
                    }
                ],
                "extension": null,
                "relatedObjects": null,
                "versions": [
                    {
                        "name": "operator",
                        "version": "0.0.1-2019-03-07-153050"
                    }
                ]
            }
        }
 ```

**Comment by awgreene** [comment]
@anik120 @aravindhp @kevinrizza Please review

File: cmd/manager/main.go [hunk 11]
```diff
@@ -61,23 +63,24 @@ func main() {
 		log.Fatal(err)
 	}
 
-	status := status.New(cfg, mgr, namespace)
```
aravindhp: Do you really need this variable change now that you are just adding to the older implementation? Not changing the variable name will reduce the code churn.

deploy/operator.yaml
```diff
@@ -47,3 +47,6 @@ spec:
                   fieldPath: metadata.namespace
             - name: OPERATOR_NAME
               value: "marketplace-operator"
+            - name: "RELEASE_VERSION"
+              # The string "0.0.1-snapshot" is substituted by the CVO with the version of the payload
```
aravindhp: This won't happen given this yaml is used only by the SDK. I suggest removing the comment and leaving the rest for consistency.

I am also not sure if we need this here. @kevinrizza what do you think?

manifests/10_operator.yaml
```diff
@@ -47,3 +47,6 @@ spec:
                   fieldPath: metadata.namespace
             - name: OPERATOR_NAME
               value: "marketplace-operator"
+            # The string "0.0.1-snapshot" is substituted by the CVO with the version of the payload
```
aravindhp: Please move the commit to the correct spot

cmd/manager/main.go
awgreene: I implemented the name change because the variable `status` conflicted with the package of the same name. If you don't believe this is an issue I can revert the change.

aravindhp: This shouldn't be an issue given that's the way it was before. But now that you point that out I agree with you.

manifests/10_operator.yaml
awgreene: Will do

aravindhp:
> In the latest e2e-aws-operator test, the version featured in the clusteroperator for marketplace was updated from the `0.0.1-snapshot` placeholder value:

Is CVO just replacing the `snapshot` part of the version?

awgreene:
@aravindhp
> @awgreene have you confirmed that the version changes get reflected correctly during an upgrade?

I did not test the version change through an official upgrade. I had investigated if this was even possible given that to make this change we had to override the CVO to ignore the marketplace operator. I reached out to Ben Parees, author of the original PR that you directed me to, and was told that the only test they have in this area is [here}(https://github.com/openshift/cluster-image-registry-operator/pull/225/files#diff-53f8d9834e73fd1861ace685be728221R122). 

As part of this test the CVO is disabled, the `RELEASE_VERSION` is updated and checks that the `clusteroperator` status is updated accordingly. I have done this manually.

awgreene:
@aravindhp 
>Is CVO just replacing the snapshot part of the version?

The `0.0.1` is present in the `clusteroperator` version but I am unsure if they replace only the snapshot or the entire string. It was confirmed that we need `0.0.1-snapshot` by the CVO team.

cmd/manager/main.go
awgreene: It's doesn't appear to be an issue unless you want to use a different method from the `status` package - once the variable is declared with the same name as the package you will only be able to use methods from the `Status` interface which the status stuct implements.

deploy/operator.yaml
kevinrizza: @aravindhp I think it doesn't hurt to keep these yamls in sync. If we ever updated our build process somehow to actually use the sdk to build, then it would be useful not to have to go back and make all these changes. But I am also okay with not doing it.

aravindhp: Let's keep it in sync then

PR #133, opened 2019-03-13
Bug 1679284: [csc] Fix PackageRepositoryVersions json string

kevinrizza:
Problem:
When adding catalogsourceconfig.status.PackageRepositioryVersions to
the schema of CatalogSourceConfig, the json field was defined improperly

Solution:
Update the `status.PackageRepositoryVersion` json string name from
`packageMap` to `packageRepositioryVersions`

This should have been included as part of the fix for:
https://bugzilla.redhat.com/show_bug.cgi?id=1679284


kevinrizza:
> Please link to the BZ (in header and body) and mention that it should have been part of that fix.

@aravindhp addressed.

PR #134, opened 2019-03-13
Bug 1686121 & 1689443: Update management of clusterOperator

awgreene:
Cluster operators are expected to report the version of the payload
they are included in once they are "deployed", and also to keep the
cluster operator object created. With this update the marketplace
operator will keep CO up to date, report the payload version once it
hits available, and use the count of successful syncs as a
probalistic measurement of "available".

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1686121

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -2,14 +2,13 @@ package main
 
 import (
 	"flag"
-	"fmt"
-	"log"
 	"net/http"
 	"os"
 	"runtime"
 	"time"
 
 	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/signals"
 	"github.com/operator-framework/operator-marketplace/pkg/apis"
 	"github.com/operator-framework/operator-marketplace/pkg/catalogsourceconfig"
 	"github.com/operator-framework/operator-marketplace/pkg/controller"
@@ -17,10 +16,10 @@ import (
 	"github.com/operator-framework/operator-marketplace/pkg/status"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
+	log "github.com/sirupsen/logrus"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
-	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
 )
 
 const (
@@ -63,27 +62,29 @@ func main() {
 		log.Fatal(err)
 	}
 
-	operatorStatus := status.New(cfg, mgr, namespace, os.Getenv("RELEASE_VERSION"))
-	operatorStatus.SetProgressing(fmt.Sprintf("Version %s of the operator is being deployed", operatorStatus.GetVersion()))
-
 	log.Print("Registering Components.")
 
 	catalogsourceconfig.InitializeStaticSyncer(mgr.GetClient(), initialWait)
 	registrySyncer := operatorsource.NewRegistrySyncer(mgr.GetClient(), initialWait, resyncInterval, updateNotificationSendWait, catalogsourceconfig.Syncer, catalogsourceconfig.Syncer)
 
+	// monitorStopCh is used to send a signal to stop reporting cluteroperator status
+	monitorStopCh := make(chan struct{})
+	// monitorDoneCh will recieve a signal when threads have stopped updating cluster operator status
+	monitorDoneCh := status.StartReporting(cfg, mgr, namespace, os.Getenv("RELEASE_VERSION"), monitorStopCh)
+
 	// Setup Scheme for all defined resources
 	if err := apis.AddToScheme(mgr.GetScheme()); err != nil {
-		fatal(err, operatorStatus)
+		fatal(err, monitorDoneCh, monitorStopCh)
 	}
 
 	// Add external resource to scheme
 	if err := olm.AddToScheme(mgr.GetScheme()); err != nil {
-		fatal(err, operatorStatus)
+		fatal(err, monitorDoneCh, monitorStopCh)
 	}
 
 	// Setup all Controllers
 	if err := controller.AddToManager(mgr); err != nil {
-		fatal(err, operatorStatus)
+		fatal(err, monitorDoneCh, monitorStopCh)
 	}
 
 	// Serve a health check.
@@ -98,13 +99,19 @@ func main() {
 	go registrySyncer.Sync(stopCh)
 	go catalogsourceconfig.Syncer.Sync(stopCh)
 
-	operatorStatus.SetAvailable(fmt.Sprintf("Version %s of the operator is available", operatorStatus.GetVersion()))
-
 	// Start the Cmd
-	log.Fatal(mgr.Start(stopCh))
+	// Update clusteroperator status and log a fatal error
+	fatal(mgr.Start(stopCh), monitorDoneCh, monitorStopCh)
 }
 
-func fatal(err error, operatorStatus status.Status) {
-	operatorStatus.SetFailing(err.Error())
+// fatal updates clusteroperator status and logs a fatal error
+func fatal(err error, monitorDoneCh <-chan struct{}, monitorStopCh chan struct{}) {
+	// Stop reporting clusterOperator status
+	close(monitorStopCh)
+
+	// Wait for clusterOperator status reporting to stop
+	<-monitorDoneCh
+
+	// exit
 	log.Fatal(err)
 }
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -14,12 +14,6 @@ rules:
   - update
   - list
 - apiGroups:
-  - apiextensions.k8s.io
-  resources:
-  - customresourcedefinitions
-  verbs:
-  - get
-- apiGroups:
   - config.openshift.io
   resources:
   - clusteroperators
--- a/pkg/controller/catalogsourceconfig/catalogsourceconfig_controller.go
+++ b/pkg/controller/catalogsourceconfig/catalogsourceconfig_controller.go
@@ -5,6 +5,7 @@ import (
 
 	marketplacev1alpha1 "github.com/operator-framework/operator-marketplace/pkg/apis/marketplace/v1alpha1"
 	catalogsourceconfighandler "github.com/operator-framework/operator-marketplace/pkg/catalogsourceconfig"
+	"github.com/operator-framework/operator-marketplace/pkg/status"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -81,10 +82,19 @@ type ReconcileCatalogSourceConfig struct {
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileCatalogSourceConfig) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	log.Printf("Reconciling CatalogSourceConfig %s/%s\n", request.Namespace, request.Name)
+	// Reconcile kicked off, message Sync Channel
+	status.SendSyncMessage(nil)
 
+	// If err is not nil at the end of this function message the syncCh channel
+	var err error
+	defer func() {
+		if err != nil {
+			status.SendSyncMessage(err)
+		}
+	}()
 	// Fetch the CatalogSourceConfig instance
 	instance := &marketplacev1alpha1.CatalogSourceConfig{}
-	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
+	err = r.client.Get(context.TODO(), request.NamespacedName, instance)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -1,22 +1,54 @@
 package status
 
 import (
-	"context"
+	"fmt"
+	"sync"
+	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
 	configclient "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	cohelpers "github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
 	operatorhelpers "github.com/openshift/library-go/pkg/operator/v1helpers"
 	log "github.com/sirupsen/logrus"
-	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
-const clusterOperatorName = "marketplace-operator"
+const (
+	// clusterOperatorName is the name of the cluster operator
+	clusterOperatorName = "marketplace-operator"
+
+	// minSyncsBeforeReporting is the minimum number of syncs we wish to see
+	// before reporting that the operator is available
+	minSyncsBeforeReporting = 4
+
+	// successRatio is the ratio of successful syncs / total syncs that we
+	// want to see in order to report that the marketplace operator is healthy.
+	// This value is low right now because the failed syncs come from invalid CRs.
+	// As the status reporting evolves we can tweek this ratio to be a better
+	// representation of the operator's health.
+	successRatio = 0.3
+
+	// syncsBeforeTruncate is used to prevent the totalSyncs and
+	// failedSyncs values from overflowing in a long running operator.
+	// Once totalSyncs reaches the maxSyncsBeforeTruncate value, totalSyncs
+	// and failedSyncs will be recalculated with the following
+	// equation: updatedValue = currentValue % syncTruncateValue.
+	syncsBeforeTruncate = 10000
+	syncTruncateValue   = 100
+
+	// coStatusReportInterval is the interval at which the cluster operator status is updated
+	coStatusReportInterval = 20 * time.Second
+)
+
+// status will be a singleton
+var instance *status
+var once sync.Once
 
 type status struct {
 	configClient    *configclient.ConfigV1Client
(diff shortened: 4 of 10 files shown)


pkg/status/status.go
```diff
@@ -195,8 +205,85 @@ func (s *status) setStatusCondition(condition configv1.ClusterStatusConditionTyp
...
+			// grab any other sync events that have accumulated
+			for len(SyncCh) > 0 {
+				if err := <-SyncCh; err == nil {
+					successfulSyncs++
+				}
+				syncs++
+			}
+			// if we haven't yet accumulated enough syncs, wait longer
+			// TODO: replace these magic numbers with a better measure of syncs
+			if successfulSyncs < 3 && syncs < 5 {
```
awgreene: I would appreciate if people would share their opinion on whether or not they believe that we are waiting to process enough syncs.

pkg/status/status.go
```diff
@@ -81,17 +90,17 @@ func New(cfg *rest.Config, mgr manager.Manager, namespace string, version string
...
-func (s *status) SetAvailable(message string) {
+func (s *status) setAvailable(message string) {
 	s.setStatus(configv1.OperatorAvailable, message)
 }
 
 // SetProgressing reports that the operator is being deployed
-func (s *status) SetProgressing(message string) {
+func (s *status) setProgressing(message string) {
 	s.setStatus(configv1.OperatorProgressing, message)
 }
```
awgreene: At this point we do not allow setting multiple conditions. This becomes an issue when v1 of the operator is installed and available but v2 is being rolled out. Technically the CO should report that the operator is available for version v1 and is progressing to v2. Let me know your thoughts.

pkg/status/status.go
```diff
@@ -195,8 +205,85 @@ func (s *status) setStatusCondition(condition configv1.ClusterStatusConditionTyp
 
 // updateStatus makes the API call to update the ClusterOperator
 func (s *status) updateStatus() {
-	_, err := s.configClient.ClusterOperators().UpdateStatus(s.clusterOperator)
+	newClusterOperator, err := s.configClient.ClusterOperators().UpdateStatus(s.clusterOperator)
 	if err != nil {
 		log.Fatalf("Error %v updating ClusterOperator", err)
+	} else {
+		s.clusterOperator = newClusterOperator
```
awgreene: Note: We must update the clusterOperator to match the new CR in the cluster.

awgreene:
@aravindhp @kevinrizza could you please review

awgreene:
@tkashem could you review as well?

pkg/status/status.go
```diff
@@ -200,3 +227,71 @@ func (s *status) updateStatus() {
...
+			time.Sleep(5 * time.Minute)
+		} else {
+			s.setFailing(fmt.Sprintf("Waiting for updates to take effect"))
+		}
+
+	}, 5*time.Second, stopCh)
+}
+
+// succeeding returns whether or not the cluster operator should report a successful state
+func (s *status) succeeding() bool {
```
kevinrizza: Can this be `isSucceeding`? The current name makes it sound like it's not a question, but a setter.

cmd/manager/main.go
```diff
@@ -93,18 +93,15 @@ func main() {
...
-	operatorStatus.SetAvailable(fmt.Sprintf("Version %s of the operator is available", operatorStatus.GetVersion()))
-
 	// Start the Cmd
 	log.Fatal(mgr.Start(stopCh))
 }
 
-func fatal(err error, operatorStatus status.Status) {
-	operatorStatus.SetFailing(err.Error())
+func fatal(err error) {
+	status.SendSyncMessage(err)
```
kevinrizza: Do we need to do more than send an err to the channel? Don't we still want to set the status to failing somehow?

pkg/status/status.go
```diff
@@ -1,65 +1,89 @@
...
 )
 
-const clusterOperatorName = "marketplace-operator"
+const (
+	clusterOperatorName     = "marketplace-operator"
+	minSyncsBeforeReporting = 4
+	// At this point, a failure is critical so all syncs must be successful
+	// to report a working operator
+	successRatio           = 1.0
+	maxSyncsBeforeTruncate = 10000
```
kevinrizza: Can we document all of these constant magic numbers? I understand them now but I will probably forget if I look months from now :)

pkg/status/status.go
```diff
@@ -1,65 +1,89 @@
...
+	syncTruncateValue      = 100
+)
+
+// SyncCh is used to report a sync event
+// The ratio of Successful Syncs / Total Syncs will be used to report operator health to the CVO
+var syncCh = make(chan error)
+
+// status will be a singleton
```
kevinrizza: I know this is a singleton, but it will be hard to test this if we don't compose it at all. Can these fields all be in the status type?

cmd/manager/main.go
awgreene: Right now we only send an error on a critical failure, so if the sync ratio is less than one we report failure.

pkg/controller/catalogsourceconfig/catalogsourceconfig_controller.go
```diff
@@ -81,6 +82,8 @@ type ReconcileCatalogSourceConfig struct {
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileCatalogSourceConfig) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	log.Printf("Reconciling CatalogSourceConfig %s/%s\n", request.Namespace, request.Name)
+	// Reconcile kicked off, message Sync Channel
+	status.SendSyncMessage(nil)
```
aravindhp: Shouldn't you also send a message after a successful or failed reconciliation?

pkg/status/status.go
awgreene: Great point, will add comments

pkg/status/status.go
awgreene: Sure

pkg/status/status.go
```diff
@@ -1,65 +1,89 @@
...
 		log.Fatal(err)
 	}
-
-	// Check if the ClusterOperator API is present on the cluster. This is so
-	// that we can continue to work with Kubernetes clusters that don't have
-	// this API.
-	key := client.ObjectKey{
-		Name: "clusteroperators.config.openshift.io",
-	}
-	err = controllerClient.Get(context.TODO(), key, (&v1beta1.CustomResourceDefinition{}))
```
aravindhp: With this change, we are no longer are making `get` CRD requests. Please update the `ClusterRole`.

pkg/status/status.go
```diff
@@ -159,6 +184,8 @@ func (s *status) setStatusCondition(condition configv1.ClusterStatusConditionTyp
 	case configv1.OperatorAvailable:
 		availableStatus = configv1.ConditionTrue
 		availableMessage = message
+		// Only update the version when the operator becomes available
+		s.setOperandVersion()
```
aravindhp: A log message reflecting the current and previous version will be helpful. You can combine it with another existing log message if needed.

pkg/status/status.go
```diff
@@ -200,3 +227,71 @@ func (s *status) updateStatus() {
...
+func syncChannelConsumer(instance *status, stopCh <-chan struct{}) {
+	log.Info("Starting sync consumer")
+	go wait.Until(func() {
+		// Listen on SyncCh and update instance accordingly
+		err := <-syncCh
+		instance.syncs++
+		if err == nil {
+			instance.successfulSyncs++
+		}
+		log.Infof("Successful Syncs / Total Syncs: %d/%d", instance.successfulSyncs, instance.syncs)
```
aravindhp: Will this make the log verbose?

pkg/status/status.go
```diff
@@ -200,3 +227,71 @@ func (s *status) updateStatus() {
...
+	go syncChannelConsumer(instance, stopCh)
+	go wait.Until(func() {
+		// create the cluster operator in an initial state if it does not exist
+		if s.clusterOperator == nil {
+			s.setProgressing(fmt.Sprintf("Progressing towards %s", s.version))
+		}
+
+		// wait until the operator has synced enough objects
+		if s.syncs < minSyncsBeforeReporting {
+			log.Printf("Waiting to observe more syncs")
```
aravindhp: Please use `Infof` and mention the number syncs we are waiting for

pkg/status/status.go
```diff
@@ -200,3 +227,71 @@ func (s *status) updateStatus() {
...
+			return
+		}
+
+		// update the status with the appropriate state
+		if s.succeeding() {
+			s.setAvailable(fmt.Sprintf("%s is available", s.version))
+			// if we've reported success, we can sleep longer, otherwise we want to keep watching for
+			// successful
```
aravindhp: `keep watching until we are successful`

pkg/status/status.go
```diff
@@ -200,3 +227,71 @@ func (s *status) updateStatus() {
...
+	// exit if cluster operator api is not present
+	if s.coAPINotPresent {
+		return
+	}
+
+	go syncChannelConsumer(instance, stopCh)
+	go wait.Until(func() {
+		// create the cluster operator in an initial state if it does not exist
+		if s.clusterOperator == nil {
+			s.setProgressing(fmt.Sprintf("Progressing towards %s", s.version))
```
aravindhp: So this will happen only on a fresh cluster. Don't we need to show progressing on upgrades?

pkg/status/status.go
```diff
@@ -200,3 +227,71 @@ func (s *status) updateStatus() {
...
+		}
+
+		// wait until the operator has synced enough objects
+		if s.syncs < minSyncsBeforeReporting {
```
aravindhp: Why are we waiting to report? The fact that I asked the question implies you need to add a comment here to explain.

pkg/status/status.go
```diff
@@ -200,3 +227,71 @@ func (s *status) updateStatus() {
...
+		}
+
+		// update the status with the appropriate state
+		if s.succeeding() {
+			s.setAvailable(fmt.Sprintf("%s is available", s.version))
+			// if we've reported success, we can sleep longer, otherwise we want to keep watching for
+			// successful
+			time.Sleep(5 * time.Minute)
+		} else {
+			s.setFailing(fmt.Sprintf("Waiting for updates to take effect"))
```
aravindhp: Are we waiting for an update or for the sync ratio to be achieved?

pkg/status/status.go
awgreene: We do need to show progressing on upgrades.

This code will show progressing on upgrades.The `instance` variable is created when the operator starts the `status.StartReporting` method without a `clusterOperator` and this pointer isn't updated to reflect what is in on cluster until `s.setProgressing` is called.

pkg/status/status.go
```diff
@@ -1,65 +1,89 @@
...
-	// scheme and create our own client.
-	if err := v1beta1.AddToScheme(mgr.GetScheme()); err != nil {
-		log.Fatal(err)
+// StartReporting ensures that a status object exists and then starts updating Marketplace's clusteroperator
+func StartReporting(cfg *rest.Config, mgr manager.Manager, namespace string, version string, stopCh <-chan struct{}) {
+	if instance == nil {
+		// status is a singleton and should prevent threads from instantiating multiple instances.
+		mu.Lock()
+		defer mu.Unlock()
+		if instance == nil {
```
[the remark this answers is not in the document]
awgreene: It was to avoid excessive locking . After some additional research I believe that this should be replaced by the [sync.Once](https://golang.org/pkg/sync/#example_Once) type.

pkg/status/status.go
```diff
@@ -200,3 +227,71 @@ func (s *status) updateStatus() {
 		log.Fatalf("Error %v updating ClusterOperator", err)
 	}
 }
+
+// syncChannelConsumer will listen on the sync channel and update the status'
+// syncs and successfulSyncs fields until the stopCh is closed.
+func syncChannelConsumer(instance *status, stopCh <-chan struct{}) {
```
tkashem: you can use `select` to express what `syncChannelConsumer` does in a concise manner.
```
func syncChannelConsumer(instance *status, stopCh <-chan struct{}) {
  for {
    select {
    case <-stopCh:
        return
     case err := <-syncCh
       ...
    }
  }
}
```

pkg/status/status.go
tkashem: maybe we should convert this to `debug`?

pkg/status/status.go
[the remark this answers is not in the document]
tkashem: I believe it's the `double checked locking` pattern, you can achieve this with `do.Once`
```
once.Do(func() {
    // perform safe initialization here
})
```

cmd/manager/main.go
aravindhp: I agree with @kevinrizza. A fatal error can happen before `StartReporting()` starts and you would need to set the status to failing.

I am now wondering if we need to do something to be more resilient and retry on initialization failures. But that is for another PR.

tkashem: What the `fatal` function does is write to the channel and then immediately exit. How do we ensure that we give enough time to the other go routine to update the clusteroperator object? Maybe use some sort of `done` channel?

The other question is, we want to update the `clusteroperator` object here, irrespective of sync sync count, right?

pkg/controller/catalogsourceconfig/catalogsourceconfig_controller.go
tkashem: I think we just need to do it once, after the `handle` function is called.
```
err = r.CatalogSourceConfigHandler.Handle(context.TODO(), instance)
status.SendSyncMessage(err)
```
It covers both scenarios.

pkg/status/status.go
```diff
@@ -1,65 +1,89 @@
...
+var mu sync.Mutex
 
 type status struct {
 	configClient    *configclient.ConfigV1Client
 	coAPINotPresent bool
 	namespace       string
 	clusterOperator *configv1.ClusterOperator
 	version         string
+	successfulSyncs int
+	syncs           int
```
tkashem: both `monitorClusterStatus` and `syncChannelConsumer` are running in different go routines, one is reading from these stats and the other is writing to these. In such case, we run into race condition. Ideally we should update these stats in an atomic manner.

pkg/status/status.go
awgreene: This `if` statement exists to make sure that we've processed a few syncs before declaring that the operator is available. Also - as the code is currently written t- he status.syncRatio` method could divide by zero.

I will make sure to update the comment.

pkg/status/status.go
awgreene: This is a good catch. At this point all sync reports should be successful unless Marketplace encounters a critical error. If we set this to failing, we should report that marketplace has encountered a critical failure.

pkg/status/status.go
kevinrizza: @tkashem `monitorClusterStatus` is intended to be async from `syncChannelConsumer`. It doesn't matter if `monitorClusterStatus` reads the wrong value, as long as they are eventually consistent (ie. 5 seconds later when the sync runs again). If the only concern is thread safety, then we should just put a lock around the writes to the stats.

pkg/status/status.go
```diff
@@ -200,3 +227,71 @@ func (s *status) updateStatus() {
...
+// succeeding returns whether or not the cluster operator should report a successful state
+func (s *status) succeeding() bool {
+	if s.syncRatio() >= successRatio {
+		return true
+	}
+	return false
+}
+
+// syncRatio returns the ratio of s.successfulSyncs to s.syncs
+func (s *status) syncRatio() float32 {
```
kevinrizza: Should there be some defensive code here to make sure that `s.syncs` is not zero?

pkg/status/status.go
tkashem: yeah, each can run concurrently if we desire to. But by splitting the read and write into their own go routines we now have to make sure to eliminate race conditions.

pkg/status/status.go
awgreene: I will update the code to return an error if `s.syncs` is equal to zero.

pkg/controller/catalogsourceconfig/catalogsourceconfig_controller.go
awgreene: I don't think we want to handle this case as an error. At this point we report failing only if we have a critical error. If we reach the failing state our operator does not recover.

pkg/status/status.go
```diff
@@ -1,65 +1,89 @@
...
-}
-
-func (s *status) GetVersion() string {
-	return s.version
+// SendSyncMessage is used to send messages to the SyncCh
+func SendSyncMessage(err error) {
+	if instance != nil {
+		// If the coAPI is not available do not send messages to the sync channel
+		if !instance.coAPINotPresent {
+			syncCh <- nil
```
tkashem: I would recommend using `select` and `default` to make it a non blocking call ( even though the channel is buffered ).
```
select {
	case syncCh <- err:
	default:
	}
```

pkg/controller/catalogsourceconfig/catalogsourceconfig_controller.go
aravindhp: >I think we just need to do it once, after the handle function is called.

@tkashem We can't do it just once as we want to differentiate between a sync and a successful sync.

pkg/status/status.go
aravindhp: I suggest that we can make these variables [atomic](https://golang.org/pkg/sync/atomic/).

tkashem: You could have interleaved read between the two atomic writes ( write `successfulSyncs` and then write  `syncs` ). These two writes constitute a single logical transaction. On the other hand, these are stats where we can tolerate some staleness. I am fine either way.

pkg/status/status.go
awgreene: syncCh will need to be a global variable because it's used by controllers but I will make sure that all other fields can be set in the call.

pkg/status/status.go
awgreene: Updated.

pkg/status/status.go
awgreene: @kevinrizza These fields have been moved into the status type.

pkg/status/status.go
awgreene: Included in the latest commit.

awgreene:
@aravindhp @tkashem @kevinrizza please review.

deploy/role.yaml
```diff
@@ -13,12 +13,6 @@ rules:
   - delete
   - update
   - list
-- apiGroups:
-  - apiextensions.k8s.io
-  resources:
-  - customresourcedefinitions
-  verbs:
-  - get
```
kevinrizza: Can you refresh my memory here, why is this in the rbac in the first place and why is it being removed now?

awgreene: [Previously](https://github.com/operator-framework/operator-marketplace/pull/134/files#diff-a5666326aefdf43a8e89fd4871449fa0L62) the status package checked that the ClusterOperator CRD existed. Now we use the `k8sinterface.Discovery` approach.

pkg/status/status.go
```diff
@@ -191,12 +259,90 @@ func (s *status) setStatusCondition(condition configv1.ClusterStatusConditionTyp
...
 }
 
-// updateStatus makes the API call to update the ClusterOperator
-func (s *status) updateStatus() {
-	_, err := s.configClient.ClusterOperators().UpdateStatus(s.clusterOperator)
-	if err != nil {
-		log.Fatalf("Error %v updating ClusterOperator", err)
+// updateStatus makes the API call to update the ClusterOperator if the status has changed.
+func (s *status) updateStatus(previousStatus *configv1.ClusterOperatorStatus, updatedCondition string, message string) {
+	if reflect.DeepEqual(previousStatus, &s.clusterOperator.Status) {
```
aravindhp: Are you sure you cannot leverage the library-go [helpers](https://github.com/openshift/library-go/blob/master/pkg/config/clusteroperator/v1helpers/status.go) for ClusterOperator here?

pkg/status/status.go
```diff
@@ -129,12 +193,14 @@ func (s *status) setStatus(condition configv1.ClusterStatusConditionType, messag
 		return
 	}
 	s.ensureClusterOperator()
-	s.setStatusCondition(condition, message)
-	s.setOperandVersion()
-	s.updateStatus()
+	previousStatus := s.clusterOperator.Status.DeepCopy()
```
aravindhp: Are you sure you cannot leverage the library-go [helpers](https://github.com/openshift/library-go/blob/master/pkg/config/clusteroperator/v1helpers/status.go) for ClusterOperator here?

pkg/status/status.go
```diff
@@ -191,12 +259,90 @@ func (s *status) setStatusCondition(condition configv1.ClusterStatusConditionTyp
...
+		log.Debugf("Previous and current clusterOperator Status are the same, the clusterOperator Status will not be updated.")
+	} else {
+		log.Debugf("Previous and current clusterOperator Status are different, attempting to update the clusterOperator Status.")
+
+		// Check if the clusterOperator version has changed and log the attempt to upgrade if it has
+		previousVersion := operatorhelpers.FindOperandVersion(previousStatus.Versions, "operator")
+		currentVersion := operatorhelpers.FindOperandVersion(s.clusterOperator.Status.Versions, "operator")
+		if currentVersion != nil {
+			if previousVersion == nil {
+				log.Infof("Attempting to upgrade clusterOperator to version %s", currentVersion.Version)
```
aravindhp: Why would this be an upgrade if the `previousVersion` was `nil`?

pkg/controller/catalogsourceconfig/catalogsourceconfig_controller.go
```diff
@@ -100,6 +103,7 @@ func (r *ReconcileCatalogSourceConfig) Reconcile(request reconcile.Request) (rec
 	instance.EnsureGVK()
 
 	err = r.CatalogSourceConfigHandler.Handle(context.TODO(), instance)
+	status.SendSyncMessage(err)
```
aravindhp: How are you differentiating between a sync (event was received) and a successful sync (event was handled without errors) in the case where `err` is `nil`? This seems like you are counting a "successful sync" also as a sync.

awgreene:
@aravindhp 
> You have not renamed the ClusterOperator CR name. Are you planning on doing it in a follow up PR? I suggest doing that in this PR in a separate commit. The PR header prefix should include both bug IDs and the respective bug IDs should be the prefix for each commit.

I had planed to rename the ClusterOpeartor CR name in the next commit. I will make sure update the PR header and commit headers as you suggested.

pkg/status/status.go
awgreene: Good point. I will change this from "Attempting to upgrade..." to "Attempting to set..."

pkg/status/status.go
```diff
@@ -1,65 +1,123 @@
...
+
+	// minSyncsBeforeReporting is the minimum number of syncs we wish to see
+	// before reporting that the operator is available
+	minSyncsBeforeReporting = 4
+
+	// successRatio is the ratio of successful syncs / total syncs that we
+	// want to see in order to report that the marketplace operator is healthy.
+	successRatio = 0.6
+
+	// syncsBeforeTruncate is used to prevent the totalSyncs and
```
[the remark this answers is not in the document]
awgreene: Will do.

pkg/controller/catalogsourceconfig/catalogsourceconfig_controller.go
awgreene: The code treats both a sync event and a successful sync in the same manner. If you think we should treat them as separate events could you explain what you think that should look like?

pkg/controller/catalogsourceconfig/catalogsourceconfig_controller.go
awgreene: Per our discussion this morning, I have updated the code in the latest commit to report a sync event and a sync failure if any events occur:
```
func (r *ReconcileOperatorSource) Reconcile(request reconcile.Request) (reconcile.Result, error) {
	log.Printf("Reconciling OperatorSource %s/%s\n", request.Namespace, request.Name)
	// Reconcile kicked off, message Sync Channel with sync event
	status.SendSyncMessage(nil)

	// Message Sync Channel with Failure if err is not nil at the end of this function
	var err error
	defer func() {
		if err != nil {
			status.SendSyncMessage(err)
		}
	}()
...
...
...
```

pkg/status/status.go
awgreene: The library-go helpers does not provide a tool to copy the status.

awgreene:
The cluster fails to initialize because it is looking for the old clusteroperator name

>level=fatal msg="failed to initialize the cluster: Cluster operator marketplace-operator has not yet reported success"

aravindhp:
>commented about 2 hours ago
The cluster fails to initialize because it is looking for the old clusteroperator name

You might need to fix up the test to make this pass.

cmd/manager/main.go
```diff
@@ -63,27 +62,28 @@ func main() {
...
-	operatorStatus.SetProgressing(fmt.Sprintf("Version %s of the operator is being deployed", operatorStatus.GetVersion()))
-
 	log.Print("Registering Components.")
 
 	catalogsourceconfig.InitializeStaticSyncer(mgr.GetClient(), initialWait)
 	registrySyncer := operatorsource.NewRegistrySyncer(mgr.GetClient(), initialWait, resyncInterval, updateNotificationSendWait, catalogsourceconfig.Syncer, catalogsourceconfig.Syncer)
 
+	// monitorStopCh is used to send a signal to stop reporting cluteroperator status
+	monitorStopCh := make(chan struct{})
+	monitorDoneCh := status.StartReporting(cfg, mgr, namespace, os.Getenv("RELEASE_VERSION"), monitorStopCh)
```
aravindhp: Nit: Given you commented about `monitorStopCh` you should comment about `monitorDoneCh` does.

pkg/status/status.go
```diff
@@ -191,12 +269,99 @@ func (s *status) setStatusCondition(condition configv1.ClusterStatusConditionTyp
...
+		if err != nil {
+			return fmt.Errorf("Error %v updating ClusterOperator", err)
+		}
+		// Log status change
+		log.Infof(fmt.Sprintf("[status] Set ClusterOperator condition: %s message: %s", updatedCondition, message))
+	}
+	return err
+}
+
+// syncChannelReceiver will listen on the sync channel and update the status'
```
aravindhp: Nit: trailing '

pkg/status/status.go
```diff
@@ -191,12 +269,99 @@ func (s *status) setStatusCondition(condition configv1.ClusterStatusConditionTyp
...
+	}()
+	for {
+		select {
+		case <-s.stopCh:
+			// If the stopCh is closed, the operator will exit and CO should
+			// be set to failing.
+			s.setFailing("Operator exited")
+			return
+		// Attempt to update the status every 20 seconds
+		case <-time.After(20 * time.Second):
```
aravindhp: Nit: 20 seconds should be a constant

pkg/status/status.go
```diff
@@ -191,12 +269,99 @@ func (s *status) setStatusCondition(condition configv1.ClusterStatusConditionTyp
...
+				// wait until the operator has reconciled at least one sync
+				_, syncEvents := s.syncRatio.GetSyncs()
+				if syncEvents >= minSyncsBeforeReporting {
+					// update the status with the appropriate state
+					isSucceeding, err := s.syncRatio.IsSucceeding()
+					if err == nil {
+						if isSucceeding {
+							statusErr = s.setAvailable(fmt.Sprintf("%s is available", s.version))
+						} else {
+							statusErr = s.setFailing("The operator is not meeting the expected sync ratio.")
```
aravindhp: It will help if you add useful data in the message.

pkg/status/syncratio.go
```diff
@@ -0,0 +1,120 @@
...
+func NewSyncRatio(successRatio float32, syncsBeforeTruncate int, syncTruncateValue int) (SyncRatio, error) {
+	errMsg := ""
+	if successRatio < 0 || successRatio > 1 {
+		errMsg += "successRatio must be greater than or equal to 0 and less than or equal to 1. "
+	}
+	if syncsBeforeTruncate <= 0 {
+		errMsg += "syncsBeforeTruncate must be greater than 0. "
+	}
+	if syncTruncateValue <= 0 {
+		errMsg += "syncTruncateValue must be greater than 0."
```
aravindhp: What is the point in all these error message when you don't do anything with them in the call site?

kevinrizza: +1

awgreene: That's a good point. Since we are construct the syncRatio with constants in the status package, an error would only happen when a developer changes the constants. I could update the call site to log a fatal error if an error is returned.

pkg/status/status.go
awgreene: I will update the message to provide better information about the sync ratio.

Planned Update:
current sync ratio (%g) does not meet the expected success threshold (%g)

kevinrizza:
/lgtm

for posterity :)

PR #136, opened 2019-03-14
Bug 1688703: Properly handle delete failures

kevinrizza:
Problem:
This recent change
(https://github.com/operator-framework/operator-marketplace/pull/127)
introduced a reduced scope to our cluster role. However, when deleting
`CatalogSourceConfigs` the current implementation of the delete
reconciler still required the cluster roles that were removed.
Because of this, the reconciler was producing an error and prevented the
`csc` from being deleted. Additionally, the operator was panicking
because the transitioner logic expected an object to return in that case

Solution:
Update the `deleteCreatedResources` method in the `deletedReconciler` to
use a namespace option when getting the lists of services, deployments,
rolebindings, roles and service accounts.

Update the`Reconcile` method in both the csc's and the opsrc's
`deletedReconciler` to always return the `out` object in the case of
error, as well as return the current phase to force a retry in the case
of failure.

Update both handlers to check that the `out` object is set before
attempting to transition. If it is not set, just return.

This fixes bug https://bugzilla.redhat.com/show_bug.cgi?id=1688703

### DIFF
--- a/pkg/catalogsourceconfig/deleted.go
+++ b/pkg/catalogsourceconfig/deleted.go
@@ -5,6 +5,7 @@ import (
 
 	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/operator-framework/operator-marketplace/pkg/apis/marketplace/v1alpha1"
+	"github.com/operator-framework/operator-marketplace/pkg/phase"
 	log "github.com/sirupsen/logrus"
 	apps "k8s.io/api/apps/v1"
 	core "k8s.io/api/core/v1"
@@ -54,9 +55,11 @@ func (r *deletedReconciler) Reconcile(ctx context.Context, in *v1alpha1.CatalogS
 	r.cache.Evict(out)
 
 	// Delete all created resources
-	err = r.deleteCreatedResources(ctx, in.Name, in.Namespace)
+	err = r.deleteCreatedResources(ctx, in.Name, in.Namespace, in.Spec.TargetNamespace)
 	if err != nil {
-		return nil, nil, err
+		// Something went wrong before we removed the finalizer, let's retry.
+		nextPhase = phase.GetNextWithMessage(in.Status.CurrentPhase.Name, err.Error())
+		return
 	}
 
 	// Remove the csc finalizer from the object.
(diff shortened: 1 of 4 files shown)


pkg/catalogsourceconfig/deleted.go
```diff
@@ -82,11 +89,13 @@ func (r *deletedReconciler) deleteCreatedResources(ctx context.Context, name, na
 		CscOwnerNamespaceLabel: namespace,
 	}
 	labelSelector := labels.SelectorFromSet(labelMap)
-	options := &client.ListOptions{LabelSelector: labelSelector}
+	catalogSourceOptions := &client.ListOptions{LabelSelector: labelSelector}
```
aravindhp: Can't we also restrict this to the namespace where we know the `CatalogSource` is present by looking at the `targetNamespace`?

kevinrizza: That is true. It also means (and this is true for all of the resources we are doing this to) if they are moved to a different namespace without us being made aware of it, we won't be able to delete them. We won't error because we are using List() instead of Get(), but we will miss them.

@aravindhp @tkashem thoughts? What's the better scenario?

aravindhp: If resources got moved to other namespaces without us being aware, we are going to have other issues. My vote is for that scenario to be tackled when we starting watching our resources.

pkg/catalogsourceconfig/deleted.go
```diff
@@ -56,7 +57,9 @@ func (r *deletedReconciler) Reconcile(ctx context.Context, in *v1alpha1.CatalogS
 	// Delete all created resources
 	err = r.deleteCreatedResources(ctx, in.Name, in.Namespace)
 	if err != nil {
-		return nil, nil, err
+		// Something went wrong before we removed the finalizer, let's retry.
```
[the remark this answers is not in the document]
SamiSousa: @tkashem I think it's already grammatically correct. Do you mean that the finalizer isn't removed yet at this point?

pkg/catalogsourceconfig/deleted.go
kevinrizza: @aravindhp updated.

kevinrizza:
@aravindhp @tkashem updated, please review again.

pkg/catalogsourceconfig/deleted.go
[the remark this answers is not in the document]
kevinrizza: I think it's correct language, I suppose

`before we got to the point where we were going to remove`

Is more clear, but it also seems more verbose than is needed. IMO it is fine as is, but if everyone disagrees I can update it to something else.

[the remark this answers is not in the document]
aravindhp: I am ambivalent on this.

kevinrizza:
> Please replace the commit header with something more than `Problem:`. I am okay with a repeat of the PR header

@aravindhp Yep, fixed the header on the commit.

PR #137, opened 2019-03-15
Bug 1686121: Update Marketplace management of clusterOperator conditions

awgreene:
* Report version only after operator becomes available.

This fixes https://bugzilla.redhat.com/show_bug.cgi?id=1686121


awgreene:
This PR was created to get a quick fix in for [this bug](https://bugzilla.redhat.com/show_bug.cgi?id=1686121). However, the bug is no longer time sensetive and the change is captured in #134

PR #138, opened 2019-03-19
Bug 1686668: [csc] Ignore invalid values when reading `csc.Spec.Packages`

kevinrizza:
Problem:
In some cases, when the Openshift UI updates the package list of a csc
after removing a subscription, the update does not properly format the
package list. For example:

`packages: ',strimzi-kafka-operator-test'`

In this example, the package list accidentally includes a leading comma.

Right now, the marketplace doesn't handle poorly formatted package
lists in the csc spec. In general, if a user accidentally leaves a comma
behind the operator should ignore it and proceed as normal with the
list of valid packages.

Solution:
Any time the operator reads the package list, that read happens
from a public method that formats the package list and excludes invalid
packages (specifically empty string).
Added a private method `getValidPackageSliceFromString()` that takes a
package string and removes invalid elements from it.
Updated the two `GetPackages()` methods on the csc and spec types to
call this method before returning.
Added a new method on the csc type called `GetPackages()` that uses the
validation logic to reconstruct the `Spec.Packages` string.
Updated every case where we were referencing the `Spec.Packages` field
directly and replaced it with one of these methods.

### DIFF
--- a/pkg/apis/marketplace/v1alpha1/catalogsourceconfig_types.go
+++ b/pkg/apis/marketplace/v1alpha1/catalogsourceconfig_types.go
@@ -100,9 +100,14 @@ func (csc *CatalogSourceConfig) EnsurePublisher() {
 	}
 }
 
+func (csc *CatalogSourceConfig) GetPackages() string {
+	pkgs := getValidPackageSliceFromString(csc.Spec.Packages)
+	return strings.Join(pkgs, ",")
+}
+
 // GetPackageIDs returns the list of package(s) specified.
 func (csc *CatalogSourceConfig) GetPackageIDs() []string {
-	return strings.Split(strings.Replace(csc.Spec.Packages, " ", "", -1), ",")
+	return getValidPackageSliceFromString(csc.Spec.Packages)
 }
 
 // GetTargetNamespace returns the TargetNamespace where the OLM resources will
@@ -129,5 +134,19 @@ func (csc *CatalogSourceConfig) RemoveOwner(ownerUID types.UID) {
 
 // GetPackageIDs returns the list of package(s) specified.
 func (spec *CatalogSourceConfigSpec) GetPackageIDs() []string {
-	return strings.Split(strings.Replace(spec.Packages, " ", "", -1), ",")
+	return getValidPackageSliceFromString(spec.Packages)
+}
+
+func getValidPackageSliceFromString(pkgs string) []string {
+	pkgIds := make([]string, 0)
+
+	pkgSlice := strings.Split(strings.Replace(pkgs, " ", "", -1), ",")
+
+	for _, pkg := range pkgSlice {
+		if pkg != "" {
+			pkgIds = append(pkgIds, pkg)
+		}
+	}
+
+	return pkgIds
 }
--- a/pkg/catalogsourceconfig/cache.go
+++ b/pkg/catalogsourceconfig/cache.go
@@ -89,7 +89,7 @@ func (c *cache) Evict(csc *v1alpha1.CatalogSourceConfig) {
 
 func (c *cache) Set(csc *v1alpha1.CatalogSourceConfig) {
 	c.entries[csc.ObjectMeta.UID] = &v1alpha1.CatalogSourceConfigSpec{
-		Packages:        csc.Spec.Packages,
+		Packages:        csc.GetPackages(),
 		TargetNamespace: csc.Spec.TargetNamespace,
 	}
 }
--- a/pkg/catalogsourceconfig/registry.go
+++ b/pkg/catalogsourceconfig/registry.go
@@ -96,7 +96,7 @@ func (r *registry) GetAddress() string {
 // ensureDeployment ensures that registry Deployment is present for serving
 // the the grpc interface for the packages from the given operatorSources
 func (r *registry) ensureDeployment(operatorSources string) error {
-	registryCommand := getCommand(r.csc.Spec.Packages, operatorSources)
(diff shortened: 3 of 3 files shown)


pkg/apis/marketplace/v1alpha1/catalogsourceconfig_types.go
```diff
@@ -129,5 +134,19 @@ func (csc *CatalogSourceConfig) RemoveOwner(ownerUID types.UID) {
...
-	return strings.Split(strings.Replace(spec.Packages, " ", "", -1), ",")
+	return getValidPackageSliceFromString(spec.Packages)
+}
+
+func getValidPackageSliceFromString(pkgs string) []string {
+	pkgIds := make([]string, 0)
+
+	pkgSlice := strings.Split(strings.Replace(pkgs, " ", "", -1), ",")
+
+	for _, pkg := range pkgSlice {
```
aravindhp: Can't you just trim the string instead of doing this? This can be done right in `GetPackageIDs()`. Or is there another reason for this?

pkg/catalogsourceconfig/registry.go
```diff
@@ -96,7 +96,7 @@ func (r *registry) GetAddress() string {
 // ensureDeployment ensures that registry Deployment is present for serving
 // the the grpc interface for the packages from the given operatorSources
 func (r *registry) ensureDeployment(operatorSources string) error {
-	registryCommand := getCommand(r.csc.Spec.Packages, operatorSources)
+	registryCommand := getCommand(r.csc.GetPackages(), operatorSources)
```
aravindhp: You should also fix `getCommand()` and have all that string fixup happen in just one place.

pkg/apis/marketplace/v1alpha1/catalogsourceconfig_types.go
tkashem: I would recommend trimming white spaces from each package name. After trimming, if the package name is not empty we add it.

pkg/catalogsourceconfig/cache.go
```diff
@@ -89,7 +89,7 @@ func (c *cache) Evict(csc *v1alpha1.CatalogSourceConfig) {
 
 func (c *cache) Set(csc *v1alpha1.CatalogSourceConfig) {
 	c.entries[csc.ObjectMeta.UID] = &v1alpha1.CatalogSourceConfigSpec{
-		Packages:        csc.Spec.Packages,
+		Packages:        csc.GetPackages(),
```
aravindhp: Shouldn't we just leave this as the string we got as input?

pkg/apis/marketplace/v1alpha1/catalogsourceconfig_types.go
```diff
@@ -129,5 +134,19 @@ func (csc *CatalogSourceConfig) RemoveOwner(ownerUID types.UID) {
...
 // GetPackageIDs returns the list of package(s) specified.
 func (spec *CatalogSourceConfigSpec) GetPackageIDs() []string {
-	return strings.Split(strings.Replace(spec.Packages, " ", "", -1), ",")
+	return getValidPackageSliceFromString(spec.Packages)
+}
+
+func getValidPackageSliceFromString(pkgs string) []string {
+	pkgIds := make([]string, 0)
+
+	pkgSlice := strings.Split(strings.Replace(pkgs, " ", "", -1), ",")
```
tkashem: Is `strings.Replace` required If we sanitize each package name by trimming white spaces? For a leading comma `strings.Split` returns a package name with a space.

pkg/apis/marketplace/v1alpha1/catalogsourceconfig_types.go
kevinrizza: @aravindhp This solves the case not just for leading commas, but for any item in the list that isn't valid. Ex:

`etcd,,prometheus`
`,etcd,prometheus,`
`,etcd,,,prometheus,`

etc.

Ultimately this fix ensures that, since we are `split`ting on `","` we should ensure that we don't include any accidental additional elements.

This also means that, in the future, we could put further validation in this one method -- for example, if we want to restrict other special characters or key words or something else.

pkg/apis/marketplace/v1alpha1/catalogsourceconfig_types.go
kevinrizza: If we do it to the initial string once that will remove any extra whitespace, meaning we don't need to call `strings.Replace()` on each token.

pkg/catalogsourceconfig/cache.go
kevinrizza: I suppose technically it shouldn't matter, but it might be better to preserve the behavior for anything other than the simple read.

pkg/catalogsourceconfig/registry.go
kevinrizza: Addressed.

pkg/apis/marketplace/v1alpha1/catalogsourceconfig_types.go
kevinrizza: See below.

pkg/apis/marketplace/v1alpha1/catalogsourceconfig_types.go
kevinrizza: Additionally I prefer this approach because removing all of the whitespace is needed anyway. This also avoids the issue of spaces in package names.

kevinrizza:
@tkashem @aravindhp @awgreene Please take a look when you get a chance. Thanks.

PR #140, opened 2019-03-21
Bug 1691241: [csc] Don't ignore cache on restart

kevinrizza:
Problem:
Rebuilding the `CatalogSourceConfig` state relies on the cache
implementation. Currently, if the spec is updated on a csc it will
not automatically attempt to reconcile that change. Instead, the phase
factory checks if the spec has been updated from the previous update.
To accomplish this, it checks an in-memory cache that compares the
cached spec to the current spec. This is troublesome in the case where
the `marketplace` restarts, because the in memory cache is lost. This
bug is caused when an existing `catalogsourceconfig` gets an update:
the cache is empty for that `csc` but we do not return stale.

Resolution:
Update the `IsEntryStale` method on the `csc` cache. In all cases, if
the `csc` is not found in the cache we should return stale.
Also update `GetPhaseReconciler` to only check the cache is stale when
the phase is not set to `Initial` or `Configuring`, since we are already
in the process of reconciling and updating the cache in that case. This
also prevents infinite loops of constantly going to `Update`.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1691241

### DIFF
--- a/pkg/catalogsourceconfig/factory.go
+++ b/pkg/catalogsourceconfig/factory.go
@@ -38,19 +38,23 @@ type phaseReconcilerFactory struct {
 }
 
 func (f *phaseReconcilerFactory) GetPhaseReconciler(log *logrus.Entry, csc *v1alpha1.CatalogSourceConfig) (Reconciler, error) {
-	// Check if the cache is stale. A stale cache entry is an indicator that the
-	// object has changed and requires updating.
-	pkgStale, targetStale := f.cache.IsEntryStale(csc)
-	if pkgStale || targetStale {
-		return NewUpdateReconciler(log, f.client, f.cache, targetStale), nil
-	}
-
 	// If the object has a deletion timestamp, it means it has been marked for
 	// deletion. Return a deleted reconciler to remove that csc data from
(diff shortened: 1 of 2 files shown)


awgreene:
This looks good @kevinrizza

pkg/catalogsourceconfig/factory.go
```diff
@@ -39,10 +39,14 @@ type phaseReconcilerFactory struct {
 
 func (f *phaseReconcilerFactory) GetPhaseReconciler(log *logrus.Entry, csc *v1alpha1.CatalogSourceConfig) (Reconciler, error) {
```
tkashem: Shouldn't  `if !csc.ObjectMeta.DeletionTimestamp.IsZero() {` be at the top?

kevinrizza: Probably true, but my thinking is that such a change is outside the scope of this pr.

kevinrizza: Oh, I actually see what you're saying. We should also probably check that we shouldn't go to the deleted phase first.

PR #141, opened 2019-03-22
Bug 1691241 [csc] Prevent getting stuck in current state when cache is invalid for csc

kevinrizza:
Problem:
When the service restarts, we lose all cached knowledge about the
catalogsourceconfig, including the old targetNamespace. This is an
issue, because if the targetNamespace changed we want to remove the
catalogsource in the old targetNamespace and replace it with a new one.
Now that we are always assuming the cache is invalid if it is empty
(which is a valid assumption), we still don't have a way of cleaning up
old targetNamespace if it changed. Right now though, we will return an
error and not attempt to reconcile the CSC.

Not only that, but this error would also cause the catalogsourceconfig
to get into a stuck state. When the error was returned, we moved back
into the current state to retry. If the state was `success` or `failure`
the `catalogsourceconfig` would simply loop in that phase until the
status was dropped or the syncer set the status to initial.

Solution:
If the cache is invalid we cannot assume that the targetNamespace is
actually invalid. For now, we should just print a debug log and return
and continue to reconcile the csc. Once we have a better method
of retaining references to owned objects, we can handle this edge case
better.

Additionally, update the `deleteObjects` function to no longer return
an error. We do not want to cycle forever attempting to clean up the
`catalogsource` if we are unable to, since that would cause our
`catalogsourceconfig` to be stuck in a bad state looping on the state
until the syncer restarted the reconcilliation loop or through manual
intervention.

This issue was uncovered by the original fix for
https://bugzilla.redhat.com/show_bug.cgi?id=1691241, and this commit
is addition to that fix.

### DIFF
--- a/pkg/catalogsourceconfig/update.go
+++ b/pkg/catalogsourceconfig/update.go
@@ -2,7 +2,6 @@ package catalogsourceconfig
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/operator-framework/operator-marketplace/pkg/apis/marketplace/v1alpha1"
 	"github.com/operator-framework/operator-marketplace/pkg/phase"
@@ -37,13 +36,10 @@ func (r *updateReconciler) Reconcile(ctx context.Context, in *v1alpha1.CatalogSo
 
 	// The TargetNamespace of the CatalogSourceConfig object has changed
 	if r.targetChanged {
-		// Delete the objects in the old TargetNamespace
-		err = r.deleteObjects(in)
-		if err != nil {
-			// Next retry should resume from the current phase.
-			nextPhase = phase.GetNextWithMessage(in.Status.CurrentPhase.Name, err.Error())
-			return
-		}
+		// Best case attempt at deleting the objects in the old TargetNamespace
+		// If the csc is not cached we don't want to fail because there are
+		// cases where we won't be able to find the objects.
+		r.deleteObjects(in)
 	}
 
 	// Remove it from the cache so that it does not get picked up during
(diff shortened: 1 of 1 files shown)


pkg/catalogsourceconfig/update.go
```diff
@@ -62,10 +61,17 @@ func (r *updateReconciler) Reconcile(ctx context.Context, in *v1alpha1.CatalogSo
 // deleteObjects deletes the CatalogSource in the old TargetNamespace.
 func (r *updateReconciler) deleteObjects(in *v1alpha1.CatalogSourceConfig) error {
 	cachedCSCSpec, found := r.cache.Get(in)
-	// This should never happen as it is because the cached Spec has changed
+	// This generally won't happen as it is because the cached Spec has changed
 	// that we are in the update reconciler.
```
aravindhp: Nit: `when we are in`

pkg/catalogsourceconfig/update.go
```diff
@@ -62,10 +61,17 @@ func (r *updateReconciler) Reconcile(ctx context.Context, in *v1alpha1.CatalogSo
...
+	// However, in the case that the cache is invalidated because of a restart,
+	// we could simply be unsure if the targetNamespace changed. Ideally,  we
+	// we should be using something like a direct reference to the old object
+	// instead of an in memory cache, since when the service restarts we will be
+	// unable to find the old object if the targetNamespace changed while we were
+	// down. For now, we will just ignore it.
```
aravindhp: Like we discussed offline, this doesn't seem to be the issue @emmajiafan is talking about in the BZ.

pkg/catalogsourceconfig/update.go
```diff
@@ -62,10 +61,17 @@ func (r *updateReconciler) Reconcile(ctx context.Context, in *v1alpha1.CatalogSo
...
 	if !found {
```
tkashem: Question: I am trying to understand what real world scenario is causing this if block to be executed. We detected that the target namespace has changed ( `r.targetChanged` == true ) but the spec is not in cache.

kevinrizza: @tkashem because the cache invalidator logic changed. Previously we were cheating by saying that if we didn't find the `csc` in the cache that we should assume that everything is fine and ignore the cache. To be able to force a resync I updated that logic to assume that if the `csc` wasn't in the cache that the cache is not valid (which is an accurate assumption).

The problem is, we genuinely don't know if the targetNamespace was changed if we lose track of the cache. So we should say it's invalid and attempt to cleanup, but we don't have a way of finding the old one.

pkg/catalogsourceconfig/update.go
kevinrizza: Updated.

pkg/catalogsourceconfig/update.go
kevinrizza: This is actually not the case, we are seeing this error because the cache error forces the csc into a bad state that won't be resolved unless the status is dropped. By preventing this bad state, we won't have to wait for the syncer to help us.

kevinrizza:
@aravindhp @tkashem addressed comments, please take another look.

pkg/catalogsourceconfig/update.go
tkashem: Yeah, if the cache is missing then we can't answer the question whether the target namespace has changed. We will need to persist the namespaced name of `catsrc` in the future in `annotations` or something.

kevinrizza:
@aravindhp @tkashem I removed the unused error returned from the `updateReconciler`'s `deleteObjects()` function, which is technically what caused the endless reconcile loop described in the BZ. I also updated the comments in the source and the git commit to reflect that change.

kevinrizza:
@tkashem @aravindhp can someone also give this PR an lgtm so that can merge once the tests are complete?

aravindhp:
I already gave `/approve`. @tkashem please review and give an lgtm.

PR #145, opened 2019-03-25
[e2e] Test CSC with non-existant targetNamespace

awgreene:
Implement a test that:
* Creates a CatalogSourceConfig with non-existent targetNamespace
* Ensures that the status is not succeeded
* Creates the targetNamespace
* Ensures that the status is succeeded

Create a testutils file that contains methods used by various tests

### DIFF
--- a/test/e2e/csc_tests.go
+++ b/test/e2e/csc_tests.go
@@ -1,0 +1,133 @@
+package e2e
+
+import (
+	"fmt"
+	"testing"
+
+	operator "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	"github.com/operator-framework/operator-sdk/pkg/test"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	// nonExistingTargetNamespaceCSCName is the name of the catalogsourceconfig that points
+	// to a non-existing targetNamespace
+	nonExistingTargetNamespaceCSCName string = "non-existing-namespace-operators"
+
+	// targetNamespace is the non-existing target namespace
+	targetNamespace string = "non-existing-namespace"
+)
+
+func runCSCWithNonExistingTargetNamespace(t *testing.T) {
+	ctx := test.NewTestCtx(t)
+	defer ctx.Cleanup()
+
+	// Get global framework variables
+	f := test.Global
+	// Run tests
+	if err := cscWithNonExistingTargetNamespace(t, f, ctx); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func cscWithNonExistingTargetNamespace(t *testing.T, f *test.Framework, ctx *test.TestCtx) error {
+	// Get test namespace
+	namespace, err := ctx.GetNamespace()
+	if err != nil {
+		return fmt.Errorf("Could not get namespace: %v", err)
+	}
+
+	// Create the operatorsource to download the manifests
+	testOperatorSource := &operator.OperatorSource{
+		TypeMeta: metav1.TypeMeta{
+			Kind: operator.OperatorSourceKind,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-operators",
+			Namespace: namespace,
+		},
+		Spec: operator.OperatorSourceSpec{
+			Type:              "appregistry",
+			Endpoint:          "https://quay.io/cnr",
+			RegistryNamespace: "marketplace_e2e",
+		},
+	}
+	err = createRuntimeObject(f, ctx, testOperatorSource)
+	if err != nil {
+		return err
+	}
+
+	// Create a new catalogsourceconfig with a non-existing targetNamespace
+	nonExistingTargetNamespaceCSC := &operator.CatalogSourceConfig{
+		TypeMeta: metav1.TypeMeta{
+			Kind: operator.OperatorSourceKind,
+		}, ObjectMeta: metav1.ObjectMeta{
+			Name:      nonExistingTargetNamespaceCSCName,
+			Namespace: namespace,
+		},
+		Spec: operator.CatalogSourceConfigSpec{
+			TargetNamespace: targetNamespace,
+			Packages:        "descheduler",
+		}}
+	err = createRuntimeObject(f, ctx, nonExistingTargetNamespaceCSC)
+	if err != nil {
+		return err
+	}
+
+	// Check that we created the catalogsourceconfig with a non-existing targetNamespace
+	resultCatalogSourceConfig := &operator.CatalogSourceConfig{}
+	err = WaitForResult(t, f, resultCatalogSourceConfig, namespace, nonExistingTargetNamespaceCSCName)
+	if err != nil {
+		return err
+	}
+
+	// Check if the catalogsourceconfig phase and message are the expected values
+	expectedPhase := "Configuring"
+	expectedMessage := fmt.Sprintf("namespaces \"%s\" not found", targetNamespace)
+	// Check that the catalogsourceconfig with an non-existing targetNamespace eventually reaches the
+	// configuring phase with the expected message
+	err = wait.Poll(retryInterval, timeout, func() (bool, error) {
+		// catalogsourceconfig should exist so no wait
+		err = WaitForResult(t, f, resultCatalogSourceConfig, namespace, nonExistingTargetNamespaceCSCName)
+		if err != nil {
+			return false, err
+		}
+		if resultCatalogSourceConfig.Status.CurrentPhase.Name == expectedPhase &&
+			resultCatalogSourceConfig.Status.CurrentPhase.Message == expectedMessage {
+			return true, nil
+		}
+		return false, nil
+	})
+	if err != nil {
+		return fmt.Errorf("CatalogSourceConfig never reached expected phase/message, expected %v/%v", expectedPhase, expectedMessage)
+	}
+
+	// Create a namespace based on the targetNamespace string
+	targetNamespaceRuntimeObject := &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: targetNamespace}}
+	err = createRuntimeObject(f, ctx, targetNamespaceRuntimeObject)
+	if err != nil {
+		return err
+	}
+
+	// Now that the targetNamespace has been created, periodically check that the catalogsourceconfig
+	// has reached the Succeeded phase
+	expectedPhase = "Succeeded"
+	err = wait.Poll(retryInterval, timeout, func() (bool, error) {
+		// catalogsourceconfig should exist so no wait
+		err = WaitForResult(t, f, resultCatalogSourceConfig, namespace, nonExistingTargetNamespaceCSCName)
+		if err != nil {
+			return false, err
+		}
+		if resultCatalogSourceConfig.Status.CurrentPhase.Name == expectedPhase {
+			return true, nil
+		}
+		return false, nil
+	})
+	if err != nil {
+		return fmt.Errorf("CatalogSourceConfig never reached expected phase/message, expected %v", expectedPhase)
+	}
+
+	return nil
+}
--- a/test/e2e/helpers.go
+++ b/test/e2e/helpers.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/operator-framework/operator-sdk/pkg/test"
-
 	apps "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -15,8 +14,10 @@ import (
 )
 
 const (
-	retryInterval = time.Second * 5
-	timeout       = time.Minute * 5
+	retryInterval        = time.Second * 5
+	timeout              = time.Minute * 5
+	cleanupRetryInterval = time.Second * 1
+	cleanupTimeout       = time.Second * 30
 )
 
 // WaitForResult polls the cluster for a particular resource name and namespace
@@ -70,4 +71,16 @@ func WaitForSuccessfulDeployment(t *testing.T, f *test.Framework, deployment app
 	}
 	t.Logf("Deployment %s has been initialized successfully\n", deployment.Name)
 	return nil
+}
+
+// createRuntimeObject creates a runtime object using the test framework
+func createRuntimeObject(f *test.Framework, ctx *test.TestCtx, obj runtime.Object) error {
+	return f.Client.Create(
+		context.TODO(),
+		obj,
+		&test.CleanupOptions{
+			TestContext:   ctx,
+			Timeout:       cleanupTimeout,
+			RetryInterval: cleanupRetryInterval,
+		})
 }
--- a/test/e2e/marketplace_test.go
+++ b/test/e2e/marketplace_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"testing"
-	"time"
 
 	"github.com/operator-framework/operator-marketplace/pkg/apis"
 	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
@@ -19,9 +18,6 @@ import (
 )
 
 const (
-	cleanupRetryInterval = time.Second * 1
-	cleanupTimeout       = time.Second * 5
-
 	GroupLabel string = "opsrc-group"
 )
 
@@ -64,6 +60,7 @@ func TestMarketplace(t *testing.T) {
 	// run subtests
 	t.Run("marketplace-group", func(t *testing.T) {
 		t.Run("Cluster", MarketplaceCluster)
+		t.Run("csc-non-existing-namespace", runCSCWithNonExistingTargetNamespace)
 	})
 }
 


test/e2e/marketplace_test.go
```diff
@@ -123,9 +119,6 @@ func defaultCreateTest(t *testing.T, f *test.Framework, ctx *test.TestCtx) error
 			Timeout:       cleanupTimeout,
 			RetryInterval: cleanupRetryInterval,
 		})
```
kevinrizza: why are we removing this error check?

test/e2e/testutils.go
```diff
@@ -0,0 +1,132 @@
...
+			Namespace: namespace,
+		},
+		Spec: operator.CatalogSourceConfigSpec{
+			TargetNamespace: targetNamespace,
+			Packages:        packages,
+		}}
+}
+
+// createOperatorSource returns a OperatorSource based on the provided parameters
+func createOperatorSource(operatorSourceName string, namespace string, labels map[string]string, registryNamespace string) *operator.OperatorSource {
```
kevinrizza: What is the value of this method? What if we wanted to create an operator source with a different name? Or an invalid namespace? IMO if there was logic in this method it would be one thing, but we could just use this syntax inline in the method and have more control without much more verbosity.

test/e2e/testutils.go
```diff
@@ -0,0 +1,132 @@
...
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	cleanupRetryInterval = time.Second * 1
+	cleanupTimeout       = time.Second * 30
+)
+
+// createCatalogSourceConfig returns a CatalogSourceConfig based on the provided parameters
+func createCatalogSourceConfig(catalogSourceConfigName string, namespace string, targetNamespace string, packages string) *operator.CatalogSourceConfig {
```
kevinrizza: Same comment as the one about createOperatorSource

test/e2e/testutils.go
```diff
@@ -0,0 +1,132 @@
...
+		})
+}
+
+// pollUntilTrue will check if a function returns a true value
+// every pollInterval duration until the timeout duration has
+// passed
+//
+// pollUntilTrue function will run over the timeout if the
+// provided function hangs
+func pollUntilTrue(pollInterval time.Duration, timeout time.Duration, f func() bool) bool {
```
kevinrizza: Don't we already have a very similar method to this in helpers.go?

test/e2e/testutils.go
```diff
@@ -0,0 +1,132 @@
+package e2e
```
kevinrizza: Why don't we use the `helpers.go` file for this? Why do we need a second utilities/helpers file?

test/e2e/marketplace_test.go
```diff
@@ -74,7 +71,6 @@ func MarketplaceCluster(t *testing.T) {
 
 	// get global framework variables
```
kevinrizza: If we are going to have multiple tests and multiple files for the tests, wouldn't it be better to move this basic workflow test into a separate file as well?

test/e2e/csc_tests.go
```diff
@@ -0,0 +1,109 @@
...
+	registryNamespace := "marketplace_e2e"
+
+	// Get test namespace
+	namespace, err := ctx.GetNamespace()
+	if err != nil {
+		return fmt.Errorf("Could not get namespace: %v", err)
+	}
+
+	// Create the operatorsource to download the manifests
+	testOperatorSource := createOperatorSource(testResourceName, namespace, nil, registryNamespace)
```
kevinrizza: For CatalogSourceConfig tests, if we are running these in serial, would it be better if we did some upfront setup like create one operatorsource beforehand, and then use the same one for all the tests? Rather than create all these resources each time for each test?

test/e2e/csc_tests.go
```diff
@@ -0,0 +1,109 @@
...
+	// Check that we created the catalogsourceconfig with an invalid targetNamespace
+	err = WaitForResult(t, f, resultCatalogSourceConfig, namespace, testResourceName)
+	if err != nil {
+		return err
+	}
+
+	// Check if the catalogsourceconfig phase and message are the expected values
+	expectedPhase := "Configuring"
+	expectedMessage := fmt.Sprintf("namespaces \"%s\" not found", targetNamespace)
+	result := pollUntilTrue(5*time.Second, 1*time.Minute, func() bool {
```
kevinrizza: Don't we have these timeouts hardcoded elsewhere?

test/e2e/testutils.go
awgreene: In the helpers.go there exists a tool to check that a runtime object exists. There is not a way to check that a runtime Object has reached a certain state.

test/e2e/testutils.go
awgreene: Good point - I'll move any methods into the existing `helpers.go` file

test/e2e/marketplace_test.go
awgreene: I am planning to create another PR that would refactor the basic workflow test into it's own file. I thought it made sense to limit the scope of this PR to introducing the csc test.

test/e2e/marketplace_test.go
awgreene: This should not have been removed - my mistake.

test/e2e/csc_tests.go
awgreene: This is certainly something I had considered. I plan to do a follow up PR that refactors existing tests which would include this work.

test/e2e/csc_tests.go
awgreene: There are constants included in the testutils package that define the interval at which we attempt to create a runtime object and the duration we should attempt to do so for. I could update these calls to use those values but the timeout value would decrease which means the tests may fail more. At the very least - the values should be updated to be constants.

test/e2e/testutils.go
awgreene: I see your point. I will remove the methods in the interest of having more control over the objects we create for the tests.

test/e2e/testutils.go
awgreene: Please see comment on createOperatorSource.

test/e2e/testutils.go
awgreene: Moved.

test/e2e/csc_tests.go
aravindhp: Yes, please add constants with comments for these.

test/e2e/csc_tests.go
```diff
@@ -0,0 +1,133 @@
...
+			Kind: operator.OperatorSourceKind,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-operators",
+			Namespace: namespace,
+		},
+		Spec: operator.OperatorSourceSpec{
+			Type:              "appregistry",
+			Endpoint:          "https://quay.io/cnr",
+			RegistryNamespace: "marketplace_e2e",
```
aravindhp: Given we are going to constantly add and remove this `OperatorSource`, please reduce the number of operators in the namespace.

test/e2e/csc_tests.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// Check that we created the catalogsourceconfig
+	resultCatalogSourceConfig := &operator.CatalogSourceConfig{}
+	err = WaitForResult(t, f, resultCatalogSourceConfig, namespace, testResourceName)
+	if err != nil {
+		return err
+	}
+
+	// Delete the existing catalogsourceconfig
+	err = deleteRuntimeObject(f, resultCatalogSourceConfig, testResourceName, namespace)
```
aravindhp: Why are you deleting the datastore `CatalogSourceConfig`?

test/e2e/csc_tests.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+		return false
+	})
+
+	if !result {
+		return fmt.Errorf("CatalogSourceConfig never reached expected phase/message, expected %v/%v got %v/%v", expectedPhase, expectedMessage, resultCatalogSourceConfig.Status.CurrentPhase.Name, resultCatalogSourceConfig.Status.CurrentPhase.Message)
+	}
+
+	// Create the nonexisting targetNamespace
+	err = createNamespace(f, ctx, targetNamespace)
```
aravindhp: Does the `e2e` launched using the SDK use a super user account? Because the operator does not have `Roles` that allow it to interact with namespaces. This seems to imply that there is no way for us to verify if we are adding the correct `Roles` for the operator.

test/e2e/helpers.go
```diff
@@ -71,3 +76,80 @@ func WaitForSuccessfulDeployment(t *testing.T, f *test.Framework, deployment app
...
+	// Delete the runtime object
+	err := f.Client.Delete(
+		context.TODO(),
+		obj)
+	if err != nil {
+		return fmt.Errorf("Error deleting obj: %v", err)
+	}
+
+	// Confirm that the runtime object was deleted
+	namespacedName := types.NamespacedName{Name: name, Namespace: namespace}
```
aravindhp: Why is this function also confirming it was deleted while `createRuntimeObject` does only the creation? What if I have a use where I just want to delete an object and not care if it actually got deleted?

test/e2e/helpers.go
```diff
@@ -71,3 +76,80 @@ func WaitForSuccessfulDeployment(t *testing.T, f *test.Framework, deployment app
...
+		return nil
+	}
+	return fmt.Errorf("Runtime object %v was never deleted", obj)
+}
+
+// createNamespace creates a namespace
+func createNamespace(f *test.Framework, ctx *test.TestCtx, namespace string) error {
+	obj := &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}
+	// Create the operatorsource to download the manifests
+	return f.Client.Create(
```
aravindhp: Please reuse `createRuntimeObject()` here?

test/e2e/helpers.go
```diff
@@ -71,3 +76,80 @@ func WaitForSuccessfulDeployment(t *testing.T, f *test.Framework, deployment app
...
+		})
+}
+
+// pollUntilTrue will check if a function returns a true value
+// every pollInterval duration until the timeout duration has
+// passed
+//
+// pollUntilTrue function will run over the timeout if the
+// provided function hangs
+func pollUntilTrue(pollInterval time.Duration, timeout time.Duration, f func() bool) bool {
```
aravindhp: Are you sure a similar helper doesn't exist in the SDK or k8s libraries?

test/e2e/marketplace_test.go
```diff
@@ -32,14 +28,14 @@ func TestMarketplace(t *testing.T) {
 		TypeMeta: metav1.TypeMeta{
 			Kind: marketplace.OperatorSourceKind,
 			APIVersion: fmt.Sprintf("%s/%s",
-			marketplace.SchemeGroupVersion.Group, marketplace.SchemeGroupVersion.Version),
+				marketplace.SchemeGroupVersion.Group, marketplace.SchemeGroupVersion.Version),
```
aravindhp: Formatting fixes should be a separate PR

test/e2e/marketplace_test.go
```diff
@@ -32,14 +28,14 @@ func TestMarketplace(t *testing.T) {
...
 		},
 	}
 	catalogsourceconfig := &marketplace.CatalogSourceConfig{
 		TypeMeta: metav1.TypeMeta{
 			Kind: marketplace.CatalogSourceConfigKind,
 			APIVersion: fmt.Sprintf("%s/%s",
-			marketplace.SchemeGroupVersion.Group, marketplace.SchemeGroupVersion.Version),
+				marketplace.SchemeGroupVersion.Group, marketplace.SchemeGroupVersion.Version),
```
aravindhp: Formatting fixes should be a separate PR

test/e2e/marketplace_test.go
```diff
@@ -50,6 +46,7 @@ func TestMarketplace(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to add catalogsourceconfig custom resource scheme to framework: %v", err)
 	}
+
```
aravindhp: Why? Please don't add and remove empty lines because you prefer it that way. We can come up with a guideline for such things and fix all of this in a separate PR.

test/e2e/csc_tests.go
kevinrizza: @aravindhp There are two competing ideas here. the test client has similar super user rules that apply to it, similar to what `operator-sdk up` has (or akin to logging into oc as a super user), which is what is being used to create this namespace. That is separate from the operator itself having those roles, which are being tested here -- the operator is deployed using the deployment in the manifest/ folder when these tests are run in the e2e cluster.

aravindhp: So your take is that the `ClusterOperator` will be marked as failing if the roles are incorrect? We should then update our deployment check to include that and not run any other tests if that fails.

kevinrizza: It should, but I think that's outside the scope of this test -- ClusterOperator having a failing status is a catch all, so we could lose granularity on why the error is happening if we just make that blanket assumption. Better for us to ignore ClusterOperator status, or at least just mark a failure and continue the tests. What if we want to run a test to purposely make ClusterOperator set a failed status? Or run a test to see a fail case that we would ignore because we are looking at ClusterOperator?

aravindhp: Good point. I did not mean to have this be done as part of this PR. We should have a follow up PR for this. My desire is to have a test that will tell us if we messed up the `Roles` in the manifest folder for the operator.

kevinrizza: +1 I like the idea of explicit tests that check and ensure the state of OperatorStatus.

test/e2e/csc_tests.go
awgreene: Rather than creating an operator source for each test, would it be better to use a single operator source for all tests that rely on it? I had spoken to Kevin about doing so in a future refactor PR: https://github.com/operator-framework/operator-marketplace/pull/145#discussion_r269083130

aravindhp: +1

awgreene: Great - I will create a JIRA to refactor the code.

test/e2e/csc_tests.go
awgreene: After looking into this I realized that I shouldn't delete this CSC. I can just create the CSC that points to an invalid `targetNamespace`.

This also means that the `deleteRuntimeObject` method in the `helper.go` and so I will remove it.

test/e2e/helpers.go
awgreene: The SDK only provides the controller-runtime's client and the `WaitForDeployment` and `WaitForOperatorDeployment` functions in the test framework. I will look deeper into the k8s library and update this accordingly.

test/e2e/marketplace_test.go
awgreene: This was a mistake introduced by the go format when saving the file, I will revert the change.

test/e2e/marketplace_test.go
awgreene: This was a mistake introduced by the go format when saving the file, I will revert the change.

test/e2e/marketplace_test.go
awgreene: Addressed in the latest commit.

test/e2e/helpers.go
awgreene: After reviewing a comment you made earlier[1] this function was removed given that the test is not required to delete a csc.

test/e2e/helpers.go
awgreene: Addressed in the latest commit.

test/e2e/csc_tests.go
```diff
@@ -0,0 +1,133 @@
...
+
+	operator "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	"github.com/operator-framework/operator-sdk/pkg/test"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// pollInterval is used by the pollUntilTrue function to define the frequency that
+	// the provided function is ran
```
aravindhp: Not sure what you are trying to say. `function is run at`? Moreover did we make these constants so that any function that wants to poll can use it? If yes, please make the comment generic. Or is it specific to `pollUntilTrue()`? If yes, then you should move them into that function.

test/e2e/csc_tests.go
```diff
@@ -0,0 +1,133 @@
...
+	pollInterval time.Duration = 5 * time.Second
+
+	// pollTimeout is used by the pollUntilTrue function to define the timeout
+	pollTimeout time.Duration = 1 * time.Minute
```
aravindhp: Same comment here.

test/e2e/marketplace_test.go
```diff
@@ -74,7 +72,6 @@ func MarketplaceCluster(t *testing.T) {
 
 	// get global framework variables
 	f := test.Global
-
```
aravindhp: https://github.com/operator-framework/operator-marketplace/pull/145#discussion_r269143459 applies here too.

test/e2e/marketplace_test.go
```diff
@@ -126,7 +123,6 @@ func defaultCreateTest(t *testing.T, f *test.Framework, ctx *test.TestCtx) error
 	if err != nil {
 		return err
 	}
-
```
aravindhp: https://github.com/operator-framework/operator-marketplace/pull/145#discussion_r269143459 applies here too

test/e2e/helpers.go
awgreene: I have found a [similar method](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go#L305 ) in k8s apimachinery and will subsitute my function for it.

test/e2e/helpers.go
```diff
@@ -71,3 +72,37 @@ func WaitForSuccessfulDeployment(t *testing.T, f *test.Framework, deployment app
 	t.Logf("Deployment %s has been initialized successfully\n", deployment.Name)
 	return nil
 }
+
+// createRuntimeObject creates a runtime object using the test framework
```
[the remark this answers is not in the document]
aravindhp: This is already in a separate file called `helper.go`. How about a dedicated package called `helpers` rather than prefixing each function?

test/e2e/marketplace_test.go
```diff
@@ -61,9 +57,11 @@ func TestMarketplace(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to add catalogsource custom resource scheme to framework: %v", err)
 	}
+
 	// run subtests
 	t.Run("marketplace-group", func(t *testing.T) {
 		t.Run("Cluster", MarketplaceCluster)
+		t.Run("Cluster", runCSCWithInvalidTargetNamespace)
```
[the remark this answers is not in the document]
awgreene: Sounds good to me - are you okay doing this in the [refactor PR](https://github.com/operator-framework/operator-marketplace/pull/145#discussion_r269083130) we had discussed?

test/e2e/helpers.go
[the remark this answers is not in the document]
awgreene: The helper.go file was generated by the sdk - not a big issue given that were appending new methods to it but something worth highlighting.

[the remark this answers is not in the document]
awgreene: If we do want to move this to a separate package could we do it in the refactor PR mentioned earlier?

awgreene:
@aravindhp @kevinrizza @tkashem all feedback has been addressed.

test/e2e/helpers.go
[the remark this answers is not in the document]
kevinrizza: I don't believe helper.go was created by the sdk, I'm fairly sure I created that manually.

awgreene:
Install failure:
>level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"

/retest

test/e2e/helpers.go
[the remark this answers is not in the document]
awgreene: @kevinrizza My mistake, I had confused the helper functions you wrote with [similar helper functions provided by the sdk](https://github.com/operator-framework/operator-sdk/blob/master/pkg/test/e2eutil/wait_util.go).

aravindhp:
Removed my approval to give @tkashem a chance to review

tkashem:
Talked to @awgreene, the feedback I had will be addressed in a separate PR.
/approve

PR #147, opened 2019-03-27
Migrate to fake client for OperatorSource unit tests

anik120:
- Refractor OperatorSource configuring reconciler
- Introduce CatalogSourceConfigCrud for OperatorSource configuring reconciler
- Include tests for CatalogSourceConfigCrud
- Update OperatorSource tests with fake client
- Remove kubeclient mocks from generate-mocks target in Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,10 +34,11 @@ generate-mocks:
 	@echo making sure directory for mocks exists
 	mkdir -p $(MOCKS_DIR)
 
+	# $(mockgen) -destination=<Path/file where the mock is generated> -package=<The package that the generated mock files will belong to> -mock_names=<Original Interface name>=<Name of Generated mocked Interface> <Go package path of the original interface> <comma seperated list of the interface you want to mock>
 	$(mockgen) -destination=$(MOCKS_DIR)/$(OPERATORSOURCE_MOCK_PKG)/mock_datastore.go -package=$(OPERATORSOURCE_MOCK_PKG) -mock_names=Reader=DatastoreReader,Writer=DatastoreWriter $(PKG)/datastore Reader,Writer
 	$(mockgen) -destination=$(MOCKS_DIR)/$(OPERATORSOURCE_MOCK_PKG)/mock_phase_reconciler.go -package=$(OPERATORSOURCE_MOCK_PKG) -mock_names=Reconciler=PhaseReconciler $(PKG)/operatorsource Reconciler
 	$(mockgen) -destination=$(MOCKS_DIR)/$(OPERATORSOURCE_MOCK_PKG)/mock_phase_tansitioner.go -package=$(OPERATORSOURCE_MOCK_PKG) -mock_names=Transitioner=PhaseTransitioner $(PKG)/phase Transitioner
-	$(mockgen) -destination=$(MOCKS_DIR)/$(OPERATORSOURCE_MOCK_PKG)/mock_kubeclient.go -package=$(OPERATORSOURCE_MOCK_PKG) -mock_names=Client=KubeClient $(CONTROLLER_RUNTIME_PKG)/client Client
+	$(mockgen) -destination=$(MOCKS_DIR)/$(OPERATORSOURCE_MOCK_PKG)/mock_client.go -package=$(OPERATORSOURCE_MOCK_PKG) -mock_names=Client=Client $(PKG)/client Client
 	$(mockgen) -destination=$(MOCKS_DIR)/$(OPERATORSOURCE_MOCK_PKG)/mock_phase_reconciler_strategy.go -package=$(OPERATORSOURCE_MOCK_PKG) $(PKG)/operatorsource PhaseReconcilerFactory
 	$(mockgen) -destination=$(MOCKS_DIR)/$(OPERATORSOURCE_MOCK_PKG)/mock_appregistry.go -package=$(OPERATORSOURCE_MOCK_PKG) -mock_names=ClientFactory=AppRegistryClientFactory,Client=AppRegistryClient $(PKG)/appregistry ClientFactory,Client
 	$(mockgen) -destination=$(MOCKS_DIR)/$(OPERATORSOURCE_MOCK_PKG)/mock_syncer.go -package=$(OPERATORSOURCE_MOCK_PKG) -mock_names=PackageRefreshNotificationSender=SyncerPackageRefreshNotificationSender $(PKG)/operatorsource PackageRefreshNotificationSender
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -1,0 +1,49 @@
+package client
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ObjectKey identifies a Kubernetes Object.
+type ObjectKey = types.NamespacedName
+
+// Client is a wrapper around the raw kube client provided
+// by operator-sdk. Using the wrapper facilitates mocking of client
+// interactions with the cluster, while using fakeclient during unit testing.
+type Client interface {
+	Create(ctx context.Context, obj runtime.Object) error
+	Get(ctx context.Context, key ObjectKey, objExisting runtime.Object) error
+	Update(ctx context.Context, obj runtime.Object) error
+}
+
+// kubeClient is an implementation of the Client interface
+type kubeClient struct {
+	client client.Client
+}
+
+// NewClient returns a kubeClient that can perform
+// create, get and update operations on a runtime object
+func NewClient(client client.Client) Client {
+	return &kubeClient{
+		client: client,
+	}
+}
+
+// Create creates a new runtime object in the cluster
+func (h *kubeClient) Create(ctx context.Context, obj runtime.Object) error {
+	return h.client.Create(ctx, obj)
+}
+
+// Get gets an existing runtime object from the cluster
+func (h *kubeClient) Get(ctx context.Context, key ObjectKey, objExisting runtime.Object) error {
+	return h.client.Get(ctx, key, objExisting)
+}
+
+// Update updates an existing runtime object in the cluster
+func (h *kubeClient) Update(ctx context.Context, obj runtime.Object) error {
+	return h.client.Update(ctx, obj)
+}
--- a/pkg/operatorsource/configuring.go
+++ b/pkg/operatorsource/configuring.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	interface_client "github.com/operator-framework/operator-marketplace/pkg/client"
 	"github.com/operator-framework/operator-marketplace/pkg/datastore"
 	"github.com/operator-framework/operator-marketplace/pkg/phase"
 	log "github.com/sirupsen/logrus"
@@ -15,6 +16,16 @@ import (
 // NewConfiguringReconciler returns a Reconciler that reconciles
 // an OperatorSource object in "Configuring" phase.
 func NewConfiguringReconciler(logger *log.Entry, datastore datastore.Writer, client client.Client) Reconciler {
+	return NewReconcilerWithInterfaceClient(logger, datastore, interface_client.NewClient(client))
+}
+
+// NewReconcilerWithInterfaceClient returns a configuring Reconciler
+// that reconciles an OperatorSource object in "Configuring" phase.
+// It uses the Client interface which is a wrapper to the raw client
+// provided by the operator-sdk, instead of the raw client itself.
+// Using this interface facilitates mocking of kube client interaction
+// with the cluster, while using fakeclient during unit testing.
+func NewReconcilerWithInterfaceClient(logger *log.Entry, datastore datastore.Writer, client interface_client.Client) Reconciler {
 	return &configuringReconciler{
 		logger:    logger,
 		datastore: datastore,
@@ -28,7 +39,7 @@ func NewConfiguringReconciler(logger *log.Entry, datastore datastore.Writer, cli
 type configuringReconciler struct {
 	logger    *log.Entry
 	datastore datastore.Writer
-	client    client.Client
+	client    interface_client.Client
 	builder   *CatalogSourceConfigBuilder
 }
 
--- a/pkg/operatorsource/configuring_test.go
+++ b/pkg/operatorsource/configuring_test.go
@@ -13,7 +13,6 @@ import (
 	"github.com/operator-framework/operator-marketplace/pkg/operatorsource"
 	"github.com/operator-framework/operator-marketplace/pkg/phase"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/types"
 )
 
 // Use Case: Not configured, CatalogSourceConfig object has not been created yet.
@@ -29,9 +28,9 @@ func TestReconcile_NotConfigured_NewCatalogConfigSourceObjectCreated(t *testing.
 	}
 
 	datastore := mocks.NewDatastoreWriter(controller)
-	kubeclient := mocks.NewKubeClient(controller)
 
-	reconciler := operatorsource.NewConfiguringReconciler(helperGetContextLogger(), datastore, kubeclient)
+	fakeclient := NewFakeClient()
+	reconciler := operatorsource.NewConfiguringReconciler(helperGetContextLogger(), datastore, fakeclient)
 
 	ctx := context.TODO()
 	opsrcIn := helperNewOperatorSourceWithPhase("marketplace", "foo", phase.Configuring)
@@ -51,7 +50,6 @@ func TestReconcile_NotConfigured_NewCatalogConfigSourceObjectCreated(t *testing.
 		TargetNamespace: opsrcIn.Namespace,
 		Packages:        packages,
 	}
-	kubeclient.EXPECT().Create(context.TODO(), cscWant).Return(nil)
 
 	opsrcGot, nextPhaseGot, errGot := reconciler.Reconcile(ctx, opsrcIn)
 
@@ -75,9 +73,6 @@ func TestReconcile_CatalogSourceConfigAlreadyExists_Updated(t *testing.T) {
 	}
 
 	datastore := mocks.NewDatastoreWriter(controller)
-	kubeclient := mocks.NewKubeClient(controller)
-
-	reconciler := operatorsource.NewConfiguringReconciler(helperGetContextLogger(), datastore, kubeclient)
 
 	ctx := context.TODO()
 	opsrcIn := helperNewOperatorSourceWithPhase(namespace, name, phase.Configuring)
@@ -92,20 +87,10 @@ func TestReconcile_CatalogSourceConfigAlreadyExists_Updated(t *testing.T) {
 	packages := "a,b,c"
 	datastore.EXPECT().GetPackageIDsByOperatorSource(opsrcIn.GetUID()).Return(packages)
 
-	createErr := k8s_errors.NewAlreadyExists(schema.GroupResource{}, "CatalogSourceConfig already exists")
-	kubeclient.EXPECT().Create(context.TODO(), gomock.Any()).Return(createErr)
+	csc := helperNewCatalogSourceConfigWithLabels(namespace, name, labelsWant)
+	fakeclient := NewFakeClientWithCSC(csc)
 
-	// We expect Get to return the given CatalogSourceConfig successfully.
-	namespacedName := types.NamespacedName{Name: name, Namespace: namespace}
-	cscGet := marketplace.CatalogSourceConfig{}
-	kubeclient.EXPECT().Get(context.TODO(), namespacedName, &cscGet).Return(nil)
-
-	cscWant := helperNewCatalogSourceConfigWithLabels("", "", labelsWant)
-	cscWant.Spec = marketplace.CatalogSourceConfigSpec{
-		TargetNamespace: opsrcIn.Namespace,
-		Packages:        packages,
-	}
-	kubeclient.EXPECT().Update(context.TODO(), cscWant).Return(nil)
+	reconciler := operatorsource.NewConfiguringReconciler(helperGetContextLogger(), datastore, fakeclient)
 
 	opsrcGot, nextPhaseGot, errGot := reconciler.Reconcile(ctx, opsrcIn)
 
@@ -129,9 +114,9 @@ func TestReconcile_UpdateError_MovedToFailedPhase(t *testing.T) {
 	}
 
 	datastore := mocks.NewDatastoreWriter(controller)
-	kubeclient := mocks.NewKubeClient(controller)
+	kubeclient := mocks.NewClient(controller)
 
-	reconciler := operatorsource.NewConfiguringReconciler(helperGetContextLogger(), datastore, kubeclient)
+	reconciler := operatorsource.NewReconcilerWithInterfaceClient(helperGetContextLogger(), datastore, kubeclient)
 
 	ctx := context.TODO()
 	opsrcIn := helperNewOperatorSourceWithPhase(namespace, name, phase.Configuring)
--- a/pkg/operatorsource/handler_test.go
+++ b/pkg/operatorsource/handler_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/operator-framework/operator-marketplace/pkg/operatorsource"
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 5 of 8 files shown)


Makefile
```diff
@@ -34,10 +34,11 @@ generate-mocks:
 	@echo making sure directory for mocks exists
 	mkdir -p $(MOCKS_DIR)
 
+	# $(mockgen) -destination=<Path/file where the mock is generated> -package=<The package that the generated mock files will belong to> -mock_names=<Original Interface name>=<Name of Generated mocked Interface> <Go package path of the original interface> <comma seperated list of the interface you want to mock>
```
kevinrizza: Why add this comment?

pkg/operatorsource/catalogsourceconfigcrud.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"context"
+
+	v1 "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// CatalogSourceConfigCRUD defines CRUD operations for CatalogSourceConfig object in the cluster
+type CatalogSourceConfigCRUD interface {
```
kevinrizza: Can we think of a better name than this? It should be a noun of some kind that describes the purpose of this type, not just the action(s) that it does. `ConfiguringHelper` maybe? @tkashem @aravindhp 

This comment goes for the type, the struct, and the file name.

Makefile
anik120: @kevinrizza I had included this to help me understand/remember the meaning of the command. Thought it would be useful for future contributors who are new to `mock`.

pkg/operatorsource/catalogsourceconfigcrud.go
aravindhp: I agree with @kevinrizza. Moreover this shouldn't be in the `operatorsource` pkg. I also wonder if using `client-gen` to generate a client will help. Thoughts?

pkg/operatorsource/catalogsourceconfigcrud_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+// Use Case: CatalogSourceConfig object has not been created yet.
+// Expected Result: A properly populated CatalogSourceConfig should get created
+func TestCRUD_Create_NewCatalogSourceConfig(t *testing.T) {
+
+	// Set up new fakeclient
+	scheme := runtime.NewScheme()
+	apis.AddToScheme(scheme)
+
+	fakeclient := fake.NewFakeClientWithScheme(scheme)
```
aravindhp: Setting up the fake client warrants a helper function.

pkg/operatorsource/catalogsourceconfigcrud.go
```diff
@@ -0,0 +1,72 @@
...
+// NewCatalogSourceConfigCRUD returns a cscCrud that can perform
+// CRUD operations on a Catalogsourceconfig object
+func NewCatalogSourceConfigCRUD(client client.Client) CatalogSourceConfigCRUD {
+	return &cscCrud{
+		client: client,
+	}
+}
+
+// CreateCatalogSourceConfig creates a new CatalogSourceConfig
+func (h *cscCrud) CreateCatalogSourceConfig(ctx context.Context, in *v1.OperatorSource, manifests string) error {
```
aravindhp: The `in` should be `owner` and it technically can be any object. This comment applies everywhere you are doing this.

pkg/operatorsource/catalogsourceconfigcrud_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+	namespace, name := "marketplace", "foo"
+	context := context.TODO()
+	opsrc := helperNewOperatorSourceWithEndpoint(namespace, name, "remote")
+	manifests := "a,b,c"
+	catalogsourceconfigcrud := operatorsource.NewCatalogSourceConfigCRUD(fakeclient)
+
+	err := catalogsourceconfigcrud.CreateCatalogSourceConfig(context, opsrc, manifests)
+
+	assert.NoError(t, err)
```
aravindhp: Please do more assertions here. This comment applies in other spots too.

pkg/operatorsource/catalogsourceconfigcrud_test.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// Use Case: CatalogSourceConfig object already exists.
+// Expected Result: An error with message "CatalogSourceConfig object already exists"
```
aravindhp: Where are you checking for this?

Makefile
kevinrizza: It just seems unrelated to the rest of the pr. If you feel that this comment is useful, my suggestion is to create a separate pr for it.

pkg/operatorsource/catalogsourceconfigcrud.go
kevinrizza: +1

Makefile
aravindhp: Well he is changing a line in this file, so I am fine allowing it. @tkashem can be the tie breaker.

tkashem: A related comment is being added to the very thing worked on. It looks good to me.

pkg/operatorsource/catalogsourceconfigcrud.go
tkashem: I would go a step further and say that this interface can be promoted into a generalized client wrapper so that it can be reused:
```
type Client Interface {
   Update(ctx context.Context, obj runtime.Object) error
   Create(ctx context.Context, obj runtime.Object) error
   Get(ctx context.Context, key ObjectKey, obj runtime.Object) error
   Delete(ctx context.Context, obj runtime.Object, opts ...DeleteOptionFunc) error
}
```

This will be beneficial:
- No need to refactor your code because the fake client can't handle your use case.
- No need to write unit test(s) for your internal refactored code.

If you look at `CatalogSourceConfigCRUD` it's very similar to the `Client` interface above.

pkg/operatorsource/catalogsourceconfigcrud_test.go
```diff
@@ -0,0 +1,175 @@
...
+	namespace, name := "marketplace", "foo"
+	context := context.TODO()
+	opsrc := helperNewOperatorSourceWithEndpoint(namespace, name, "remote")
+	manifests := "a,b,c"
+	catalogsourceconfigcrud := operatorsource.NewCatalogSourceConfigCRUD(fakeclient)
+
+	err := catalogsourceconfigcrud.CreateCatalogSourceConfig(context, opsrc, manifests)
+
+	assert.NoError(t, err)
+}
```
tkashem: yeah, the assertion needs to `get` the object created from the fake client and assert that it is `equal` to the expected. This is more like `state` verification as opposed to `behavior` verification with a mock.

pkg/operatorsource/catalogsourceconfigcrud_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+	fakeclient := fake.NewFakeClientWithScheme(scheme, objs...)
+
+	context := context.TODO()
+	opsrc := helperNewOperatorSourceWithPhase(namespace, name, phase.Configuring)
+	catalogsourceconfigcrud := operatorsource.NewCatalogSourceConfigCRUD(fakeclient)
+
+	errGot := catalogsourceconfigcrud.UpdateCatalogSourceConfig(context, opsrc, *cscIn, manifests)
+
+	assert.NoError(t, errGot)
```
tkashem: I think we need to assert by `getting` the object and comparing against the expected.

pkg/operatorsource/catalogsourceconfigcrud.go
aravindhp: Isn't this what `client-gen` provides in some sense? Can't we just use that? Or won't it help in this case?

tkashem: `client-gen` also generates the client interfaces. `operator-sdk` doesn't do it by default though. But yeah, this is exactly what I am referring to.

aravindhp: I suggest we go with `client-gen` then.

anik120: I was a little uncomfortable with the name myself. If we have a unanimous decision to go with `ConfiguringHelper` I'm going with that. 
@aravindhp I'll need more info on `client-gen`

aravindhp: Going with `client-gen` which seems to be the unanimous decision. Please do some research on that. There is also an old commit where @tkashem added that to this repo and was removed when we bumped group/version.

anik120: @aravindhp I used `client-gen` to generate a typed clientset, and it looks like it might not help us in this case. 

For starters, the client used in the clientset is of type `k8s.io/client-go/rest.Interface` where as the client used by marketplace is of type `sigs.k8s.io/controller-runtime/pkg/client.Client`. If we were to use the typed clientset in the refactored code, we'd have to change the client that marketplace is currently using. I think @tkashem  was suggesting more of an interface that uses the current client instead of changing the client we use. 
Also, `client-gen` generates interfaces that are tied to `types`, while @tkashem's suggestion was a `generalized client wrapper`

aravindhp: In that case it is fine not to use `client-gen`

anik120:
@tkashem @aravindhp @kevinrizza addressed comments. Please review.

pkg/client/client.go
```diff
@@ -0,0 +1,53 @@
...
+
+// Client defines CRUD operations for runtime object in the cluster
+type Client interface {
+	Create(ctx context.Context, obj runtime.Object) error
+	Get(ctx context.Context, key ObjectKey, objExisting runtime.Object) error
+	Update(ctx context.Context, obj runtime.Object) error
+}
+
+// crud is an implementation of the Client interface
+type crud struct {
```
anik120: @aravindhp @tkashem @kevinrizza let me know if you guys can think of a better name for this struct.

pkg/operatorsource/configuring.go
```diff
@@ -15,21 +17,29 @@ import (
...
+func NewReconcilerWithMarketplaceClient(logger *log.Entry, datastore datastore.Writer, clientMarketplace client_marketplace.Client) Reconciler {
 	return &configuringReconciler{
-		logger:    logger,
-		datastore: datastore,
-		client:    client,
-		builder:   &CatalogSourceConfigBuilder{},
+		logger:            logger,
+		datastore:         datastore,
+		builder:           &CatalogSourceConfigBuilder{},
+		clientMarketplace: clientMarketplace,
```
anik120: Let me know if you guys can think of a better name than `clientMarketplace`

pkg/client/client.go
kevinrizza: We don't expose a `delete` method at all, so `crud` doesn't make sense. But also I would prefer to use some kind of noun that describes what the struct does, rather than a [Nominalization](https://en.wikipedia.org/wiki/Nominalization).

I think there's nothing wrong with calling such a thing `client` or `kubeClient`, given that that's the name of the interface it is implementing.

pkg/operatorsource/configuring.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 
 	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	v1 "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	client_marketplace "github.com/operator-framework/operator-marketplace/pkg/client"
```
kevinrizza: Why not `marketplace_client`? Adjectives generally go before nouns.

pkg/operatorsource/configuring.go
```diff
@@ -15,21 +17,29 @@ import (
 // NewConfiguringReconciler returns a Reconciler that reconciles
 // an OperatorSource object in "Configuring" phase.
 func NewConfiguringReconciler(logger *log.Entry, datastore datastore.Writer, client client.Client) Reconciler {
+	clientMarketplace := client_marketplace.NewClient(client)
```
kevinrizza: Same comment as above about word ordering and naming.

pkg/operatorsource/configuring.go
```diff
@@ -68,34 +77,36 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
-	if err == nil {
+	if !k8s_errors.IsAlreadyExists(err) {
 		nextPhase = phase.GetNext(phase.Succeeded)
 		r.logger.Info("CatalogSourceConfig object has been created successfully")
 
 		return
 	}
 
 	// If we are here, the given CatalogSourceConfig object already exists.
+	r.logger.Info("CatalogSourceConfig object already exists")
```
kevinrizza: If we are going to add this to the logger (which itself seems orthogonal to the purpose of the pr), shouldn't it be a `Debug` statement?

pkg/operatorsource/configuring.go
```diff
@@ -68,34 +77,36 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
 	cscNamespacedName := types.NamespacedName{Name: in.Name, Namespace: in.Namespace}
-	cscExisting := marketplace.CatalogSourceConfig{}
-	err = r.client.Get(ctx, cscNamespacedName, &cscExisting)
+	cscExisting := v1.CatalogSourceConfig{}
+	err = r.clientMarketplace.Get(ctx, cscNamespacedName, &cscExisting)
+	cscExisting.EnsureGVK()
```
kevinrizza: Why move this Ensure method to before the error handling?

pkg/operatorsource/configuring.go
kevinrizza: Also, do we really need to import `"sigs.k8s.io/controller-runtime/pkg/client"` anymore? It seems like we just use it for the method signature. Can't we just put the `"github.com/operator-framework/operator-marketplace/pkg/client".Client` interface on the method signature, remove the other client import, and then remove this alias entirely? Then rather than having a thing called `clientMarketplace` at all, you could just call it `client` or `kubeClient` or whatever.

pkg/operatorsource/configuring.go
```diff
@@ -119,5 +131,6 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
 	r.logger.Info("CatalogSourceConfig object has been updated successfully")
 
 	nextPhase = phase.GetNext(phase.Succeeded)
+
```
kevinrizza: Why add this space here? It seems like noise.

pkg/operatorsource/configuring.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 
 	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	v1 "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
```
kevinrizza: Why are we importing and aliasing  `"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"` twice?

pkg/operatorsource/configuring_test.go
```diff
@@ -23,15 +25,16 @@ func TestReconcile_NotConfigured_NewCatalogConfigSourceObjectCreated(t *testing.
 	controller := gomock.NewController(t)
 	defer controller.Finish()
 
-	nextPhaseWant := &marketplace.Phase{
+	nextPhaseWant := &v1.Phase{
```
kevinrizza: This is noise from an alias that we don't need. It seems like there's a bunch of that in this pr. Please remove it.

pkg/operatorsource/configuring_test.go
```diff
@@ -5,15 +5,17 @@ import (
...
 
 	"github.com/stretchr/testify/assert"
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	gomock "github.com/golang/mock/gomock"
+	"github.com/operator-framework/operator-marketplace/pkg/apis"
 	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	v1 "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
```
kevinrizza: We already imported this.

pkg/operatorsource/helper_test.go
```diff
@@ -88,3 +92,35 @@ func helperNewCatalogSourceConfigWithLabels(namespace, name string, opsrcLabels
 
 	return csc
 }
+
+func helperNewFakeClient() client.Client {
+
+	scheme := runtime.NewScheme()
+	apis.AddToScheme(scheme)
+	return fake.NewFakeClientWithScheme(scheme)
+
```
kevinrizza: The space before and after the code block in this function seems unnecessary.

pkg/operatorsource/configuring.go
anik120: @kevinrizza `NewConfiguringReconciler` takes a `client` of type `"sigs.k8s.io/controller-runtime/pkg/client".Client` as input, which is used to instantiate an object of the `"github.com/operator-framework/operator-marketplace/pkg/client".Client` interface. We need both imports here as the interface (`"github.com/operator-framework/operator-marketplace/pkg/client".Client`) is a wrapper for the `raw go-client` (`"sigs.k8s.io/controller-runtime/pkg/client".Client`)

pkg/operatorsource/helper_test.go
anik120: @kevinrizza the previous code blocks in this file have one space before and one after each. Won't it be inconsistent to not introduce spaces around any new code block we add?

pkg/operatorsource/configuring.go
kevinrizza: What's the point of using the interface if not to interpolate between types? If the client you wrote implements that interface, why not just update the method signature of `NewConfiguringReconciler` so that it uses the type instead of the interface? If the object that is passed to it also implements that interface it should compile.

anik120: @kevinrizza the two interfaces are not the same. I named mine `Client` for convenience sake. It might help to think of the client I wrote as `MarketplaceClient`.

"github.com/operator-framework/operator-marketplace/pkg/client".Client  (let's call this `MarketplaceClient`):
```  
type Client interface {
	Create(ctx context.Context, obj runtime.Object) error
	Get(ctx context.Context, key ObjectKey, objExisting runtime.Object) error
	Update(ctx context.Context, obj runtime.Object) error
}
````
"sigs.k8s.io/controller-runtime/pkg/client".Client (Let's call this `RawClient`):

```
type Client interface {
	Reader
	Writer
	StatusClient
}
```

`MarketplaceClient` is a wrapper around `RawClient` as @tkashem suggested. They are two different interfaces, and if a function expects an input of type `RawClient` we cannot pass it an input of type `MarketplaceClient`. Hence we import both `RawClient` and `MarketplaceClient`. We get the `RawClient` from the factory, and internally pass it to `MarketplaceClient` such that configuring reconciler sees `MarketplaceClient` and not `RawClient`.

File: pkg/operatorsource/helper_test.go (hunk 19 above)
Comment (anik120) [reply pkg/operatorsource/helper_test.go]: Also, cleaned up noise from commit.

File: pkg/operatorsource/configuring.go [hunk 20]
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	interface_client "github.com/operator-framework/operator-marketplace/pkg/client"
```
aravindhp: Here you name it `interface_client`...

pkg/operatorsource/configuring.go
```diff
@@ -15,6 +16,13 @@ import (
 // NewConfiguringReconciler returns a Reconciler that reconciles
 // an OperatorSource object in "Configuring" phase.
 func NewConfiguringReconciler(logger *log.Entry, datastore datastore.Writer, client client.Client) Reconciler {
+	return NewReconcilerWithMarketplaceClient(logger, datastore, interface_client.NewClient(client))
+}
+
+// NewReconcilerWithMarketplaceClient returns a configuring Reconciler
+// that reconciles an OperatorSource object in "Configuring" phase.
+// It takes an additional CatalogSourceConfig object as input.
+func NewReconcilerWithMarketplaceClient(logger *log.Entry, datastore datastore.Writer, client interface_client.Client) Reconciler {
```
aravindhp: ...and refer to it in the function name as `MarketplaceClient`. This is inconsistent. Moreover this is not a `MarketplaceClient` but a generic kube client. You should also call out that this only being used for unit testing. Also explain why you need this wrapper.

pkg/operatorsource/configuring.go
```diff
@@ -15,6 +16,13 @@ import (
 // NewConfiguringReconciler returns a Reconciler that reconciles
 // an OperatorSource object in "Configuring" phase.
 func NewConfiguringReconciler(logger *log.Entry, datastore datastore.Writer, client client.Client) Reconciler {
+	return NewReconcilerWithMarketplaceClient(logger, datastore, interface_client.NewClient(client))
+}
+
+// NewReconcilerWithMarketplaceClient returns a configuring Reconciler
+// that reconciles an OperatorSource object in "Configuring" phase.
+// It takes an additional CatalogSourceConfig object as input.
```
aravindhp: Where is it doing that?

pkg/operatorsource/helper_test.go
```diff
@@ -88,3 +92,35 @@ func helperNewCatalogSourceConfigWithLabels(namespace, name string, opsrcLabels
 
 	return csc
 }
+
+func helperNewFakeClient() client.Client {
```
aravindhp: Won't `NewFakeClient()` suffice here?

pkg/client/client.go
```diff
@@ -0,0 +1,53 @@
...
+// CRUD operations on a runtime object
+func NewClient(client client.Client) Client {
+	return &kubeClient{
+		client: client,
+	}
+}
+
+// Create creates a new runtime object
+func (h *kubeClient) Create(ctx context.Context, obj runtime.Object) error {
+
```
aravindhp: Please don't add unnecessary lines like this. Comment applies everywhere you are doing this.

pkg/operatorsource/handler_test.go
```diff
@@ -22,7 +23,10 @@ func TestHandle_PhaseHasChanged_UpdateExpected(t *testing.T) {
 	controller := gomock.NewController(t)
 	defer controller.Finish()
 
-	kubeclient := mocks.NewKubeClient(controller)
+	// Making two OperatorSource objects that are not equal to simulate a change.
+	opsrcIn, opsrcOut := helperNewOperatorSourceWithEndpoint("marketplace", "foo", "remote"), helperNewOperatorSourceWithEndpoint("marketplace", "foo", "local")
```
aravindhp: Why are you moving this code?

pkg/operatorsource/helper_test.go
kevinrizza: Well they are inconsistent. Some have a space above the code block (but not all), but as far as I see none of them have a space after.

pkg/operatorsource/handler_test.go
anik120: @aravindhp we need `opsrcIn` to set up the fake client with an OperatorSource. The mock client did not need the opsrcIn, hence the `opsrcIn` was initialized after the kubeclient. Since I need the `OpsrcIn` for the fakeclient, I had to move it up to initialized it first and feed it to the fakeclient.

pkg/operatorsource/helper_test.go
aravindhp: Please don't add newlines at the beginning and end of every function. It makes code looks verbose and is a bad precedent to set.

anik120: @kevinrizza I see your point. Removing the spaces.

anik120:
@kevinrizza @aravindhp updated comments and commit messages. Removed unnecessary spaces. Please let me know if the commit messages and comments are capturing the purpose of the `Client` interface sufficiently.

anik120:
@aravindhp @tkashem @kevinrizza I just realized that we might have had a better way of reviewing my [comment](https://github.com/operator-framework/operator-marketplace/pull/147#discussion_r273227460) about `clientset` use-case for our scenario. I've uploaded the generated `clientset` [here]( https://github.com/anik120/clientset/tree/master/typed/operators/v1)
I want to specifically call out the [client](https://github.com/anik120/clientset/blob/master/typed/operators/v1/operators_client.go#L36) used by the `clientset`. Please let me know if I have missed anything in my assessment.

kevinrizza:
> @aravindhp @tkashem @kevinrizza I just realized that we might have had a better way of reviewing my [comment](https://github.com/operator-framework/operator-marketplace/pull/147#discussion_r273227460) about `client-gen` use-case for our scenario. I've uploaded the generated `clientset` [here](https://github.com/anik120/clientset/tree/master/typed/operators/v1)
> I want to specifically call out the [client](https://github.com/anik120/clientset/blob/master/typed/operators/v1/operators_client.go#L36) used by the `clientset`. Please let me know if I have missed anything in my assessment.

I think the statement you made above was enough.

anik120:
@aravindhp @kevinrizza this PR should be good to go now.

PR #149, opened 2019-03-29
Bug 1693375: [csc] Fix operator-registry command

aravindhp:
In the case where a CatalogSourceConfig had packages from two different app registries, the generated command was getting mangled by the time it to got to the registry pod. This was fixed creating a list of -r options that was then passed to the registry pod.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1693375

### DIFF
--- a/pkg/catalogsourceconfig/registry.go
+++ b/pkg/catalogsourceconfig/registry.go
@@ -3,7 +3,6 @@ package catalogsourceconfig
 import (
 	"context"
 	"strconv"
-	"strings"
 	"time"
 
 	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
@@ -69,7 +68,7 @@ func NewRegistry(log *logrus.Entry, client client.Client, reader datastore.Reade
 // Ensure ensures a registry-pod deployment and its associated
 // resources are created.
 func (r *registry) Ensure() error {
-	appRegistries, secretIsPresent := r.getAppRegistryCmdLineOptions()
+	appRegistries, secretIsPresent := r.getAppRegistries()
 
 	// We create a ServiceAccount, Role and RoleBindings only if the registry
 	// pod needs to access private registry which requires access to a secret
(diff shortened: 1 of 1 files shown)


pkg/catalogsourceconfig/registry.go
```diff
@@ -243,27 +242,35 @@ func (r *registry) getLabel() map[string]string {
...
+func (r *registry) getAppRegistries() (appRegistries []string, secretIsPresent bool) {
 	for _, packageID := range r.csc.Spec.GetPackageIDs() {
 		opsrcMeta, err := r.reader.Read(packageID)
 		if err != nil {
 			r.log.Errorf("Error %v reading package %s", err, packageID)
 			continue
 		}
-		//--registry="https://quay.io/cnr|community-operators" --regisry="https://quay.io/cnr|custom-operators|mynamespace/mysecret"
-		appRegistry := "--registry=" + opsrcMeta.Endpoint + "|" + opsrcMeta.RegistryNamespace
+		// {base url with cnr prefix}|{quay registry namespace}
```
kevinrizza: I know the old comment specifically called out Quay, but maybe this comment should just call it `app registry namespace` instead? Technically this particular pattern could be used for any `app-registry` that the `operator-registry` implemented

pkg/catalogsourceconfig/registry.go
```diff
@@ -243,27 +242,35 @@ func (r *registry) getLabel() map[string]string {
...
+		appRegistry := opsrcMeta.Endpoint + "|" + opsrcMeta.RegistryNamespace
 		if opsrcMeta.SecretNamespacedName != "" {
+			// {base url with cnr prefix}|{quay registry namespace}|{secret namespace/secret name}
```
kevinrizza: same comment as above.

pkg/catalogsourceconfig/registry.go
aravindhp: I copied this from https://github.com/operator-framework/operator-registry/commit/717cba98d4254ea5983579a5370354653ecfa553#diff-1577e3efd6ba0d0bb82f2e00c849137aR37. But you do have a point. @tkashem this should be fixed there too.

aravindhp:
@kevinrizza I have addressed your comments

aravindhp:
@cuppett we were able to get the console e2e to pass locally against this PR. So I am going to re-run `e2e-aws` and `e2e-aws-operator` tests against this to be on the safe side and merge it once they pass.

PR #151, opened 2019-04-01
[e2e] csc with non-existent package

galletti94:
Adds e2e test to check that a csc with a non-existent package fails
properly and does not create child resources

### DIFF
--- a/test/helpers/helpers.go
+++ b/test/helpers/helpers.go
@@ -175,20 +175,12 @@ func CreateOperatorSourceDefinition(namespace string) *marketplace.OperatorSourc
 		},
 	}
 }
-
-// CheckCatalogSourceConfigAndChildResourcesCreated checks that a CatalogSourceConfig
-// and it's child resources were deployed.
-func CheckCatalogSourceConfigAndChildResourcesCreated(client test.FrameworkClient, cscName string, namespace string, targetNamespace string) error {
-	// Check that the CatalogSourceConfig was created.
-	resultCatalogSourceConfig := &marketplace.CatalogSourceConfig{}
-	err := WaitForResult(client, resultCatalogSourceConfig, namespace, cscName)
-	if err != nil {
-		return err
-	}
-
+// CheckCscChildResourcesCreated checks that a CatalogSourceConfig's
+// child resources were deployed.
+func CheckCscChildResourcesCreated(client test.FrameworkClient, cscName string, namespace string, targetNamespace string) error {
 	// Check that the CatalogSource was created.
 	resultCatalogSource := &olm.CatalogSource{}
-	err = WaitForResult(client, resultCatalogSource, targetNamespace, cscName)
+	err := WaitForResult(client, resultCatalogSource, targetNamespace, cscName)
 	if err != nil {
 		return err
 	}
@@ -215,19 +207,12 @@ func CheckCatalogSourceConfigAndChildResourcesCreated(client test.FrameworkClien
 	return nil
 }
 
-// CheckCatalogSourceConfigAndChildResourcesDeleted checks that a CatalogSourceConfig
-// and it's child resources were deleted.
-func CheckCatalogSourceConfigAndChildResourcesDeleted(client test.FrameworkClient, cscName string, namespace string, targetNamespace string) error {
-	// Check that the CatalogSourceConfig was deleted.
-	resultCatalogSourceConfig := &marketplace.CatalogSourceConfig{}
-	err := WaitForNotFound(client, resultCatalogSourceConfig, namespace, cscName)
-	if err != nil {
-		return err
-	}
-
+// CheckCscChildResourcesDeleted checks that a CatalogSourceConfig's
+// child resources were deleted.
+func CheckCscChildResourcesDeleted(client test.FrameworkClient, cscName string, namespace string, targetNamespace string) error {
 	// Check that the CatalogSource was deleted.
 	resultCatalogSource := &olm.CatalogSource{}
-	err = WaitForNotFound(client, resultCatalogSource, targetNamespace, cscName)
+	err := WaitForNotFound(client, resultCatalogSource, targetNamespace, cscName)
 	if err != nil {
 		return err
 	}
@@ -245,5 +230,43 @@ func CheckCatalogSourceConfigAndChildResourcesDeleted(client test.FrameworkClien
 	if err != nil {
 		return err
 	}
+	return nil
+}
+
+// CheckCscSuccessfulCreation checks that a CatalogSourceConfig
+// and it's child resources were deployed.
+func CheckCscSuccessfulCreation(client test.FrameworkClient, cscName string, namespace string, targetNamespace string) error {
+	// Check that the CatalogSourceConfig was created.
+	resultCatalogSourceConfig := &marketplace.CatalogSourceConfig{}
+	err := WaitForResult(client, resultCatalogSourceConfig, namespace, cscName)
+	if err != nil {
+		return err
+	}
+
+	// Check that all child resources were created.
+	err = CheckCscChildResourcesCreated(client, cscName, namespace, targetNamespace)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// CheckCscSuccessfulDeletion checks that a CatalogSourceConfig
+// and it's child resources were deleted.
+func CheckCscSuccessfulDeletion(client test.FrameworkClient, cscName string, namespace string, targetNamespace string) error {
+	// Check that the CatalogSourceConfig was deleted.
+	resultCatalogSourceConfig := &marketplace.CatalogSourceConfig{}
+	err := WaitForNotFound(client, resultCatalogSourceConfig, namespace, cscName)
+	if err != nil {
+		return err
+	}
+
+	// Check that all child resources were deleted.
+	err = CheckCscChildResourcesDeleted(client, cscName, namespace, targetNamespace)
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
--- a/test/testgroups/operatorsourcetests.go
+++ b/test/testgroups/operatorsourcetests.go
@@ -29,4 +29,5 @@ func OperatorSourceTestGroup(t *testing.T) {
 	// Run the test suites.
 	t.Run("opsrc-creation-test-suite", testsuites.OpSrcCreation)
 	t.Run("csc-target-namespace-test-suite", testsuites.CscTargetNamespace)
+	t.Run("csc-packages-test-suite", testsuites.TestCscWithNonExistingPackage)
 }
--- a/test/testsuites/const.go
+++ b/test/testsuites/const.go
@@ -8,4 +8,7 @@ const (
 
 	// targetNamespace is the targetNamespace value of the CatalogSourceConfig used in this test suite.
 	targetNamespace string = "target-namespace"
+
+	// nonExistingPackageName is the non-existing package
+	nonExistingPackageName string = "non-existing-package"
 )
--- a/test/testsuites/const.go
+++ b/test/testsuites/const.go
@@ -1,0 +1,11 @@
+package testsuites
+
+const (
+	// CSC constants
+
+	// cscName is the name of the CatalogSourceConfig used in this test suite.
+	cscName string = "test-operators-csc"
+
+	// targetNamespace is the targetNamespace value of the CatalogSourceConfig used in this test suite.
+	targetNamespace string = "target-namespace"
+)
--- a/test/testsuites/cscpackagetests.go
+++ b/test/testsuites/cscpackagetests.go
@@ -1,0 +1,84 @@
+package testsuites
+
+import (
+	"fmt"
+	"testing"
+
+	operator "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	"github.com/operator-framework/operator-marketplace/test/helpers"
+	"github.com/operator-framework/operator-sdk/pkg/test"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TestCscWithNonExistingPackage tests that a csc with a non-existing package
+// is handled correctly by the marketplace operator and its child resources are not
+// created.
+func TestCscWithNonExistingPackage(t *testing.T) {
+	ctx := test.NewTestCtx(t)
+	defer ctx.Cleanup()
+
+	// Get global framework variables.
+	client := test.Global.Client
+
+	// Get test namespace
+	namespace, err := ctx.GetNamespace()
+	if err != nil {
+		t.Fatalf("Could not get namespace: %v", err)
+	}
+
+	// Create a new catalogsourceconfig with a non-existing Package
+	nonExistingPackageCSC := &operator.CatalogSourceConfig{
+		TypeMeta: metav1.TypeMeta{
+			Kind: operator.CatalogSourceConfigKind,
+		}, ObjectMeta: metav1.ObjectMeta{
+			Name:      cscName,
+			Namespace: namespace,
+		},
+		Spec: operator.CatalogSourceConfigSpec{
+			TargetNamespace: namespace,
+			Packages:        nonExistingPackageName,
+		}}
+
+	err = helpers.CreateRuntimeObject(client, ctx, nonExistingPackageCSC)
+	if err != nil {
+		t.Fatalf("Unable to create test CatalogSourceConfig: %v", err)
+	}
+
+	// Check status is updated correctly then check child resources are not created
+	t.Run("configuring-state-when-package-name-does-not-exist", configuringStateWhenPackageNameDoesNotExist)
+	t.Run("child-resources-not-created", childResourcesNotCreated)
+}
+
+// configuringStateWhenTargetNamespaceDoesNotExist is a test case that creates a CatalogSourceConfig
+// with a targetNamespace that doesn't exist and ensures that the status is updated to reflect the
+// nonexistent namespace.
+func configuringStateWhenPackageNameDoesNotExist(t *testing.T) {
+	namespace, err := test.NewTestCtx(t).GetNamespace()
+	if err != nil {
+		t.Fatalf("Could not get namespace: %v", err)
+	}
+
+	// Check that the catalogsourceconfig with an non-existing package eventually reaches the
+	// configuring phase with the expected message
+	expectedPhase := "Configuring"
+	expectedMessage := fmt.Sprintf("Still resolving package(s) - %v. Please make sure these are valid packages.", nonExistingPackageName)
+	err = helpers.WaitForExpectedPhaseAndMessage(test.Global.Client, cscName, namespace, expectedPhase, expectedMessage)
+	if err != nil {
+		t.Fatalf("CatalogSourceConfig never reached expected phase/message, expected %v/%v", expectedPhase, expectedMessage)
+	}
+}
+
+// childResourcesNotCreated checks that once a CatalogSourceConfig with a wrong package name
+// is created that all expected runtime objects are not created.
+func childResourcesNotCreated(t *testing.T) {
+	// Get test namespace.
+	namespace, err := test.NewTestCtx(t).GetNamespace()
+	if err != nil {
+		t.Fatalf("Could not get namespace: %v", err)
+	}
+	// Check that the CatalogSourceConfig's child resources were not created.
+	err = helpers.CheckCscChildResourcesDeleted(test.Global.Client, cscName, namespace, namespace)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
--- a/test/testsuites/csctargetnamespacetests.go
+++ b/test/testsuites/csctargetnamespacetests.go
@@ -11,14 +11,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const (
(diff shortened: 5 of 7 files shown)


test/e2e/csc_tests.go
```diff
@@ -18,21 +21,22 @@ const (
...
 	f := test.Global
-	// Run tests
-	if err := cscWithNonExistingTargetNamespace(t, f, ctx); err != nil {
-		t.Fatal(err)
-	}
-}
 
-func cscWithNonExistingTargetNamespace(t *testing.T, f *test.Framework, ctx *test.TestCtx) error {
 	// Get test namespace
 	namespace, err := ctx.GetNamespace()
```
SamiSousa: What is the benefit of adding the namespace check here? Each test also individually checks for the namespace as well.

test/e2e/marketplace_test.go
```diff
@@ -60,7 +60,7 @@ func TestMarketplace(t *testing.T) {
 	// run subtests
 	t.Run("marketplace-group", func(t *testing.T) {
 		t.Run("Cluster", MarketplaceCluster)
-		t.Run("csc-non-existing-namespace", runCSCWithNonExistingTargetNamespace)
+		t.Run("csc-tests", runCSCTests)
```
aravindhp: Please don't collapse the tests. Each test should test a unique case and report failures on each.

test/e2e/csc_tests.go
```diff
@@ -18,21 +21,22 @@ const (
 
 	// targetNamespace is the non-existing target namespace
 	targetNamespace string = "non-existing-namespace"
+
+	// nonExistingPackageCSCName is the name of the catalogsourceconfig that points
+	// to a non-existing Package
+	nonExistingPackageCSCName string = "non-existing-package-operators"
+
+	// packageName is the non-existing package
+	packageName string = "non-existing-package"
```
aravindhp: These shouldn't be global constants as they are only being used in one test case. I now realize `targetNamespace` might be setting a bad pattern.

test/e2e/csc_tests.go
```diff
@@ -131,3 +151,83 @@ func cscWithNonExistingTargetNamespace(t *testing.T, f *test.Framework, ctx *tes
...
+
+	// Check that we created the catalogsourceconfig with a non-existing package
+	resultCatalogSourceConfig := &operator.CatalogSourceConfig{}
+	err = WaitForResult(t, f, resultCatalogSourceConfig, namespace, nonExistingPackageCSCName)
+	if err != nil {
+		return err
+	}
+
+	// Check if the catalogsourceconfig phase and message are the expected values
+	expectedPhase := "Configuring"
```
SamiSousa: Should the phase be "Failed" in this case, since the message mentions the package wasn't found? If it is expected to be "Configuring" does the object eventually resolve to "Failed" or does it get stuck?

test/e2e/csc_tests.go
```diff
@@ -131,3 +151,83 @@ func cscWithNonExistingTargetNamespace(t *testing.T, f *test.Framework, ctx *tes
...
+	expectedMessage := fmt.Sprintf("Failed to find package: %v", packageName)
+	// Check that the catalogsourceconfig with an non-existing package eventually reaches the
+	// configuring phase with the expected message
+	err = wait.Poll(retryInterval, timeout, func() (bool, error) {
+		// catalogsourceconfig should exist so no wait
```
aravindhp: I don't understand this comment

test/e2e/marketplace_test.go
```diff
@@ -28,14 +28,14 @@ func TestMarketplace(t *testing.T) {
 		TypeMeta: metav1.TypeMeta{
 			Kind: marketplace.OperatorSourceKind,
 			APIVersion: fmt.Sprintf("%s/%s",
-			marketplace.SchemeGroupVersion.Group, marketplace.SchemeGroupVersion.Version),
+				marketplace.SchemeGroupVersion.Group, marketplace.SchemeGroupVersion.Version),
```
aravindhp: Please don't mix formatting fixes when adding new test cases.

test/e2e/csc_tests.go
kevinrizza: @SamiSousa it doesn't get set to failed because it retries until it could succeed. consider the case where a new package was specified but just wasn't picked up by the operator source yet through an update. in this case, it would be eventually consistent since it retries until it can succeed. otherwise a failed state requires manual intervention.

galletti94:
@aravindhp @kevinrizza I added some changes based on your comments. Please review when you have time.

test/e2e/csc_tests.go
```diff
@@ -56,9 +44,32 @@ func cscWithNonExistingTargetNamespace(t *testing.T, f *test.Framework, ctx *tes
 	}
 	err = createRuntimeObject(f, ctx, testOperatorSource)
 	if err != nil {
-		return err
+		t.Fatal(err)
 	}
 
+	// Run tests
```
awgreene: Resources that are created with ctx will not be deleted until the defer command is ran. This means that resources created in one test may affect another those in another test. Is this the approach that we want to take?

test/e2e/csc_tests.go
```diff
@@ -131,3 +142,89 @@ func cscWithNonExistingTargetNamespace(t *testing.T, f *test.Framework, ctx *tes
...
+		return fmt.Errorf("CatalogSourceConfig never reached expected phase/message, expected %v/%v", expectedPhase, expectedMessage)
+	}
+
+	catalogSourceName := nonExistingPackageCSCName
+	serviceName := nonExistingPackageCSCName
+	deploymentName := nonExistingPackageCSCName
+
+	// Check catalog source was not created.
+	resultCatalogSource := &olm.CatalogSource{}
+	err = WaitForResult(t, f, resultCatalogSource, namespace, catalogSourceName)
```
awgreene: In this example, you seem to be using the WaitForResult to confirm that the resource doesn't exist by waiting for a timeout, which [takes 5 minutes](https://github.com/operator-framework/operator-marketplace/blob/master/test/e2e/helpers.go#L28). You seem to do this 3 times - wouldn't this add 15 minutes of checks?

test/e2e/marketplace_test.go
```diff
@@ -60,7 +60,7 @@ func TestMarketplace(t *testing.T) {
 	// run subtests
 	t.Run("marketplace-group", func(t *testing.T) {
```
awgreene: I don't see why we have this layer. With this implementation, test output will be similar to:
`TestMarketplace marketplace-group Cluster` vs `TestMarketplace marketplace-group csc-tests`. I don't see the value add of `marketplace-group`.

test/e2e/marketplace_test.go
awgreene: I am not opposed to grouping tests together, but the `runCSCTests` function should use the `t.Run(x,y)` method to call it's tests so the logs specify which test failed - the current approach will only print that the `csc-tests` failed and log the error.

test/e2e/csc_tests.go
galletti94: Good point - I guess since these resources are created independently I can just concurrently run these

awgreene: Is that the right approach? You would still be adding a 5 minute test to marketplace e2e tests. You could create a [new function that waits for a 404 on a resource](https://github.com/awgreene/operator-marketplace/commit/a16dc5591d766daccc83acb27cb6b5af24ad2709#diff-c92b68c95022bb062e31eb7aef425457R79)

test/e2e/marketplace_test.go
aravindhp: This is the point I was making earlier too but @kevinrizza seems to think that this the way to go. Thoughts?

test/e2e/csc_tests.go
aravindhp: We could also make `waitForResult` configurable in addition.

awgreene: In fact since these resources should never exist you could just confirm that they don't exist.

aravindhp: @galletti94 such re-factoring should ideally be separate PRs or at a minimum a separate commit in this PR.

test/e2e/marketplace_test.go
kevinrizza: Yes we should use t.Run inside each group (subtests). The value there is that you can set up some resources if a set of tests are all going to test against them (i.e. if your tests are `CatalogSourceConfig` specific and they all need an `OperatorSource` to run, it is convenient to create that `OperatorSource` once upfront, run that subset of tests with `t.Run()` and then when `ctx.Cleanup()` is run for that test group the `OperatorSource` is deleted.

galletti94:
@aravindhp @awgreene @kevinrizza I rebased and updated this PR

test/testgroups/operatorsourcetests.go
```diff
@@ -32,4 +32,5 @@ func OperatorSourceTestGroup(t *testing.T) {
 	// Run the test suites.
 	t.Run("opsrc-creation-test-suite", testsuites.OpSrcCreation)
 	t.Run("csc-target-namespace-test-suite", testsuites.CscTargetNamespace)
+	t.Run("csc-packages-test-suite", testsuites.CscPackages)
```
kevinrizza: Why do we need a separate test suite for this test?

test/testsuites/cscpackagetests.go
```diff
@@ -0,0 +1,123 @@
...
+
+	// Check deployment was not created.
+	resultDeployment := &apps.Deployment{}
+	err = helpers.WaitForResult(t, f, resultDeployment, namespace, deploymentName)
+	if err == nil {
+		return fmt.Errorf("CatalogSourceConfig with non-existing package created a Deployment")
+	}
+
+	return nil
+}
```
aravindhp: Please add a newline

test/testsuites/cscpackagetests.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	// Check service was not created.
+	resultService := &corev1.Service{}
+	err = helpers.WaitForResult(t, f, resultService, namespace, serviceName)
+	if err == nil {
+		return fmt.Errorf("CatalogSourceConfig with non-existing package created a Service")
+	}
+
+	// Check deployment was not created.
```
aravindhp: The three checks will result in this test taking 15 minutes to complete. That is not acceptable.

test/testgroups/operatorsourcetests.go
galletti94: target namespace has one - im just following the implied template.

test/testsuites/cscpackagetests.go
galletti94: makes sense - so just launching them in parallel is ok?

aravindhp: I vaguely remember having a discussion with @awgreene about parametrizing this.

galletti94: ok will follow up

test/testgroups/operatorsourcetests.go
awgreene: I support having a separate test suite here.

test/testsuites/cscpackagetests.go
awgreene: >I vaguely remember having a discussion with @awgreene about parametrizing this.

That conversation was from earlier in this thread: https://github.com/operator-framework/operator-marketplace/pull/151#discussion_r273581053

galletti94:
@aravindhp @kevinrizza @awgreene I updated the PR based on your collective feedback

test/testsuites/cscpackagetests.go
```diff
@@ -0,0 +1,123 @@
...
+		return fmt.Errorf("CatalogSourceConfig never reached expected phase/message, expected %v/%v", expectedPhase, expectedMessage)
+	}
+
+	catalogSourceName := nonExistingPackageCSCName
+	serviceName := nonExistingPackageCSCName
+	deploymentName := nonExistingPackageCSCName
+
+	// Check catalog source was not created.
+	resultCatalogSource := &olm.CatalogSource{}
+	err = helpers.WatiForNotFound(t, resultCatalogSource, namespace, catalogSourceName)
```
aravindhp: Did you actually try running this after you made this change? Please test locally before asking us to review.

galletti94:
@aravindhp tests passed locally

galletti94:
@aravindhp @kevinrizza @awgreene updated the PR and ran the tests locally

test/helpers/helpers.go
```diff
@@ -96,6 +96,33 @@ func WaitForSuccessfulDeployment(t *testing.T, deployment apps.Deployment) error
 	return nil
 }
 
+// WaitForNotFound polls the cluster for a particular resource name and namespace.
+// If the request fails because of the resource is found, it retries until the specified timeout.
+// If it succeeds in receiving an IsNotFound error, it returns nil.
+func WaitForNotFound(t *testing.T, result runtime.Object, namespace, name string) error {
```
awgreene: This function exists in PRs opened by both @kevinrizza and myself - the race is on to see who won't have to rebase.

test/testsuites/cscpackagetests.go
```diff
@@ -0,0 +1,125 @@
...
+			Packages:        packageName,
+		}}
+	err = helpers.CreateRuntimeObject(f, ctx, nonExistingPackageCSC)
+	if err != nil {
+		return err
+	}
+
+	// Check that we created the catalogsourceconfig with a non-existing package
+	resultCatalogSourceConfig := &operator.CatalogSourceConfig{}
+	err = helpers.WaitForResult(t, resultCatalogSourceConfig, namespace, nonExistingPackageCSCName)
```
awgreene: You run this same WaitForResult function below when checking for the phase, you can remove this.

kevinrizza:
/lgtm cancel

It actually looks like e2e-aws-operator is failing for a valid reason.

test/testsuites/const.go
```diff
@@ -0,0 +1,11 @@
...
+
+const (
+	// CSC constants
+
+	// cscName is the name of the CatalogSourceConfig used in this test suite.
+	cscName string = "test-operators-csc"
+
+	// targetNamespace is the targetNamespace value of the CatalogSourceConfig used in this test suite.
+	targetNamespace string = "target-namespace"
+)
```
awgreene: Nit: newline/

test/testsuites/cscpackagetests.go
```diff
@@ -0,0 +1,89 @@
...
+
+	operator "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	"github.com/operator-framework/operator-marketplace/test/helpers"
+	"github.com/operator-framework/operator-sdk/pkg/test"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CscPackages is a test suite that confirms that the packages field within
+// a CatalogSourceConfig is handled correctly.
+func CscPackages(t *testing.T) {
```
awgreene: This function is unnecessary.

test/helpers/helpers.go
```diff
@@ -163,20 +163,12 @@ func CreateOperatorSourceDefinition(namespace string) *marketplace.OperatorSourc
...
-	// Check that the CatalogSourceConfig was created.
-	resultCatalogSourceConfig := &marketplace.CatalogSourceConfig{}
-	err := WaitForResult(client, resultCatalogSourceConfig, namespace, cscName)
-	if err != nil {
-		return err
-	}
-
+// CheckCatalogSourceConfigChildResourcesCreated checks that a CatalogSourceConfig's
+// child resources were deployed.
+func CheckCatalogSourceConfigChildResourcesCreated(client test.FrameworkClient, cscName string, namespace string, targetNamespace string) error {
```
awgreene: If anyone has any suggestions on how to further distinguish `CheckCatalogSourceConfigChildResourcesCreated` from `CheckCatalogSourceConfigAndChildResourcesCreated` please share.

test/helpers/helpers.go
```diff
@@ -203,19 +195,12 @@ func CheckCatalogSourceConfigAndChildResourcesCreated(client test.FrameworkClien
...
-	// Check that the CatalogSourceConfig was deleted.
-	resultCatalogSourceConfig := &marketplace.CatalogSourceConfig{}
-	err := WaitForNotFound(client, resultCatalogSourceConfig, namespace, cscName)
-	if err != nil {
-		return err
-	}
-
+// CheckCatalogSourceConfigChildResourcesDeleted checks that a CatalogSourceConfig's
+// child resources were deleted.
+func CheckCatalogSourceConfigChildResourcesDeleted(client test.FrameworkClient, cscName string, namespace string, targetNamespace string) error {
```
awgreene: If anyone has any suggestions on how to further distinguish `CheckCatalogSourceConfigChildResourcesDeleted` from `CheckCatalogSourceConfigAndChildResourcesDeleted` please share.

kevinrizza: Can't we just make them two functions? One that just checks that the CatalogSoureConfig itself is deleted, and one that checks the child resources? Why does one have to be a superset of the other?

awgreene: There's value add having a single call that ensures that a `csc` and all it's child resources are created.

test/testsuites/const.go
```diff
@@ -8,4 +8,7 @@ const (
 
 	// targetNamespace is the targetNamespace value of the CatalogSourceConfig used in this test suite.
 	targetNamespace string = "target-namespace"
+
+	// packageName is the non-existing package
+	packageName string = "non-existing-package"
```
awgreene: This variable name is misleading. I would assume `packageName` is a working package name. I suggest changing it to `nonExistingPackageName` for clarity.

test/testsuites/cscpackagetests.go
```diff
@@ -0,0 +1,83 @@
...
+
+	operator "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	"github.com/operator-framework/operator-marketplace/test/helpers"
+	"github.com/operator-framework/operator-sdk/pkg/test"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// testCscWithNonExistingPackageName creates context and calls the
+// cscWithNonExistingPackageName test case.
+func TestCscWithNonExistingPackage(t *testing.T) {
```
awgreene: `testCscWithNonExistingPackageName` != `TestCscWithNonExistingPackage`

awgreene: Also - this description is out of date.

test/testsuites/cscpackagetests.go
```diff
@@ -0,0 +1,83 @@
...
+
+// childResourcesNotCreated checks that once a CatalogSourceConfig with a wrong package name
+// is created that all expected runtime objects are not created.
+func childResourcesNotCreated(t *testing.T) {
+	// Get test namespace.
+	namespace, err := test.NewTestCtx(t).GetNamespace()
+	if err != nil {
+		t.Fatalf("Could not get namespace: %v", err)
+	}
+	// Check that the CatalogSourceConfig and its child resources were created.
```
awgreene: This comment is wrong.

test/testsuites/cscpackagetests.go
```diff
@@ -0,0 +1,84 @@
...
+			Packages:        nonExistingPackageName,
+		}}
+
+	err = helpers.CreateRuntimeObject(client, ctx, nonExistingPackageCSC)
+	if err != nil {
+		t.Fatalf("Unable to create test CatalogSourceConfig: %v", err)
+	}
+
+	// Run tests.
+	t.Run("configuring-state-when-package-name-dne", configuringStateWhenPackageNameDoesNotExist)
```
aravindhp: Please expand `dne`

test/helpers/helpers.go
aravindhp: Please be consistent and either use `CSC` or `Csc` or `CatalogSourceConfig` in the function names.

test/helpers/helpers.go
```diff
@@ -235,3 +220,41 @@ func CheckCatalogSourceConfigAndChildResourcesDeleted(client test.FrameworkClien
 	}
 	return nil
 }
+
+// CheckSuccessfulCreationCsc checks that a CatalogSourceConfig
+// and it's child resources were deployed.
+func CheckSuccessfulCreationCsc(client test.FrameworkClient, cscName string, namespace string, targetNamespace string) error {
```
aravindhp: Going with the naming pattern you are following this should be `CheckCscSuccessfulCreation`

test/helpers/helpers.go
```diff
@@ -235,3 +220,41 @@ func CheckCatalogSourceConfigAndChildResourcesDeleted(client test.FrameworkClien
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// CheckSuccessfulDeletionCsc checks that a CatalogSourceConfig
+// and it's child resources were deleted.
+func CheckSuccessfulDeletionCsc(client test.FrameworkClient, cscName string, namespace string, targetNamespace string) error {
```
aravindhp: Going with the naming pattern you are following this should be `CheckCscSuccessfulDeletion`

galletti94:
@aravindhp I addressed your comments. Please review

kevinrizza:
@galletti94 I think you need to pull latest on top of this PR since it is going to fail to compile now that https://github.com/operator-framework/operator-marketplace/pull/170 has merged

galletti94:
@kevinrizza just rebased - please review and lmk if it lgty

PR #158, opened 2019-04-05
[e2e] fix wrong csc kind

galletti94:
csc kind was set to operatorsource kind in the targetNamespace e2e tests. The kind gets overwritten by the createRuntimeObject function and that's why this error wasn't caught at runtime. This PR fixes the kind to ensure that no runtime error gets produced in the future.


galletti94:
@aravindhp I replaced [bug] with [e2e]. Let me know if that lgty

galletti94:
@kevinrizza @aravindhp I rebased and updated this PR

galletti94:
can we remove the hold label?

PR #161, opened 2019-04-11
[e2e] Refactor and documentation

awgreene:
[doc] Document E2E Testing
    * Document Marketplace e2e testing.
    * Move instruction on running e2e tests from the README to the e2e 

[e2e] Refactor e2e tests   
    Update the e2e test repository to:
    * Reuse an existing operatorsource for a series of test suites.
    * Move e2e helper functions to the helpers package.
    * Created a testsuites package which contains files that group tests
       together in test suites.


docs/e2e-testing.md
```diff
@@ -0,0 +1,77 @@
+# Marketplace End to End (e2e) Testing
+
+The marketplace operator leverages the [Operator-SDK](https://github.com/operator-framework/operator-sdk/) to create and run it's e2e tests. While the [Operator-SDK's e2e Test Framework](https://github.com/operator-framework/operator-sdk/blob/master/doc/test-framework/writing-e2e-tests.md) provides a number of tools that makes writing e2e tests easier, it is up to the authors of the operator to decide how to implement the tests. This document will cover the approach that the marketplace team has taken and will provide users with enough information to extend the e2e test portfolio.
```
kevinrizza: `to create and run it's e2e tests`

should be

`to create and run its e2e tests`

test/e2e/marketplace_test.go
```diff
@@ -1,41 +1,67 @@
 package e2e
```
kevinrizza: Does it make sense that, since this file is also now a harness, that `main_test.go` is no longer a useful abstraction and we should just move the `TestMain(m *testing.M)` function into `marketplace_test.go` ?

docs/e2e-testing.md
```diff
@@ -0,0 +1,77 @@
...
+        t.Errorf("Could not create operator source: %v", err)
+    }
+
+    // Run the tests that rely on the runtime objects created earlier.
+    t.Run("test1", testsuites.Test1)
+    t.Run("test2", testsuites.Test2)
+}
+```
+
+Adhear to the following practices when implementing your test suites:
```
kevinrizza: Should be `Adhere`

docs/e2e-testing.md
```diff
@@ -0,0 +1,77 @@
...
+
+* Do not add multiple testsuites to a single file. When creating a new test suites, the file name you create should describe the focus of the tests and end with `tests.go`.
```
kevinrizza: Maybe `When creating a new test suite` or `When creating a new testsuites file`?

test/e2e/marketplace_test.go
```diff
@@ -1,41 +1,67 @@
...
 	"fmt"
 	"testing"
 
+	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/operator-framework/operator-marketplace/pkg/apis"
 	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
-
-	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+	"github.com/operator-framework/operator-marketplace/test/helpers"
+	testsuites "github.com/operator-framework/operator-marketplace/test/testsuites"
```
kevinrizza: Isn't this alias unnecessary since it is the same string as the package name?

test/e2e/marketplace_test.go
```diff
@@ -1,41 +1,67 @@
...
-	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const (
-	GroupLabel string = "opsrc-group"
-)
-
-// Test marketplace is the root function that triggers the set of e2e tests
+// Test marketplace is the root function that triggers the set of e2e tests.
```
kevinrizza: `TestMarketplace`

test/e2e/marketplace_test.go
```diff
@@ -1,41 +1,67 @@
...
 func TestMarketplace(t *testing.T) {
-	// Add marketplace types to test framework scheme
+	initTestingFramework(t)
+
+	// init-operator-source creates an opeatorsource and then runs a a series of test suites that rely on this resource.
```
kevinrizza: `operatorsource`

test/e2e/marketplace_test.go
```diff
@@ -1,41 +1,67 @@
...
+		// Get test namespace.
+		namespace, err := test.NewTestCtx(t).GetNamespace()
+		if err != nil {
+			t.Errorf("Could not get namespace: %v", err)
+		}
+
+		// Get global framework variables.
+		f := test.Global
+
+		// Create the operatorsource.
```
kevinrizza: It seems like the configuration bits could get very long if we had lots of tests. Maybe the configuration which is general to all of these tests (for example, getting the namespace would happen in every test) should happen in this function, and code which is custom setup per suite (like creating an `operatorsource` should be defined in each suite file?

awgreene:
cc @anik120 @jeremylinlin

test/testsuites/opsrcpropagationtests.go
```diff
@@ -0,0 +1,73 @@
...
+	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	helpers "github.com/operator-framework/operator-marketplace/test/helpers"
+	"github.com/operator-framework/operator-sdk/pkg/test"
+	apps "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// OPSRCPropagation is a test suit that confirms that the expected kubernetets resources are
+// created by marketplace after the creation of an operatorsource.
+func OPSRCPropagation(t *testing.T) {
```
kevinrizza: Maybe this should be capitalized like `OpSrcPropagation`? We have that in other places in the source.

test/e2e/marketplace_test.go
awgreene: I'd support this change.

test/e2e/marketplace_test.go
awgreene: Yep, I will remove the alias.

test/testsuites/opsrcpropagationtests.go
awgreene: I'm happy to make this change.

docs/e2e-testing.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Marketplace e2e Design
+
+### Entry Point
+
+Marketplace e2e test code is contained within the [marketplace_test.go](../test/e2e/marketplace_test.go) file. When the e2e tests are kicked off, the `TestMarketplace` function will be called which will initialize the test framework and run all e2e tests.
+
+### Test Suites
+
+Marketplace tests that validate similar functionality are group together in test suites. The [testsuites package](../test/testsuites) contains functions that implements these test suites. It often makes sense to reuse runtime objects between tests that have similar dependencies - this can be acomplished by following the steps outlined below:
```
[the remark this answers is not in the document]
awgreene: Yep, good catch.

test/e2e/marketplace_test.go
awgreene: I'm glad that you brought this up @kevinrizza,

I'm not opposed to creating resources within in the test suites, but this means that resources can't be shared between test suites. If everyone thinks that the clarity outweighs the dip in performance brought on by recreating resources per test suite I am happy to make the change.

awgreene:
@galletti94 @tkashem I know you guys had worked on e2e tests as well, please let me know if you have any thoughts / concerns.

test/e2e/marketplace_test.go
kevinrizza: No, I'm saying that you can still create things within the context of groups of tests. The calling of the tests and the resource generation of the tests are orthogonal, and it would be more readable if the resources created for each test group happened somewhere other than this method. In the same way, it might be nice to be able to create a hierarchy of tests more than one level deep (for example, a bunch of tests will all use the same `operatorsource`, and some sets of tests want to share the same `catalogsourceconfig` but not all). If that was a common pattern, having the top level resources created in this method doesn't make much sense.

test/testsuites/csctargetnamespacetests.go
```diff
@@ -1,65 +1,51 @@
-package e2e
+package testsuites
```
aravindhp: Is the pattern you are suggesting is to create a new file for each test case?

test/testsuites/csctargetnamespacetests.go
```diff
@@ -1,65 +1,51 @@
...
 
-const (
-	// nonExistingTargetNamespaceCSCName is the name of the catalogsourceconfig that points
-	// to a non-existing targetNamespace
-	nonExistingTargetNamespaceCSCName string = "non-existing-namespace-operators"
-
-	// targetNamespace is the non-existing target namespace
-	targetNamespace string = "non-existing-namespace"
-)
+// CSCTargetNamespace is a test suit that confirms that the targetNamespace field within
```
aravindhp: Please `Csc` instead of `CSC` everywhere as that is the pattern we have been following so far.

test/testsuites/csctargetnamespacetests.go
```diff
@@ -1,65 +1,51 @@
...
+// a catalogsourceconfigs is handled correctly.
```
aravindhp: Nit: It would be nice to use `CatalogSourceConfig` and `OperatorSource` everywhere. It just makes it easier to read.

test/testsuites/csctargetnamespacetests.go
awgreene: The pattern I'm suggesting is to create a file that has a single test suite any number of test cases. Furthermore, the test suite function should be public and all test case functions should be private.

test/testsuites/csctargetnamespacetests.go
awgreene: Understood, will do

test/testsuites/csctargetnamespacetests.go
awgreene: Will do!

test/testsuites/csctargetnamespacetests.go
aravindhp: Then it might make sense to have separate sub-packages for opsrc and csc.

kevinrizza: @aravindhp I sort of disagree with this, because I think it will pigenhole us into particular use cases. The right solution is to define these not by the resource we are testing, but by the individual workflow we want to test. Perhaps there can be "generic opsrc workflows" and "generic csc workflows" test groups, but consider this example:

```
Create an opsrc + all the datastore bits
Create a csc
Delete the opsrc
```
Does that go in the opsrc bucket or the csc bucket? It definitely involves domain knowledge of the csc because creating one is required, but ultimately the context of "what happens" is up to deleting an opsrc.

Another example:

```
Create opsrc + relevant bits
Delete opsrc
```

At first glance, it seems like this should be part of the opsrc tests group. But in reality, it kind of needs to be its own test suite; the statement we are making about these test groups is:

1. Set up your stuff for your test group
2. Run your test group
3. Tear down your set up through the sdk cleanup() predicate

That delete test needs to be in its own group, because unless we order the tests in a very particular way (and rely on the sdk not breaking with an unusual cleanup) we can't ensure that the opsrc created for those tests doesn't get deleted by this very test.

test/helpers/helpers.go
```diff
@@ -0,0 +1,130 @@
+package helpers
```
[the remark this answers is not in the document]
awgreene: I used `git mv` and after adding the changes git decided to delete and create.

docs/e2e-testing.md
awgreene: I'll go with When creating a new test suite

test/e2e/marketplace_test.go
awgreene: Did the same for helpers in the test suite files.

test/testsuites/csctargetnamespacetests.go
aravindhp: Good point @kevinrizza. I withdraw my suggestion.

awgreene:
Comments have been addressed @aravindhp @kevinrizza

awgreene:
Forgot to save a few changes before making the commit, included now.

test/helpers/helpers.go
```diff
@@ -0,0 +1,130 @@
...
+		obj,
+		&test.CleanupOptions{
+			TestContext:   ctx,
+			Timeout:       time.Second * 30,
+			RetryInterval: time.Second * 1,
+		})
+}
+
+// CreateOperatorSource returns an OperatorSource that can be used by tests that depend on an existing OperatorSource.
+func CreateOperatorSource(namespace string) *marketplace.OperatorSource {
```
aravindhp: We should rename this to `CreateOperatorSourceDefinition` or something along those lines as we are not actually creating the object here.

test/testgroups/operatorsourcetesgs.go
```diff
@@ -0,0 +1,37 @@
+package testgroups
```
aravindhp: Please fix the file name

test/testgroups/operatorsourcetesgs.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"testing"
+
+	"github.com/operator-framework/operator-marketplace/test/helpers"
+	"github.com/operator-framework/operator-marketplace/test/testsuites"
+	"github.com/operator-framework/operator-sdk/pkg/test"
+)
+
+// OperatorSourceTestGroup creates an OperatorSource and then runs a series of test suites that rely on this resource.
+func OperatorSourceTestGroup(t *testing.T) {
```
aravindhp: Please re-factor as per @kevinrizza's suggestion

test/testsuites/csctargetnamespacetests.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// Run tests.
+	if err := cscWithNonExistingTargetNamespace(t, f, ctx); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func cscWithNonExistingTargetNamespace(t *testing.T, f *test.Framework, ctx *test.TestCtx) error {
+	// Get test namespace.
+	namespace, err := test.NewTestCtx(t).GetNamespace()
```
aravindhp: Why are you testing for this in all your tests? Can't it be done in just one place?

test/testsuites/opsrcpropagationtests.go
```diff
@@ -0,0 +1,73 @@
...
+
+	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	"github.com/operator-framework/operator-marketplace/test/helpers"
+	"github.com/operator-framework/operator-sdk/pkg/test"
+	apps "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// OpSrcPropagation is a test suite that confirms that the expected kubernetets resources are
```
aravindhp: Why is this `propagation`? You are testing creating here.

test/testsuites/csctargetnamespacetests.go
awgreene: the `t.run` function doesn't allow you to pass namespaces into the call and a namespace is required when using the client. We could create a global variable in the `helpers package` and provide a method that allows you to set/get the namespace.

test/testsuites/opsrcpropagationtests.go
awgreene: Fair - I had used propagation because the test ensured that labels were propagated correctly but I think `OpSrcCreation` captures the spirit of the test better.

test/testgroups/operatorsourcetesgs.go
awgreene: Will do!

test/testsuites/csctargetnamespacetests.go
aravindhp: Might not be worth it. Please wait for feedback from others.

kevinrizza: I'd prefer to leave it as is if we can't pass it around. I don't think it's a guarantee that every test will use the same namespace.

awgreene: @kevinrizza this is a guarantee right now because we use the [--namespace](https://github.com/operator-framework/operator-marketplace/blob/master/scripts/e2e-tests.sh#L8) tag which [forces a all tests to be ran in a single namespace](https://github.com/operator-framework/operator-sdk/blob/master/doc/test-framework/writing-e2e-tests.md#namespace-flag).

kevinrizza: This is only a guarantee that the operator is running in the namespace. I suppose from the test context call that is true, but there's nothing stopping you from writing code that touches multiple namespaces.

awgreene: Yep that's true - anyways my vote is to leave as is for now since the benefit seems minimal.

test/testsuites/opsrcpropagationtests.go
awgreene: Resolved in the latest commit.

test/testgroups/operatorsourcetesgs.go
awgreene: Resolved in the latest commit.

test/testgroups/operatorsourcetesgs.go
awgreene: Resolved in the latest commit.

test/helpers/helpers.go
awgreene: Resolved in the latest commit.

test/e2e/main_test.go
```diff
@@ -1,11 +1,66 @@
...
+
+// Test marketplace is the root function that triggers the set of e2e tests.
+func TestMarketplace(t *testing.T) {
+	initTestingFramework(t)
+
+	// Run Test Groups
+	testgroups.OperatorSourceTestGroup(t)
+}
+
+// initTestingFramework adds the marketplace operatorsource and CatalogSourceConfig types as well as the
```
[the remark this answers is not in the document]
awgreene: Resolved in the latest commit.

awgreene:
Requested changes have been made @aravindhp @kevinrizza

test/testgroups/operatorsourcetests.go
```diff
@@ -0,0 +1,35 @@
...
+	namespace, err := ctx.GetNamespace()
+	if err != nil {
+		t.Errorf("Could not get namespace: %v", err)
+	}
+
+	// Get global framework variables.
+	f := test.Global
+
+	// Create the OperatorSource.
+	err = helpers.CreateRuntimeObject(f, ctx, helpers.CreateOperatorSourceDefinition(namespace))
```
kevinrizza: So with the change you made, why don't we do this at the beginning of each test suite that needs it?

I'm still thinking about the test that creates an operator source and then deletes it. Creating the operator source here will make that test impossible -- it would be nicer if we could create that test in its own test suite (which means that each test suite that currently exists would need to create an operator source at the start)

awgreene: I don't see how the test you described is limited with the existing approach - the test that creates/deletes an operator source could be done in it's own test suite that is not a part of this group.

awgreene: The original intent with test groups was to prepare resources that could be used across test suites.

kevinrizza: Oh, this is my bad. I thought this was the harness. Reading comprehension :)

test/e2e/main_test.go
```diff
@@ -1,11 +1,66 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestMain(m *testing.M) {
-	f.MainEntry(m)
+	test.MainEntry(m)
+}
+
+// Test marketplace is the root function that triggers the set of e2e tests.
```
aravindhp: Nit: `TestMarketplace`

test/testsuites/csctargetnamespacetests.go
```diff
@@ -0,0 +1,119 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+// CscTargetNamespace is a test suit that confirms that the targetNamespace field within
+// a CatalogSourceConfig is handled correctly.
+func CscTargetNamespace(t *testing.T) {
+	t.Run("non-existing-target-namespace", testCscWithNonExistingTargetNamespace)
+}
+
+func testCscWithNonExistingTargetNamespace(t *testing.T) {
```
aravindhp: Please documentation stating what the test does. Comment applies to all test cases.

test/testsuites/opsrcpropagationtests.go
```diff
@@ -0,0 +1,73 @@
+package testsuites
```
aravindhp: Please rename the filename too.

README.md
```diff
@@ -170,22 +170,3 @@ $ oc create -f your-operator-source.yaml
 Once created, the Marketplace operator will use the `OperatorSource` to download your operator artifact from the app registry and display your operator offering in the Marketplace UI.
 
 You can also access private AppRegistry repositories via an authenticated `OperatorSource`, which you can learn more about [here](docs/how-to-authenticate-private-repositories.md).
-
-## Running End to End (e2e) Tests
```
aravindhp: Please add a link to the e2e doc here instead of removing the full section.

awgreene: Addressed in the updated commit.

test/testsuites/opsrcpropagationtests.go
awgreene: Addressed in the updated commit.

test/testsuites/csctargetnamespacetests.go
awgreene: Addressed in the updated commit.

test/e2e/main_test.go
awgreene: Addressed in the updated commit.

awgreene:
Addressed your comments @aravindh

PR #165, opened 2019-04-17
Bug 1700416:[status] Fix Status Reporting

awgreene:
Problem:
When the marketplace-operator starts up, it starts a cluster status monitoring
routine. In some cases, the marketplace-operator will incorrectly mark the
status as failing when a large enough number of operands fail to reconcile
successfully (for example, OperatorSources that are pointing to an invalid
app registry path).

Solution:
Update marketplace-operator to report availability based on whether or not it
is able to reconcile CatalogSourceConfigs and OperatorSources. Report failing
if OperatorSources or CatalogSourceConfigs fall below golden ratio, but keep
available condition set to true.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1700416

### DIFF
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -28,7 +28,7 @@ const (
 	minSyncsBeforeReporting = 4
 
 	// successRatio is the ratio of successful syncs / total syncs that we
-	// want to see in order to report that the marketplace operator is healthy.
+	// want to see in order to report that the marketplace operator is not degraded.
 	// This value is low right now because the failed syncs come from invalid CRs.
 	// As the status reporting evolves we can tweek this ratio to be a better
 	// representation of the operator's health.
@@ -152,21 +152,6 @@ func new(cfg *rest.Config, mgr manager.Manager, namespace string, version string
 	}
 }
 
-// setDegrading reports that operator has degraded along with the error message
-func (s *status) setDegrading(message string) error {
-	return s.setStatus(configv1.OperatorDegraded, message)
-}
-
-// setAvailable reports that the operator is available to process events
-func (s *status) setAvailable(message string) error {
-	return s.setStatus(configv1.OperatorAvailable, message)
-}
-
-// setProgressing reports that the operator is being deployed
-func (s *status) setProgressing(message string) error {
-	return s.setStatus(configv1.OperatorProgressing, message)
-}
-
 // ensureClusterOperator ensures that a ClusterOperator CR is present on the
 // cluster
 func (s *status) ensureClusterOperator() error {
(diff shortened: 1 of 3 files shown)


awgreene:
cc @kevinrizza

pkg/status/status.go
```diff
@@ -334,12 +282,33 @@ func (s *status) monitorClusterStatus() {
...
 	}()
+	startupFinished := false
 	for {
 		select {
 		case <-s.stopCh:
 			// If the stopCh is closed, the operator will exit and CO should
 			// be set to failing.
-			s.setFailing("Operator exited")
+			time := v1.Now()
+			statusConditions := []configv1.ClusterOperatorStatusCondition{{
```
aravindhp: This seems like an overly verbose way of having to set the conditions. I think this can be modularized and will help easily indicate what you are setting here.

pkg/status/status.go
```diff
@@ -197,17 +182,20 @@ func (s *status) ensureClusterOperator() error {
 
 // setStatus handles setting all the required fields for the given
 // ClusterStatusConditionType
-func (s *status) setStatus(condition configv1.ClusterStatusConditionType, message string) error {
+func (s *status) setStatus(statusConditions []configv1.ClusterOperatorStatusCondition) error {
 	if s.coAPINotPresent {
-		return nil
+		return fmt.Errorf("ClusterOperator API not present")
```
aravindhp: Why is this an error? Won't it break vanilla kube?

awgreene: `setStatus` is only called in the `monitorClusterStatus` routine which [isn't started on vanilla](https://github.com/operator-framework/operator-marketplace/blob/4fec4b65c1693eedc49f0f37fe13121fee59dfb8/pkg/status/status.go#L93). This would handle the usecase where the `CO API` was present on startup but was removed.

pkg/status/status.go
awgreene: I agree that creating 3 `ClusterOperatorStatusCondition` structures in an array is a bit verbose (@tkashem had brought this up as well). I had considered creating a builder function but that seemed like overkill. What are your thoughts on creating a function that returns  a `ClusterOperatorStatusCondition` object when provided `Type/Status/Message/LastTransactionTime` as arguments? This code would then be updated to create an array that call that function 3 times.

aravindhp: Why is a builder an overkill? It might be a good fit here but it depends on how it ends up looking. You other suggestion seems like it would be doing what a builder does. Others might have other suggestions so please wait until you hear back from them too.

awgreene: @tkashem @kevinrizza thoughts?

kevinrizza: I think a builder is fine. A builder with methods like `SetFailing(bool)` or `SetAvailable(bool)` seems like a good pattern here. The useful change here is that you have separated setting those fields from actually passing them to the client.

tkashem: Yeah, a builder pattern works here. Another option is to use anonymous functions that can `closure` the local variables. The anonymous functions can be defined outside the `for {` loop.

awgreene:
Ready for feedback @kevinrizza @tkashem @galletti94 @anik120

pkg/status/status.go
```diff
@@ -28,7 +28,7 @@ const (
 	minSyncsBeforeReporting = 4
 
 	// successRatio is the ratio of successful syncs / total syncs that we
-	// want to see in order to report that the marketplace operator is healthy.
+	// want to see in order to report that the marketplace operator is not degrading.
```
kevinrizza: nit: `is not degraded`

awgreene: Good catch!

anik120:
@awgreene Looks good to me!

awgreene:
Could you please review @tkashem ?

PR #166, opened 2019-04-18
[e2e] Add tests for invalid OperatorSources

SamiSousa:
- Creates NoSetupTestGroup for tests that don't require any
  resources upfront
- Testsuite for invalid OperatorSource tests
- Check for OperatorSource with invalid URL
- Check for OperatorSource with invalid endpoint
- Check for OperatorSource with non-existent regsitry namespace

These OperatorSource tests each create an invalid OperatorSource
and check that the status reaches the appropriate failed state.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -24,6 +24,7 @@ func TestMarketplace(t *testing.T) {
 
 	// Run Test Groups
 	t.Run("operator-source-test-group", testgroups.OperatorSourceTestGroup)
+	t.Run("no-setup-test-group", testgroups.NoSetupTestGroup)
 }
 
 // initTestingFramework adds the marketplace OperatorSource and CatalogSourceConfig types as well as the
--- a/test/testgroups/nosetuptests.go
+++ b/test/testgroups/nosetuptests.go
@@ -1,0 +1,13 @@
+package testgroups
+
+import (
+	"testing"
+
+	"github.com/operator-framework/operator-marketplace/test/testsuites"
+)
+
+// NoSetupTestGroup runs test suites that do not require any resources upfront
+func NoSetupTestGroup(t *testing.T) {
+	// Run the test suites.
+	t.Run("invalid-operator-source-test-suite", testsuites.InvalidOpSrc)
+}
--- a/test/testsuites/invalidopsrctests.go
+++ b/test/testsuites/invalidopsrctests.go
@@ -1,0 +1,202 @@
+package testsuites
+
+import (
+	"testing"
+	"strings"
+
+	operator "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	"github.com/operator-framework/operator-marketplace/test/helpers"
+	"github.com/operator-framework/operator-sdk/pkg/test"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	// endpointType represents the endpoint we currently support
+	endpointType string = "appregistry"
+	// marketplaceRegistryNamespace is the e2e registry namespace in quay
+	marketplaceRegistryNamespace string = "marketplace_e2e"
+)
+
+// InvalidOpSrc tests OperatorSources created with invalid values
+// to make sure the expected failure state is reached
+func InvalidOpSrc(t *testing.T) {
+	t.Run("invalid-endpoint", testOpSrcWithInvalidEndpoint)
+	t.Run("invalid-url", testOpSrcWithInvalidURL)
+	t.Run("nonexistent-registry-namespace", testOpSrcWithNonexistentRegistryNamespace)
+}
+
+// Create OperatorSource with invalid endpoint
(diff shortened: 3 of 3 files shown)


SamiSousa:
@aravindhp @kevinrizza @awgreene

test/testsuites/opsrcnonexistentregistrynamespacetest.go
```diff
@@ -0,0 +1,88 @@
...
+			return true, nil
+		}
+		return false, nil
+	})
+	if err != nil {
+		return fmt.Errorf("OperatorSource never reached expected phase/message, expected %v", expectedPhase)
+	}
+
+	return nil
+}
```
aravindhp: Please add new line.

test/testgroups/noresourcetests.go
```diff
@@ -0,0 +1,15 @@
+package testgroups
+
+import (
+	"testing"
+
+	"github.com/operator-framework/operator-marketplace/test/testsuites"
+)
+
+// NoResourceTestGroup runs test suites that do not require any setup
```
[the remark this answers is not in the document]
SamiSousa: Great point, I'll rename and explain that this test group is for testing invalid operatorsources

test/testsuites/opsrcinvalidendpointtest.go
```diff
@@ -0,0 +1,82 @@
...
+	"strings"
+
+	operator "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	"github.com/operator-framework/operator-marketplace/test/helpers"
+	"github.com/operator-framework/operator-sdk/pkg/test"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+func OpSrcWithInvalidEndpoint(t *testing.T) {
```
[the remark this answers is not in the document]
SamiSousa: Addressed

SamiSousa:
@aravindhp Addressed your comments

SamiSousa:
@tkashem @kevinrizza Could you take a look? Thank you!

test/testsuites/opsrcinvalidendpointtest.go
```diff
@@ -0,0 +1,82 @@
+package testsuites
```
awgreene: Should each of these be in their own test suite? All tests focus on invalid `operator sources` which I would argue should be included together in a `InvalidOperatorSourceTestSuite`.

test/testgroups/invalidopsrctests.go
```diff
@@ -0,0 +1,15 @@
+package testgroups
+
+import (
+	"testing"
+
+	"github.com/operator-framework/operator-marketplace/test/testsuites"
+)
+
+// InvalidOpSrcTestGroup runs test suites for invalid operatorsources
+func InvalidOpSrcTestGroup(t *testing.T) {
```
awgreene: `Test Groups` are typically used to create any resources that testsuites / tests in the group depend on. In this case, you've created a `test group` called `InvalidOpSrcTestGroup` - I think it makes more sense to create a `test group` meant for test suites that require no set up.

SamiSousa: Following the conversation on Kevin's PR, I'll update my tests to follow this format.

test/testsuites/opsrcinvalidendpointtest.go
SamiSousa: Sounds good to me. I will put these tests into a single test-suite

awgreene: These could even go in the `opsrc-creation-test-suite`.

SamiSousa: I considered that, but I didn't need the operatorsource that is created as part of the TestGroup. Although, it doesn't seem as though the `opsrccreationtests.go` uses it either

awgreene: >I considered that, but I didn't need the operatorsource that is created as part of the TestGroup. 
That's a good point 

>Although, it doesn't seem as though the opsrccreationtests.go uses it either

The test suite runs a series of tests that depend on the `OpSrc` being created before executing.

SamiSousa: Ah, that's right. I think it still makes more sense to have the invalid tests be a separate testsuite since they don't use the same operatorsource and because they test a different use case.

SamiSousa:
@awgreene @kevinrizza @aravindhp 
Updated to put each test inside one testsuite for invalid OperatorSources, and renamed the TestGroup to `NoSetupTestGroup` based on the conversation here: https://github.com/operator-framework/operator-marketplace/pull/170#discussion_r277372037

SamiSousa:
Tests are failing because I needed to rebase on master, since many helper functions changed which arguments they take. Didn't realize this would affect my PR. Just pushed the rebased changes.

SamiSousa:
Thank you @aravindhp ! I addressed your comments, please take a look!

PR #167, opened 2019-04-18
[e2e] Add CSC test where namespace != targetNamespace

awgreene:
Extended the targetNamespace testsuite to check that a CatalogSourceConfig
with different namespace and targetNamespace values is properly deployed
and cleaned up.

Other notes:
* Updates CSC packages to pull package from marketplace_e2e operator source.
* Add new functions to helpers package

### DIFF
--- a/test/helpers/helpers.go
+++ b/test/helpers/helpers.go
@@ -2,12 +2,13 @@ package helpers
 
 import (
 	"context"
-	"testing"
 	"time"
 
+	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
 	"github.com/operator-framework/operator-sdk/pkg/test"
 	apps "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -40,36 +41,24 @@ const (
 // WaitForResult polls the cluster for a particular resource name and namespace.
 // If the request fails because of an IsNotFound error it retries until the specified timeout.
 // If it succeeds it sets the result runtime.Object to the requested object.
-func WaitForResult(t *testing.T, result runtime.Object, namespace, name string) error {
-	// Get global framework variables
-	f := test.Global
-
+func WaitForResult(client test.FrameworkClient, result runtime.Object, namespace, name string) error {
 	namespacedName := types.NamespacedName{Name: name, Namespace: namespace}
-	err := wait.Poll(RetryInterval, Timeout, func() (done bool, err error) {
-		err = f.Client.Get(context.TODO(), namespacedName, result)
+	return wait.PollImmediate(RetryInterval, Timeout, func() (done bool, err error) {
+		err = client.Get(context.TODO(), namespacedName, result)
 		if err != nil {
 			if errors.IsNotFound(err) {
-				t.Logf("Waiting for creation of %s runtime object\n", name)
 				return false, nil
 			}
 			return false, err
 		}
 		return true, nil
 	})
-	if err != nil {
-		return err
-	}
-	t.Logf("Runtime object %s has been created\n", name)
-	return nil
 }
 
 // WaitForSuccessfulDeployment checks if a given deployment has readied all of
 // its replicas. If it has not, it retries until the deployment is ready or it
 // reaches the timeout.
-func WaitForSuccessfulDeployment(t *testing.T, deployment apps.Deployment) error {
-	// Get global framework variables
-	f := test.Global
-
+func WaitForSuccessfulDeployment(client test.FrameworkClient, deployment apps.Deployment) error {
 	// If deployment is already ready, lets just return.
 	if deployment.Status.ReadyReplicas == *deployment.Spec.Replicas {
 		return nil
@@ -77,28 +66,44 @@ func WaitForSuccessfulDeployment(t *testing.T, deployment apps.Deployment) error
 
 	namespacedName := types.NamespacedName{Name: deployment.Name, Namespace: deployment.Namespace}
 	result := &apps.Deployment{}
-	err := wait.Poll(RetryInterval, Timeout, func() (done bool, err error) {
-		err = f.Client.Get(context.TODO(), namespacedName, result)
+	return wait.PollImmediate(RetryInterval, Timeout, func() (done bool, err error) {
+		err = client.Get(context.TODO(), namespacedName, result)
 		if err != nil {
 			return false, err
 		}
 		if *deployment.Spec.Replicas == result.Status.ReadyReplicas {
 			return true, nil
 		}
-		t.Logf("Waiting for deployment %s to have (%d/%d) replicas ready\n", deployment.Name, result.Status.ReadyReplicas,
-			*deployment.Spec.Replicas)
 		return false, nil
 	})
-	if err != nil {
-		return err
-	}
-	t.Logf("Deployment %s has been initialized successfully\n", deployment.Name)
-	return nil
+}
+
+// WaitForExpectedPhaseAndMessage checks if a CatalogSourceConfig with the given name exists in the namespace
+// and makes sure that the phase and message matches the expected values.
+// If expectedMessage is an empty string, only the expectedPhase is checked.
+func WaitForExpectedPhaseAndMessage(client test.FrameworkClient, cscName string, namespace string, expectedPhase, expectedMessage string) error {
+	// Check that the CatalogSourceConfig exists.
+	resultCatalogSourceConfig := &marketplace.CatalogSourceConfig{}
+	return wait.PollImmediate(RetryInterval, Timeout, func() (bool, error) {
+		err := WaitForResult(client, resultCatalogSourceConfig, namespace, cscName)
+		if err != nil {
+			return false, err
+		}
+		// Check for the expected phase
+		if resultCatalogSourceConfig.Status.CurrentPhase.Name == expectedPhase {
+			// If the expected message is not empty make sure that it matches the actual message
+			if expectedMessage == "" || resultCatalogSourceConfig.Status.CurrentPhase.Message == expectedMessage {
+				return true, nil
+			}
+			return false, nil
+		}
+		return false, nil
+	})
 }
 
 // CreateRuntimeObject creates a runtime object using the test framework.
-func CreateRuntimeObject(f *test.Framework, ctx *test.TestCtx, obj runtime.Object) error {
-	return f.Client.Create(
+func CreateRuntimeObject(client test.FrameworkClient, ctx *test.TestCtx, obj runtime.Object) error {
+	return client.Create(
 		context.TODO(),
 		obj,
 		&test.CleanupOptions{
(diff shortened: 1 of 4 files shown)


test/testsuites/csctargetnamespacetests.go
```diff
@@ -4,75 +4,100 @@ import (
...
+
+	// targetNamespace is the targetNamespace value of the CatalogSourceConfig used in this test suite.
+	targetNamespace string = "target-namespace"
+)
 
-// testCscWithNonExistingTargetNamespace creates context and calls the
-// cscWithNonExistingTargetNamespace test case.
-func testCscWithNonExistingTargetNamespace(t *testing.T) {
+// CscTargetNamespace is a test suit that confirms that targetNamespace values are handled appropriately.
+func CscTargetNamespace(t *testing.T) {
```
aravindhp: It looks like you are re-factoring and adding new tests cases in the same commit. Is it possible to separate them out.

awgreene:
> It looks like you are re-factoring and adding new tests cases in the same commit. Is it possible to separate them out?

Will do @aravindhp

test/testsuites/csctargetnamespacetests.go
awgreene: Absolutely! Will address in an updated commit

awgreene: Addressed in latest update.

awgreene:
Please review @aravindhp @galletti94 @kevinrizza

test/testsuites/csctargetnamespacetests.go
```diff
@@ -12,77 +12,90 @@ import (
...
+		t.Fatalf("Unable to create test namespace: %v", err)
+	}
+
+	t.Run("succeeded-state-after-target-namespace-created", succeededStateAfterTargetNamespaceCreated)
+}
+
+// configuringStateWhenTargetNamespaceDne is a test case that creates a CatalogSourceConfig
+// with a targetNamespace that doesn't exist and ensures that the status is updated to reflect the
+// nonexistant namespace.
+func configuringStateWhenTargetNamespaceDne(t *testing.T) {
```
aravindhp: Nit: I am not a fan of the `Dne` abbreviation given that is not well known. I prefer the expansion.

test/testsuites/csctargetnamespacetests.go
```diff
@@ -138,3 +150,74 @@ func succeededStateAfterTargetNamespaceCreated(t *testing.T) {
...
+// childResourcesCreated checks that once a CatalogSourceConfig is created that all expected runtime
+// objects are created as well.
+func childResourcesCreated(t *testing.T) {
+	// Get test namespace.
+	namespace, err := test.NewTestCtx(t).GetNamespace()
+	if err != nil {
+		t.Fatalf("Could not get namespace: %v", err)
+	}
+
+	// Check for the CatalogSource.
```
aravindhp: These checks are being done during `OperatorSource` creation tests to. Please modularize.

kevinrizza: +1

test/testsuites/csctargetnamespacetests.go
awgreene: Sure thing, will change it to `configuringStateWhenTargetNamespaceDoesNotExist`

awgreene:
@galletti94 @kevinrizza @SamiSousa Please review so we can work on getting e2e tests through.

@aravindhp I have made the changes that you requested.

test/testsuites/csctargetnamespacetests.go
```diff
@@ -4,122 +4,152 @@ import (
...
-
-	// targetNamespace is the non-existing target namespace.
-	targetNamespace := "non-existing-namespace"
-
 	// Create a new CatalogSourceConfig with a non-existing targetNamespace.
-	nonExistingTargetNamespaceCsc := &operator.CatalogSourceConfig{
+	nonExistingTargetNamespaceCsc := &marketplace.CatalogSourceConfig{
 		TypeMeta: metav1.TypeMeta{
-			Kind: operator.CatalogSourceConfigKind,
+			Kind: marketplace.OperatorSourceKind,
```
galletti94: why is the kind OperatorSource here?

test/testsuites/csctargetnamespacetests.go
```diff
@@ -4,122 +4,152 @@ import (
...
 	if err != nil {
-		return err
+		t.Fatalf("Unable to create test CatalogSourceConfig: %v", err)
 	}
 
 	// Check that we created the CatalogSourceConfig with a non-existing targetNamespace.
-	resultCatalogSourceConfig := &operator.CatalogSourceConfig{}
-	err = helpers.WaitForResult(t, resultCatalogSourceConfig, namespace, nonExistingTargetNamespaceCscName)
+	resultCatalogSourceConfig := &marketplace.CatalogSourceConfig{}
+	err = helpers.WaitForResult(t, resultCatalogSourceConfig, namespace, cscName)
```
galletti94: Aren't you doing this check in the `helpers.WaitForExpectedPhaseAndMessage` which is called below in the `t.Run` functions?

test/testsuites/csctargetnamespacetests.go
```diff
@@ -4,122 +4,152 @@ import (
...
+		t.Fatal(err)
+	}
+
+	t.Run("child-resources-deleted", childResourcesDeleted)
+}
+
+// configuringStateWhenTargetNamespaceDoesNotExist is a test case that creates a CatalogSourceConfig
+// with a targetNamespace that doesn't exist and ensures that the status is updated to reflect the
+// nonexistant namespace.
+func configuringStateWhenTargetNamespaceDoesNotExist(t *testing.T) {
```
galletti94: what do you think of making namespace a parameter to each of these functions?

awgreene: The testing framework's `Run` functions only takes functions that have a single `testing.T` argument.

test/testsuites/csctargetnamespacetests.go
awgreene: Great catch, thanks!

test/testsuites/csctargetnamespacetests.go
awgreene: My reasoning is that the `t.Run` functions below are supposed to be ran with the expectation that the `CSC` exists. If the CSC isn't created the user should be looking in this method. If we remove this check, the output from the e2e tests will tell people to look in the other test cases instead.

galletti94: That's a great point

kevinrizza: Can we document this thought process in comments?

awgreene: Will do @kevinrizza

awgreene:
@galletti94 addressed your commits and added the comment as @kevinrizza suggested.

test/testsuites/csctargetnamespacetests.go
```diff
@@ -4,122 +4,154 @@ import (
...
 		},
-		Spec: operator.CatalogSourceConfigSpec{
+		Spec: marketplace.CatalogSourceConfigSpec{
 			TargetNamespace: targetNamespace,
-			Packages:        "descheduler",
+			Packages:        "camel-k-marketplace-e2e-tests",
 		}}
-	err = helpers.CreateRuntimeObject(f, ctx, nonExistingTargetNamespaceCsc)
+
+	// Create the CatalogSourceConfig and if an error occures do not run tests that
```
SamiSousa: nit: `occurs`

test/testsuites/csctargetnamespacetests.go
```diff
@@ -4,122 +4,154 @@ import (
...
+	// rely on the existance of the CatalogSourceConfig.
```
SamiSousa: nit: `existence`

test/testsuites/csctargetnamespacetests.go
```diff
@@ -4,122 +4,154 @@ import (
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Run("child-resources-deleted", childResourcesDeleted)
+}
+
+// configuringStateWhenTargetNamespaceDoesNotExist is a test case that creates a CatalogSourceConfig
+// with a targetNamespace that doesn't exist and ensures that the status is updated to reflect the
+// nonexistant namespace.
```
SamiSousa: nit: `nonexistent`

SamiSousa:
Found a few spelling errors (which I probably made in my PR too...)

awgreene:
Addressed @SamiSousa

SamiSousa:
@tkashem Could you PTAL?

test/testsuites/csctargetnamespacetests.go
```diff
@@ -4,122 +4,108 @@ import (
...
 	if err != nil {
-		return err
+		t.Fatalf("Unable to create test CatalogSourceConfig: %v", err)
 	}
 
 	// Check that we created the CatalogSourceConfig with a non-existing targetNamespace.
-	resultCatalogSourceConfig := &operator.CatalogSourceConfig{}
-	err = helpers.WaitForResult(t, resultCatalogSourceConfig, namespace, nonExistingTargetNamespaceCscName)
+	resultCatalogSourceConfig := &marketplace.CatalogSourceConfig{}
+	err = helpers.WaitForResult(t, resultCatalogSourceConfig, namespace, cscName)
```
tkashem: If the response to `Create API` inside `helpers.CreateRuntimeObject` is a success then it's a guarantee that the given object has been created/stored successfully. This makes `helpers.WaitForResult` redundant here unless you want to ensure that the object is in certain state.

test/helpers/helpers.go
```diff
@@ -202,3 +236,43 @@ func CheckCatalogSourceConfigAndChildResourcesCreated(t *testing.T, cscName stri
...
+// CheckCatalogSourceConfigAndChildResourcesDeleted checks that a CatalogSourceConfig
+// and it's child resources were deleted.
+func CheckCatalogSourceConfigAndChildResourcesDeleted(t *testing.T, cscName string, namespace string, targetNamespace string) error {
+	// Get test namespace.
+	namespace, err := test.NewTestCtx(t).GetNamespace()
+	if err != nil {
+		return fmt.Errorf("Could not get namespace: %v", err)
+	}
+
+	// Check that the CatalogSourceConfig was created.
```
tkashem: typo: `deleted`

test/helpers/helpers.go
```diff
@@ -124,6 +124,33 @@ func WaitForExpectedPhaseAndMessage(t *testing.T, cscName string, namespace stri
 	return err
 }
 
+// WaitForNotFound polls the cluster for a particular resource name and namespace
+// If the request fails because the runtime object is found it retries until the specified timeout
+// If the request returns a IsNotFound error the method will return true
+func WaitForNotFound(t *testing.T, result runtime.Object, namespace, name string) error {
```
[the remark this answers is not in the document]
awgreene: I went with `WaitForNotFound` because we only return true when the `errors.IsNotFound` call passes.

>I think this function is being introduced by a separate PR as well. I assume we are going to do some refactor to keep one?

Yep! We're pushing this PR before rebasing the others.

test/helpers/helpers.go
```diff
@@ -136,6 +163,13 @@ func CreateRuntimeObject(f *test.Framework, ctx *test.TestCtx, obj runtime.Objec
 		})
 }
 
+// DeleteRuntimeObject deletes a runtime object using the test framework
+func DeleteRuntimeObject(f *test.Framework, obj runtime.Object) error {
```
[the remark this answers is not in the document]
awgreene: Yep! We're pushing this PR before rebasing the others.

test/testsuites/csctargetnamespacetests.go
```diff
@@ -106,6 +118,40 @@ func succeededStateAfterTargetNamespaceCreated(t *testing.T) {
...
+// objects are deleted as well.
+func childResourcesDeleted(t *testing.T) {
+	// Get test namespace
+	namespace, err := test.NewTestCtx(t).GetNamespace()
+	if err != nil {
+		t.Fatalf("Could not get namespace: %v", err)
+	}
+
+	// Check that the CatalogSourceConfig and its child resources were deleted.
+	err = helpers.CheckCatalogSourceConfigAndChildResourcesDeleted(t, cscName, namespace, targetNamespace)
```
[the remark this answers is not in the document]
awgreene: My mistake, good catch!

test/testsuites/csctargetnamespacetests.go
awgreene: This is good to know, thanks!

test/helpers/helpers.go
awgreene: Thanks!

test/testsuites/csctargetnamespacetests.go
awgreene: I will remove the check.

test/testsuites/csctargetnamespacetests.go
```diff
@@ -106,6 +118,40 @@ func succeededStateAfterTargetNamespaceCreated(t *testing.T) {
...
+
+	// Check that the Deployment was deleted.
+	err = helpers.WaitForNotFound(t, &apps.Deployment{}, namespace, cscName)
```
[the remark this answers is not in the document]
awgreene: `CheckCatalogSourceConfigAndChildResourcesDeleted` checks that the deployment was deleted, this shouldn't exist here. Thanks for the catch.

test/helpers/helpers.go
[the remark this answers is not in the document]
awgreene: Good idea. I will update the appropriate methods where this applies.

test/helpers/helpers.go
```diff
@@ -96,6 +99,31 @@ func WaitForSuccessfulDeployment(t *testing.T, deployment apps.Deployment) error
 	return nil
 }
 
+// WaitForExpectedPhaseAndMessage checks if a CatalogSourceConfig with the given name exists in the namespace
+// and makes sure that the phase and message matches the expected values.
+// If expectedMessage is an empty string, only the expectedPhase is checked.
+func WaitForExpectedPhaseAndMessage(t *testing.T, cscName string, namespace string, expectedPhase, expectedMessage string) error {
```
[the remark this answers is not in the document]
awgreene: I think this is a great idea but it seems like we would need to define an interface with functions that return the phase Name and Message. I would like to introduce those changes along with a refactor of this method in a separate PR.

awgreene:
@tkashem @kevinrizza @galletti94 Made some additional refactor changes and addressed all open comments.

PR #170, opened 2019-04-22
[e2e] Add opsrc delete test

kevinrizza:
-Added an e2e test which confirms that if an operator source is deleted
the resulting resources are deleted as well
-Added a helper function to create a resource without cleanup options

### DIFF
--- a/test/helpers/helpers.go
+++ b/test/helpers/helpers.go
@@ -135,6 +135,18 @@ func CreateRuntimeObject(client test.FrameworkClient, ctx *test.TestCtx, obj run
 		})
 }
 
+// CreateRuntimeObjectNoCleanup creates a runtime object without any cleanup
+// options. Using this method to create a runtime object means that the framework
+// will not automatically delete the object after test execution, and it must
+// be manually deleted.
+func CreateRuntimeObjectNoCleanup(client test.FrameworkClient, obj runtime.Object) error {
+	return client.Create(
+		context.TODO(),
+		obj,
+		nil,
+	)
+}
+
 // DeleteRuntimeObject deletes a runtime object using the test framework
 func DeleteRuntimeObject(client test.FrameworkClient, obj runtime.Object) error {
 	return client.Delete(
--- a/test/testsuites/opsrcdeletetests.go
+++ b/test/testsuites/opsrcdeletetests.go
@@ -1,0 +1,47 @@
+package testsuites
+
+import (
+	"testing"
+
+	"github.com/operator-framework/operator-marketplace/test/helpers"
+	"github.com/operator-framework/operator-sdk/pkg/test"
+	"github.com/stretchr/testify/require"
+)
+
+// DeleteOpSrc tests that the correct cleanup occurs when an OpSrc is deleted
+func DeleteOpSrc(t *testing.T) {
+	t.Run("delete-operator-source", testDeleteOpSrc)
+}
+
+// testDeleteOpSrc ensures that after deleting an OperatorSource that the
+// objects created as a result are deleted
+func testDeleteOpSrc(t *testing.T) {
+	ctx := test.NewTestCtx(t)
+	defer ctx.Cleanup()
+
+	// Get global framework variables
+	client := test.Global.Client
+
+	// Get test namespace
+	namespace, err := test.NewTestCtx(t).GetNamespace()
+	require.NoError(t, err, "Could not get namespace.")
+
+	testOperatorSource := helpers.CreateOperatorSourceDefinition(namespace)
+
+	// Create the OperatorSource with no cleanup options.
+	err = helpers.CreateRuntimeObjectNoCleanup(client, testOperatorSource)
+	require.NoError(t, err, "Could not create operator source.")
+
+	// Check for the datastore CatalogSourceConfig and its child resources.
+	err = helpers.CheckCatalogSourceConfigAndChildResourcesCreated(test.Global.Client, testOperatorSource.Name, namespace, namespace)
+	require.NoError(t, err, "Could not ensure that CatalogSourceConfig and its child resources were deleted")
+
+	// Now let's delete the OperatorSource
+	err = helpers.DeleteRuntimeObject(client, testOperatorSource)
+	require.NoError(t, err, "OperatorSource could not be deleted successfully. Client returned error.")
+
+	// Now let's wait until the OperatorSource is successfully deleted and the
+	// child resources are removed.
+	err = helpers.CheckCatalogSourceConfigAndChildResourcesDeleted(test.Global.Client, testOperatorSource.Name, namespace, targetNamespace)
+	require.NoError(t, err, "Could not ensure that CatalogSourceConfig and its child resources were deleted.")
+}
--- a/test/testgroups/nosetuptests.go
+++ b/test/testgroups/nosetuptests.go
@@ -10,4 +10,5 @@ import (
 func NoSetupTestGroup(t *testing.T) {
 	// Run the test suites.
 	t.Run("invalid-operator-source-test-suite", testsuites.InvalidOpSrc)
+	t.Run("delete-operator-source-test-suite", testsuites.DeleteOpSrc)
 }


test/testgroups/emptyresourcetests.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"testing"
+
+	"github.com/operator-framework/operator-marketplace/test/testsuites"
+)
+
+// EmptyResourceTestGroup runs tests without creating any resources upfront. Use this group if you can't
+// rely on the framework to clean up after your tests.
+func EmptyResourceTestGroup(t *testing.T) {
```
SamiSousa: This is similar to the test group that I'm creating in my PR, since my tests didn't require any prior resources. Should we use the same test group in this case?
Another question: should test groups be named for the resources they create (or don't create), or should they be used to organize similar/ related tests?

test/helpers/helpers.go
```diff
@@ -108,6 +135,27 @@ func CreateRuntimeObject(f *test.Framework, ctx *test.TestCtx, obj runtime.Objec
...
+	err := f.Client.Delete(
+		context.TODO(),
+		obj,
+	)
+
+	return err
+}
+
+// CreateRuntimeObject creates a runtime object using the test framework.
+func CreateRuntimeObjectNoCleanup(f *test.Framework, ctx *test.TestCtx, obj runtime.Object) error {
```
awgreene: Why does this method accept a ctx if it isn't using it?

test/helpers/helpers.go
```diff
@@ -108,6 +135,27 @@ func CreateRuntimeObject(f *test.Framework, ctx *test.TestCtx, obj runtime.Objec
 		})
 }
 
+// CreateRuntimeObject creates a runtime object using the test framework.
+func DeleteRuntimeObject(f *test.Framework, ctx *test.TestCtx, obj runtime.Object) error {
```
awgreene: Why does this method accept a ctx if it isn't using it?

test/testgroups/emptyresourcetests.go
awgreene: Honestly, there's no reason to create a test group if the test suites don't rely on any resources. The only benefit seems to be that the `TestMarketplace` function remains clean.

If we do want to go down the path of creating a test group for tests that don't require any setup, could we change the name? How does `EmptyResource` convey that no resources were created? What is an `EmptyResource`? Could we go with something clearer, such as `NoResourceSetupTestGroup`? Or just `NoSetupTestGroup`

test/testsuites/opsrcdeletiontests.go
```diff
@@ -0,0 +1,96 @@
...
+	operator "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	"github.com/operator-framework/operator-marketplace/test/helpers"
+	"github.com/operator-framework/operator-sdk/pkg/test"
+	apps "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+// OpSrcCreation is a test suite that ensures that the expected kubernetets resources are
+// created by marketplace after the creation of an OperatorSource.
```
awgreene: Should be

```
// OpSrcDeletion is a test suite that ensures that the expected kubernetets resources are
// deleted by after an OperatorSource is deleted.
```

test/testsuites/opsrcdeletiontests.go
```diff
@@ -0,0 +1,96 @@
...
+
+func testOperatorSourceDeletesSuccessfully(t *testing.T) {
+	ctx := test.NewTestCtx(t)
+	defer ctx.Cleanup()
+
+	// Get global framework variables.
+	f := test.Global
+
+	// Get test namespace
+	namespace, err := test.NewTestCtx(t).GetNamespace()
```
awgreene: `test.NewTestCtx(t).GetNamespace()` can just be `ctx.GetNamespace()`

test/helpers/helpers.go
```diff
@@ -108,6 +135,27 @@ func CreateRuntimeObject(f *test.Framework, ctx *test.TestCtx, obj runtime.Objec
...
+	err := f.Client.Delete(
+		context.TODO(),
+		obj,
+	)
+
+	return err
+}
+
+// CreateRuntimeObject creates a runtime object using the test framework.
```
awgreene: Should be
```
// CreateRuntimeObjectNoCleanup creates a runtime object without cleanup options using the test framework.
```

test/helpers/helpers.go
```diff
@@ -63,6 +67,29 @@ func WaitForResult(t *testing.T, result runtime.Object, namespace, name string)
...
+	err := wait.Poll(RetryInterval, DeleteTimeout, func() (done bool, err error) {
+		err = f.Client.Get(context.TODO(), namespacedName, result)
+		if err == nil {
+			t.Logf("Waiting for deletion of %s runtime object\n", name)
+			return false, nil
+		} else {
+			if errors.IsNotFound(err) {
+				return true, nil
+			} else {
+				return true, err
```
awgreene: Should we return true if the error isn't `errors.IsNotFound`?

test/helpers/helpers.go
aravindhp: I see that an helper exists for this at https://github.com/operator-framework/operator-marketplace/blob/master./test/helpers/helpers.go#L100. Can't that be reused?

test/testsuites/opsrcdeletiontests.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Create the OperatorSource with no cleanup options.
+	err = helpers.CreateRuntimeObjectNoCleanup(f, ctx, testOperatorSource)
+	if err != nil {
+		t.Errorf("Could not create operator source: %v", err)
+	}
+
+	// Now that the OperatorSource has been created, let's poll until the
+	// Status.CurrentPhase is Succeeded
+	expectedPhase := "Succeeded"
```
aravindhp: I am seeing this pattern being repeated. Please add a helper if possible.

test/testsuites/opsrcdeletiontests.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	// now let's delete the operator source
+	err = helpers.DeleteRuntimeObject(f, ctx, testOperatorSource)
+	if err != nil {
+		t.Errorf("OperatorSource could not be deleted successfully. Client returned error: %v", err)
+	}
+
+	// Let's check that we deleted it.
+	err = helpers.WaitForDelete(t, f, testOperatorSource, namespace, testOperatorSource.Name)
```
aravindhp: This should be modularized either here or in https://github.com/operator-framework/operator-marketplace/pull/167 and then reused.

awgreene: I suggest doing it in #167 given that it contains a refactor commit.

test/testsuites/opsrcdeletiontests.go
awgreene: This can be done in #167 as well

test/testgroups/emptyresourcetests.go
tkashem: I have not looked into the test framework in details yet, are the following assumptions correct?
* A group of test(s) share the same setup and teardown
* The test(s) inside a group run sequentially, while different group(s) may run in parallel
 
I am trying to see what criteria to consider when we create test group(s).

test/testsuites/opsrcdeletiontests.go
```diff
@@ -0,0 +1,96 @@
...
+		if err != nil {
+			return false, err
+		}
+		if testOperatorSource.Status.CurrentPhase.Name == expectedPhase {
+			return true, nil
+		}
+		return false, nil
+	})
+	if err != nil {
+		t.Errorf("OperatorSource never reached expected phase/message, expected %v", expectedPhase)
```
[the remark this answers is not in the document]
kevinrizza: I think it's a different case. For the case where the opsrc is still reconciling and we delete the reconciliation may have not created the child resources yet. We want to know that, if those resources are created, were they cleaned up afterwards.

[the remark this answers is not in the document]
tkashem: Okay, so this is the use case you are trying to validate - `for a successfully reconciled opsrc, all its resources(s) should be cleaned up when it is deleted`.

`t.Errorf` marks the test as failed but continues execution of the test. If the precondition `phase=Succeeded` is not met then the test is marked as failed and we can return/abort the test. There is no value moving forward.

kevinrizza:
@awgreene @tkashem @aravindhp @galletti94 @SamiSousa @anik120 

Rebased and significantly simplified this pr based on the changes made to our test harness over the last week. Please take another look.

test/helpers/helpers.go
```diff
@@ -135,6 +135,15 @@ func CreateRuntimeObject(client test.FrameworkClient, ctx *test.TestCtx, obj run
 		})
 }
 
+// CreateRuntimeObject creates a runtime object using the test framework.
```
aravindhp: Please mention the difference between `CreateRuntimeObject()`

test/testsuites/opsrcdeletetests.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		t.Errorf("Could not get namespace: %v", err)
+	}
+
+	testOperatorSource := helpers.CreateOperatorSourceDefinition(namespace)
+
+	// Create the OperatorSource with no cleanup options.
+	err = helpers.CreateRuntimeObjectNoCleanup(client, testOperatorSource)
+	if err != nil {
+		t.Errorf("Could not create operator source: %v", err)
```
aravindhp: Please use `assert` given that we are going that route now. Comment applies to all instances.

kevinrizza:
@aravindhp comments addressed.

test/testsuites/opsrcdeletetests.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// Get test namespace
+	namespace, err := test.NewTestCtx(t).GetNamespace()
+	assert.NoError(t, err, "Could not get namespace.")
+
+	testOperatorSource := helpers.CreateOperatorSourceDefinition(namespace)
+
+	// Create the OperatorSource with no cleanup options.
+	err = helpers.CreateRuntimeObjectNoCleanup(client, testOperatorSource)
+	assert.NoError(t, err, "Could not create operator source.")
```
tkashem: I am thinking whether it should be a `require.NoError`. If this precondition fails then we can immediately abort the test.

test/testsuites/opsrcdeletetests.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// Check for the datastore CatalogSourceConfig and its child resources.
+	err = helpers.CheckCatalogSourceConfigAndChildResourcesCreated(test.Global.Client, testOperatorSource.Name, namespace, namespace)
+	assert.NoError(t, err, "Could not ensure that CatalogSourceConfig and its cild resources were deleted")
```
tkashem: Same here, if we don't have a properly initialized `opsrc` then we can abort the test, I assume.

test/testsuites/opsrcdeletetests.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// Now let's delete the OperatorSource
+	err = helpers.DeleteRuntimeObject(client, testOperatorSource)
+	assert.NoError(t, err, "OperatorSource could not be deleted successfully. Client returned error.")
```
tkashem: Same here, `delete` being successful is a prerequisite I think.

test/testsuites/opsrcdeletetests.go
aravindhp: If this applies to https://github.com/operator-framework/operator-marketplace/pull/177, please mention it on that PR too.

kevinrizza: This is a good point. It also seems like, for basically every test we have that requires the previous step to complete before the next, almost every since call should be `require` instead of `assert`. In this case, I just updated the pr to do just that. We should probably ensure that this is also the case for https://github.com/operator-framework/operator-marketplace/pull/177

kevinrizza:
@aravindhp @tkashem addressed comments, please take another look

test/testsuites/opsrcdeletetests.go
```diff
@@ -0,0 +1,47 @@
...
+
+	testOperatorSource := helpers.CreateOperatorSourceDefinition(namespace)
+
+	// Create the OperatorSource with no cleanup options.
+	err = helpers.CreateRuntimeObjectNoCleanup(client, testOperatorSource)
+	require.NoError(t, err, "Could not create operator source.")
+
+	// Check for the datastore CatalogSourceConfig and its child resources.
+	err = helpers.CheckCatalogSourceConfigAndChildResourcesCreated(test.Global.Client, testOperatorSource.Name, namespace, namespace)
+	require.NoError(t, err, "Could not ensure that CatalogSourceConfig and its cild resources were deleted")
```
aravindhp: `child`

kevinrizza: @aravindhp nice catch, updated.

test/testsuites/opsrcdeletetests.go
tkashem: A test usually has three regions:
```
func Test(t *testing.T) {
  // Arrange
  // This is where the dependencies of function under test are resolved. If we fail to initialize here 
  // we need to abort the test. require is more appropriate. 
  require.NoError()

  // Act:
  // This is where we invoke the function under test.

  // Assert:
  // This is where we assert that output from act meets expectation. So assert is more appropriate 
  // here. if I have N asserts, I can see all the failed expectations.
  assert.NoError()
}
```

kevinrizza: @tkashem In the case of unit tests I would agree, but these are functional tests and we aren't just setting up objects and calling an API or a set of APIs and checking the result, each step (Arrange, Act, or Assert) is an actual functional API call. We aren't mocking anything, so we can't assume any result at any time.

test/testsuites/opsrcdeletetests.go
```diff
@@ -0,0 +1,47 @@
...
+	require.NoError(t, err, "Could not ensure that CatalogSourceConfig and its child resources were deleted")
+
+	// Now let's delete the OperatorSource
+	err = helpers.DeleteRuntimeObject(client, testOperatorSource)
+	require.NoError(t, err, "OperatorSource could not be deleted successfully. Client returned error.")
+
+	// Now let's wait until the OperatorSource is successfully deleted and the
+	// child resources are removed.
+	err = helpers.CheckCatalogSourceConfigAndChildResourcesDeleted(test.Global.Client, testOperatorSource.Name, namespace, targetNamespace)
+	require.NoError(t, err, "Could not ensure that CatalogSourceConfig and its child resources were deleted.")
```
anik120: @kevinrizza what's the difference between `require` and `assert`? I see that the signature is the same.

test/testsuites/opsrcdeletetests.go
tkashem: This is applicable to all type of tests - for this test `delete opsrc`:
Arrange or prerequisites: A properly initialized opsrc object.
Act: delete the opsrc object.
Assert: Make sure all resources are cleaned up.

So yeah, even if it is functional, we do need to setup expectations so that we can assert on them.

PR #174, opened 2019-04-30
[docs] Document Conflicting Package Name Behavior

awgreene:
Extend troubleshooting.md to describe how the marketplace operator handles
conflicting package names.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -13,6 +13,7 @@ Table of contents
 2. [OperatorSource failing to download from datastore](#operatorSource-failing-to-download-from-datastore)
 3. [OperatorSource `MESSAGE` reads unknown error (status 404)](#operatorsource-`message`-reads-unknown-error-(status-404)) 
 4. [CatalogSourceConfig stuck in `Configuring` phase](#catalogsourceconfig-stuck-in-configuring-phase)
+5. [Conflicting Package Names](#conflicting-package-names)
 
 
 ## No packages show up in the UI (No OperatorHub Items Found)
@@ -121,6 +122,15 @@ Some of the common reasons we have encountered that caused this `MESSAGE` are qu
 
 If you are trying to [install an operator from the cli](https://github.com/operator-framework/operator-marketplace#installing-an-operator-using-marketplace), the packages listed in the CatalogSourceConfig must have valid package names. If you have private app-resgistries, please follow the steps listed [here](https://github.com/operator-framework/operator-marketplace/blob/master/docs/how-to-authenticate-private-repositories.md).
 
+## Conflicting Package Names
+
+Package names are global across OperatorSources. To avoid unexpected behavior users should avoid adding any OperatorSources that contain existing packages.
+
+Users can view existing package names with the following command:
+
+```bash
+$ oc get packagemanifests -n openshift-marketplace
+```
 
 ## Where to go for help
 


awgreene:
I will rebase this PR once PR 171 is in.

awgreene:
Ready for review @kevinrizza @anik120 @tkashem @SamiSousa @galletti94

docs/troubleshooting.md
```diff
@@ -0,0 +1,10 @@
+## Conflicting Package Names
+
+If two Operator Sources claim the same package name the Marketplace operator does not allow the user to pick which grpc pod the operator is installed from. Instead, the Marketplace operator will randomly select one of the existing grpc pods to deploy the operator from. To avoid unexpected behavior users should avoid adding any operators with conflicting package names.
```
kevinrizza: nit: `operator-registry pod` instead of `grpc pod`

Also my preference is that, rather than describe the actual issue, we just mention that operator sources with conflicting package names are not supported -- I don't really see the need to describe the specific behavior of the limitation.

docs/troubleshooting.md
```diff
@@ -0,0 +1,10 @@
...
+
+Users can view packages available within their current namespace with the following command:
+```bash
+oc get packagemanifests
+```
+
+The operator-sources that are shipped with openshift will never have conflicting package names with one another.
```
kevinrizza: This is also a detail that is probably not necessary in open source documentation.

awgreene: Removed.

docs/troubleshooting.md
awgreene: Removed specific behavior in the latest commit.

docs/troubleshooting.md
```diff
@@ -0,0 +1,8 @@
+## Conflicting Package Names
+
+The Marketplace operator does not support reusing package names between two Operator Sources. To avoid unexpected behavior users should avoid adding any operator sources that reuse existing package names.
```
anik120: `nit`: Isn't `OperatorSource`, a marketplace API, different than `Operator Sources`?

kevinrizza: In addition to space/no space and singular/plural, also make sure that capitalization is consistent.

awgreene: I'm not sure what you are trying to highlight in this nit. I don't see where it mentions `OperatorSource`.

awgreene: Are you saying that we should use `OperatorSource` instead of `Operator Sources`?

anik120: @awgreene yes exactly that.

awgreene: Sure thing.

docs/troubleshooting.md
```diff
@@ -0,0 +1,8 @@
...
+
+Users can view existing package names with the following command:
+```bash
+$ oc get packagemanifests
```
SamiSousa: Should this use the marketplace namespace in order to list the packages available via operatorsources?
```bash
$ oc get packagemanifests -n marketplace
```
Although, maybe it should use `--all-namespaces` to list all existing packages on cluster since some may be installed via CatalogSource directly

awgreene: Changed to `--all-namespaces`.

awgreene:
Ready for review @kevinrizza @aravindhp @anik120

docs/troubleshooting.md
```diff
@@ -121,6 +122,15 @@ Some of the common reasons we have encountered that caused this `MESSAGE` are qu
...
 If you are trying to [install an operator from the cli](https://github.com/operator-framework/operator-marketplace#installing-an-operator-using-marketplace), the packages listed in the CatalogSourceConfig must have valid package names. If you have private app-resgistries, please follow the steps listed [here](https://github.com/operator-framework/operator-marketplace/blob/master/docs/how-to-authenticate-private-repositories.md).
 
+## Conflicting Package Names
+
+The Marketplace operator does not support reusing package names between two OperatorSources. To avoid unexpected behavior users should avoid adding any OperatorSources that reuse existing package names.
+
+Users can view existing package names with the following command:
+
+```bash
+$ oc get packagemanifests --all-namespaces
```
aravindhp: This should be `oc get packagemanifests -n openshift-marketplace`

docs/troubleshooting.md
```diff
@@ -121,6 +122,15 @@ Some of the common reasons we have encountered that caused this `MESSAGE` are qu
 
 If you are trying to [install an operator from the cli](https://github.com/operator-framework/operator-marketplace#installing-an-operator-using-marketplace), the packages listed in the CatalogSourceConfig must have valid package names. If you have private app-resgistries, please follow the steps listed [here](https://github.com/operator-framework/operator-marketplace/blob/master/docs/how-to-authenticate-private-repositories.md).
 
+## Conflicting Package Names
+
+The Marketplace operator does not support reusing package names between two OperatorSources. To avoid unexpected behavior users should avoid adding any OperatorSources that reuse existing package names.
```
aravindhp: I would rather say that Package names are global across OperatorSources. You should also mention what will happen if there are conflicting package names.

kevinrizza: I actually think it is better to avoid describing the particular behavior and issues associated with multiple package names -- if we do, and that behavior changes, then we have to update this doc to describe a different bug. Why not leave it as simply "marketplace doesn't currently support multiple operator sources with the same package names" ? What value do we get by describing the specific limitations?

aravindhp: Aren't both statements saying saying the same thing i.e. describing the specific limitations? Or are you saying that we don't mention what happens when there are conflicts? It is useful to describe that here because I can see that happening and users wondering what is going on. Moreover documentation is for describing current behavior. And yes, when that behavior changes we need to update the documentation.

kevinrizza: I just think in this specific case what we are describing is a possibly non deterministic behavior (because in some cases it will succeed if the in memory cache happens to respond just the right way) and describing it in that detail may leave users more confused than simply stating that it can cause unexpected behavior.

anik120: Please use `Marketplace Operator` (instead of `operator`) everywhere to be consistent with the rest of the document.

awgreene: Will do.

aravindhp: I am fine with adding a line explicitly stating that it can cause unexpected behavior.

awgreene: > I am fine with adding a line explicitly stating that it can cause unexpected behavior.

Isn't this already accomplished with the existing line: 
> To avoid unexpected behavior users should avoid adding any OperatorSources that reuse existing package names.

aravindhp: You are right. I take back my request. I would slightly modify the statement though. `To avoid unexpected behavior users should avoid adding any OperatorSources that contain existing packages`

awgreene: Wonderful - I will use your suggested wording, thanks for the review.

awgreene:
@aravindhp @kevinrizza @anik120 should be all set.

docs/troubleshooting.md
```diff
@@ -121,6 +122,15 @@ Some of the common reasons we have encountered that caused this `MESSAGE` are qu
...
 If you are trying to [install an operator from the cli](https://github.com/operator-framework/operator-marketplace#installing-an-operator-using-marketplace), the packages listed in the CatalogSourceConfig must have valid package names. If you have private app-resgistries, please follow the steps listed [here](https://github.com/operator-framework/operator-marketplace/blob/master/docs/how-to-authenticate-private-repositories.md).
 
+## Conflicting Package Names
+
+Package names are global across OperatorSources. To avoid unexpected behavior users should avoid adding any OperatorSources that contain existing packages.
+
+Users can view existing package names with the following command:
+
+```bash
+$ oc get packagemanifests --openshift-marketplace
```
SamiSousa: `-n openshift-marketplace`

aravindhp: Good catch @SamiSousa

awgreene: +1

PR #177, opened 2019-05-03
[e2e] Use asserts/require instead of error checks

anik120:
Use asserts/require instead of error checks

### DIFF
--- a/docs/e2e-testing.md
+++ b/docs/e2e-testing.md
@@ -35,9 +35,7 @@ func genericTestSuite(t *testing.T) {
 
     // Create the operatorsource.
     err = helpers.CreateRuntimeObject(f, ctx, helpers.CreateOperatorSource(namespace))
-    if err != nil {
-        t.Errorf("Could not create operator source: %v", err)
-    }
+    require.NoError(t, err, "Could not create OperatorSource")
 
     // Run the tests that rely on the runtime objects created earlier.
     t.Run("test1", testsuites.Test1)
--- a/test/testsuites/csctargetnamespacetests.go
+++ b/test/testsuites/csctargetnamespacetests.go
@@ -7,6 +7,8 @@ import (
 	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
 	"github.com/operator-framework/operator-marketplace/test/helpers"
 	"github.com/operator-framework/operator-sdk/pkg/test"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -21,9 +23,7 @@ func CscTargetNamespace(t *testing.T) {
 
 	// Get test namespace.
 	namespace, err := ctx.GetNamespace()
-	if err != nil {
-		t.Fatalf("Could not get namespace: %v", err)
-	}
+	require.NoError(t, err, "Could not get namespace")
 
 	// Create a new CatalogSourceConfig with a non-existing targetNamespace.
 	nonExistingTargetNamespaceCsc := &marketplace.CatalogSourceConfig{
@@ -42,9 +42,7 @@ func CscTargetNamespace(t *testing.T) {
 	// rely on the existence of the CatalogSourceConfig.
 	// The CatalogSourceConfig is created with nil ctx and must be deleted manually before test suite exits.
 	err = helpers.CreateRuntimeObject(client, nil, nonExistingTargetNamespaceCsc)
-	if err != nil {
-		t.Fatalf("Unable to create test CatalogSourceConfig: %v", err)
-	}
+	require.NoError(t, err, "Unable to test CatalogSourceConfig")
 
 	// Run tests.
 	t.Run("configuring-state-when-target-namespace-dne", configuringStateWhenTargetNamespaceDoesNotExist)
@@ -52,9 +50,7 @@ func CscTargetNamespace(t *testing.T) {
 	// Create a namespace based on the targetNamespace string.
 	resultNamespace := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: targetNamespace}}
 	err = helpers.CreateRuntimeObject(client, ctx, resultNamespace)
-	if err != nil {
-		t.Fatalf("Unable to create test namespace: %v", err)
-	}
+	require.NoError(t, err, "Unable to create test namespace")
 
 	t.Run("succeeded-state-after-target-namespace-created", succeededStateAfterTargetNamespaceCreated)
 
@@ -62,9 +58,7 @@ func CscTargetNamespace(t *testing.T) {
 
 	// Delete the CatalogSourceConfig.
 	err = helpers.DeleteRuntimeObject(client, nonExistingTargetNamespaceCsc)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err, "Unable to delete CatalogSourceConfig")
 
 	t.Run("child-resources-deleted", childResourcesDeleted)
 }
@@ -74,18 +68,14 @@ func CscTargetNamespace(t *testing.T) {
 // nonexistent namespace.
 func configuringStateWhenTargetNamespaceDoesNotExist(t *testing.T) {
 	namespace, err := test.NewTestCtx(t).GetNamespace()
-	if err != nil {
-		t.Fatalf("Could not get namespace: %v", err)
-	}
+	require.NoError(t, err, "Could not get namespace")
 
 	// Check that the CatalogSourceConfig with an non-existing targetNamespace eventually reaches the
 	// configuring phase with the expected message.
 	expectedPhase := "Configuring"
 	expectedMessage := fmt.Sprintf("namespaces \"%s\" not found", targetNamespace)
 	err = helpers.WaitForExpectedPhaseAndMessage(test.Global.Client, cscName, namespace, expectedPhase, expectedMessage)
-	if err != nil {
-		t.Fatalf("CatalogSourceConfig never reached expected phase/message, expected %v/%v", expectedPhase, expectedMessage)
-	}
+	assert.NoError(t, err, fmt.Sprintf("CatalogSourceConfig never reached expected phase/message, expected %v/%v", expectedPhase, expectedMessage))
 }
 
 // succeededStateAfterTargetNamespaceCreated is a test case that confirms that a csc that had a
@@ -93,17 +83,13 @@ func configuringStateWhenTargetNamespaceDoesNotExist(t *testing.T) {
 func succeededStateAfterTargetNamespaceCreated(t *testing.T) {
 	// Get test namespace.
 	namespace, err := test.NewTestCtx(t).GetNamespace()
-	if err != nil {
-		t.Fatalf("Could not get namespace: %v", err)
-	}
+	require.NoError(t, err, "Could not get namespace")
 
 	// Now that the targetNamespace has been created, periodically check that the CatalogSourceConfig
 	// has reached the Succeeded phase.
 	expectedPhase := "Succeeded"
 	err = helpers.WaitForExpectedPhaseAndMessage(test.Global.Client, cscName, namespace, expectedPhase, "")
-	if err != nil {
-		t.Fatalf("CatalogSourceConfig never reached expected phase, expected %v", expectedPhase)
-	}
+	assert.NoError(t, err, fmt.Sprintf("CatalogSourceConfig never reached expected phase, expected %v", expectedPhase))
 }
 
 // childResourcesCreated checks that once a CatalogSourceConfig is created that all expected runtime
@@ -111,15 +97,10 @@ func succeededStateAfterTargetNamespaceCreated(t *testing.T) {
 func childResourcesCreated(t *testing.T) {
 	// Get test namespace.
 	namespace, err := test.NewTestCtx(t).GetNamespace()
-	if err != nil {
-		t.Fatalf("Could not get namespace: %v", err)
-	}
+	require.NoError(t, err, "Could not get namespace")
 	// Check that the CatalogSourceConfig and its child resources were created.
 	err = helpers.CheckCscSuccessfulCreation(test.Global.Client, cscName, namespace, targetNamespace)
-	if err != nil {
-		t.Fatal(err)
-
-	}
+	assert.NoError(t, err)
 }
 
 // childResourcesDeleted checks that once a CatalogSourceConfig is deleted that all expected runtime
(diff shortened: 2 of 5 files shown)


test/testsuites/opsrccreationtests.go
```diff
@@ -23,34 +24,21 @@ func OpSrcCreation(t *testing.T) {
...
-		t.Errorf(
-			"The created CatalogSource %s does not have the right label[%s] - want=%s got=%s",
-			resultCatalogSource.Name,
-			helpers.TestOperatorSourceLabelKey,
-			helpers.TestOperatorSourceLabelValue,
-			groupGot,
-		)
-	}
+	assert.Equal(t, true, ok)
+	assert.Equal(t, groupGot, helpers.TestOperatorSourceLabelValue)
```
tkashem: This is the signature of `assert.Equal`
```
Equal(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) bool {
```

Suggestion: By convention we specify expected first and then actual. `groupGot` is actual and I think `helpers.TestOperatorSourceLabelValue` is your expected here.

aravindhp: @anik120 please ensure that you follow this convention in all places.

anik120:
@tkashem thank you for your feedback. I've addressed your comments. 
Note that I've preserved all comments expect for the comment `CatalogSourceConfig never reached expected phase/message, expected %v/%v", expectedPhase, expectedMessage`. Since there is an `assert.NoError` statement in place of that, I believe the `assert` statement will print out the appropriate message instead. Please let me know if the previous statement is incorrect.

aravindhp:
> Should I update my invalid OpSrc test using `assert` in a follow up PR?

@SamiSousa If that has been merged then it should be done in this PR by @anik120

anik120:
@SamiSousa I pull your changes down. Thank you for notifying me about the merged PR. I'm updating your test now. 
cc: @aravindhp

anik120:
@SamiSousa updated `invalidOpSrc` test. Please review.

anik120:
@tkashem @aravindhp @SamiSousa introduced `require.NoError`. Ready for review.

test/testsuites/csctargetnamespacetests.go
```diff
@@ -82,66 +76,49 @@ func CscTargetNamespace(t *testing.T) {
...
-		t.Fatalf("Could not get namespace: %v", err)
-	}
+	require.NoError(t, err, "Could not get namespace")
 	// Check that the CatalogSourceConfig and its child resources were created.
 	err = helpers.CheckCatalogSourceConfigAndChildResourcesCreated(test.Global.Client, cscName, namespace, targetNamespace)
-	if err != nil {
-		t.Fatal(err)
-
-	}
+	assert.NoError(t, err)
```
aravindhp: My suggestion is for this helper to return the error and have `asserts` or `require` in the call site. This comment applies for all helpers. 
PS: Please wait for comment from others before making this change.

aravindhp:
@anik120 I suggest saying `Use asserts/require in error checks` for both commit and PR headers and messages.

test/testsuites/csctargetnamespacetests.go
anik120: @aravindhp looks like helpers like these are just returning the `client`'s result. Doesn't look like the helpers need to return custom errors, and can just return whatever error the `client` might return.

aravindhp: My point was to not have `asserts` or `require` in the helpers. I never said anything about custom errors.

kevinrizza: +1, there's no reason for the helpers to have anything succeed or fail directly. They should just return information about what occurred, and the test methods should do the `asserting` and `requiring`

anik120: @aravindhp `csctargetnamespacetests.go` is a test file. None of the helper functions use `asserts` or `require` as of now. They return the error, which are then `assert`-ed or `require`-ed at the call site, as you were suggesting.

anik120:
@aravindhp changed commit messages and PR header and message.

test/testsuites/invalidopsrctests.go
```diff
@@ -78,9 +76,7 @@ func testOpSrcWithInvalidEndpoint(t *testing.T) {
 		}
 		return false, nil
 	})
-	if err != nil {
-		t.Errorf("OperatorSource never reached expected phase/message, expected %v: %v", expectedPhase, err)
-	}
+	assert.NoError(t, err)
```
SamiSousa: Why remove the message for this one? Does `assert.NoError` not support string formatting?

SamiSousa: @anik120 FYI my comment refers to each instance of this

test/testsuites/csctargetnamespacetests.go
aravindhp: `childResourcesCreated()` for practical purposes is a helper function. Point I have been trying to get across is to call `asserts` and `requires` in the main test functions and nowhere else.

test/testsuites/invalidopsrctests.go
aravindhp: +1 on this. I thought @tkashem had already made this request in his review.

awgreene:
@anik120 Please update [e2e doc](https://github.com/operator-framework/operator-marketplace/blob/master/docs/e2e-testing.md) with assert example / reasoning.

aravindhp:
> @anik120 Please update [e2e doc](https://github.com/operator-framework/operator-marketplace/blob/master/docs/e2e-testing.md) with assert example / reasoning.

Not sure if we need to call that out. It is pretty standard way doing things in tests.

test/testsuites/csctargetnamespacetests.go
anik120: @aravindhp looks like `childResourcesCreated()` is a func that tests an individual unit (existence of child resources after creation of a CSC). It uses a helper itself, i.e `CheckCatalogSourceConfigAndChildResourcesCreated()`, and it itself isn't being called by any other function as a `helper`. I don't see how `childResourcesCreated()` is a `helper` function.

awgreene:
>Not sure if we need to call that out. It is pretty standard way doing things in tests.

I don't agree @aravindhp, In the e2e test documentation we showcase [example](https://github.com/operator-framework/operator-marketplace/blob/master/docs/e2e-testing.md#test-suites)  that uses the outdated `err != nil` approach. If we've decided that this is the wrong approach, why highlight it?

anik120:
> I don't agree @aravindhp, In the e2e test documentation we showcase example that uses the outdated err != nil approach. If we've decided that this is the wrong approach, why highlight it?

@awgreene I see your point. Updating the e2e doc along with this PR.

awgreene:
>Not sure if we need to call that out. It is pretty standard way doing things in tests.

I see your point regarding not highlighting the reasoning.

aravindhp:
> > Not sure if we need to call that out. It is pretty standard way doing things in tests.
> 
> I don't agree @aravindhp. In the e2e test documentation we showcase an [example](https://github.com/operator-framework/operator-marketplace/blob/master/docs/e2e-testing.md#test-suites) that uses the outdated `err != nil` approach. If we've decided that this is the wrong approach, why highlight it?

It is fine to update the code snippet

kevinrizza:
> > I don't agree @aravindhp. In the e2e test documentation we showcase an [example](https://github.com/operator-framework/operator-marketplace/blob/master/docs/e2e-testing.md#test-suites) that uses the outdated `err != nil` approach. If we've decided that this is the wrong approach, why highlight it?
> 
> It is fine to update the code snippet

+1, just update the examples

anik120:
@aravindhp @awgreene @SamiSousa rebased and updated e2e doc, message included for error at the spot that I missed. Please review.

test/testsuites/invalidopsrctests.go
```diff
@@ -41,9 +43,7 @@ func testOpSrcWithInvalidEndpoint(t *testing.T) {
 
 	// Get test namespace
 	namespace, err := ctx.GetNamespace()
-	if err != nil {
-		t.Errorf("Could not get namespace: %v", err)
-	}
+	assert.NoError(t, err, "Could not get namespace")
```
aravindhp: This should be `require`

docs/e2e-testing.md
```diff
@@ -35,9 +35,7 @@ func genericTestSuite(t *testing.T) {
 
     // Create the operatorsource.
     err = helpers.CreateRuntimeObject(f, ctx, helpers.CreateOperatorSource(namespace))
-    if err != nil {
-        t.Errorf("Could not create operator source: %v", err)
-    }
+    require.NoError(t, err, "Could not create OperatorSource")
```
aravindhp: Any reason to not print the error using `fmt.Sprintf("Could not create OperatorSource: %v", err)`? Comment applies to all places.

anik120: @aravindhp I believe the test framework prints out the error in a standard format if the assertion or require fail. in fact, I am still of the opinion that we don't need any custom error messages at all, but I we decided to keep them.

aravindhp: Please post a snippet of the output.

anik120: @aravindhp I changed an `assert.True` to `assert.False` without any custom output message and this is the log from the logger: 
```
FAIL: TestMarketplace (165.64s)
    --- FAIL: TestMarketplace/operator-source-test-group (98.38s)
    	client.go:57: resource type OperatorSource with namespace/name (openshift-marketplace/test-operators) created
        --- FAIL: TestMarketplace/operator-source-test-group/opsrc-creation-test-suite (25.29s)
            --- FAIL: TestMarketplace/operator-source-test-group/opsrc-creation-test-suite/operator-source-generates-expected-objects (25.29s)
            	opsrccreationtests.go:43: 
            			Error Trace:	opsrccreationtests.go:43
            			Error:      	Should be false
            			Test:       	TestMarketplace/operator-source-test-group/opsrc-creation-test-suite/operator-source-generates-expected-objects
```

anik120: I forced an error before a `require`: 
```
err = k8s_errors.NewAlreadyExists(schema.GroupResource{}, "This is a test error")
require.NoError(t, err)
```
and this is the output: 
```--- FAIL: TestMarketplace (145.67s)
    --- FAIL: TestMarketplace/operator-source-test-group (83.33s)
        client.go:57: resource type OperatorSource with namespace/name (openshift-marketplace/test-operators) created
        --- FAIL: TestMarketplace/operator-source-test-group/opsrc-creation-test-suite (20.26s)
            --- FAIL: TestMarketplace/operator-source-test-group/opsrc-creation-test-suite/operator-source-generates-expected-objects (20.26s)
                require.go:794:
                        Error Trace:    opsrccreationtests.go:35
                        Error:          Received unexpected error:
                                         "This is test error" already exists
                        Test:           TestMarketplace/operator-source-test-group/opsrc-creation```

Comment (anik120) [reply docs/e2e-testing.md]: Given these output, I vote for not having custom error messages as they can get too verbose and the output from these error point exactly to the `assert`/`require` statement that's failing which is efficient enough.

Comment (aravindhp) [reply docs/e2e-testing.md]: I disagree. This will force us to go look at the source code to figure out what is wrong. If there is a properly worded error messages, we can figure out what is wrong at a glance. This will be important especially in the CI pipeline and for folks who are not familiar with the operator. But if others agree with you I will take back my objection.

Comment (anik120) [reply docs/e2e-testing.md]: @aravindhp updated PR with `fmt.Sprintf` at all places.

**Comment by anik120** [comment]
@kevinrizza @awgreene let me know if you guys have any other thoughts on this. Should be ready otherwise.

File: docs/e2e-testing.md (hunk 5 above)
Comment (awgreene) [reply docs/e2e-testing.md]: Quick question - I reran the test locally:

```
func qTest(t *testing.T) {
	err := fmt.Errorf("this is my error")
	require.NoError(t, err, fmt.Sprintf("error: %v", err))

	/* OUTPUT
	Error Trace:	main_test.go:72
	Error:      	Received unexpected error:
					this is my error
	Test:       	TestMarketplace/test
	Messages:   	error: this is my error
	*/
}
```

It looks like the message is printed twice in the current format. Is this intended?

File: docs/e2e-testing.md [hunk 6]
```diff
@@ -35,9 +35,7 @@ func genericTestSuite(t *testing.T) {
 
     // Create the operatorsource.
     err = helpers.CreateRuntimeObject(f, ctx, helpers.CreateOperatorSource(namespace))
-    if err != nil {
-        t.Errorf("Could not create operator source: %v", err)
-    }
+    require.NoError(t, err, fmt.Sprintf("Could not create OperatorSource: %v", err))
```
awgreene: This is an example of: https://github.com/operator-framework/operator-marketplace/pull/177#discussion_r285264592

docs/e2e-testing.md
anik120: Interesting. One of them is from the test framework, and the other is the `custom` message. According to @aravindhp's point, the error from the test framework can be used to hunt down the error in the code, while the `custom` error message will help out in CI pipeline. 
@awgreene does the output look different if you use `require.NoError(t, err, "this is my error")` instead?

awgreene: I believe the example provided ealier, https://github.com/operator-framework/operator-marketplace/pull/177#discussion_r284726972, failed to add a message which is why this slipped through.

awgreene: >....@awgreene does the output look different if you use require.NoError(t, err, "this is my error") instead?

```
func qTest(t *testing.T) {
	err := fmt.Errorf("this is my error")
	//require.NoError(t, err, fmt.Sprintf("error: %v", err))
	require.NoError(t, err, "this is my error")
	/* OUTPUT
	Error Trace:	main_test.go:73
	Error:      	Received unexpected error:
					this is my error
	Test:       	TestMarketplace/test
	Messages:   	this is my error

	*/
}
```

aravindhp: If the output provides the error, I am fine leaving the error out from the message. In that case `fmt.Sprintf()` should only be used in cases where we trying to show some other variable for additional informational. But I am still a +1 on custom error messages.

anik120: @awgreene in [#177(comment)](https://github.com/operator-framework/operator-marketplace/pull/177#discussion_r284726972) the message has been left out intentionally to showcase what the test framework prints `without` a `custom` error message.

awgreene: > @awgreene in #177(comment) the message has been left out intentionally to showcase what the test framework prints without a custom error message.

Ah thank you

awgreene: > If the output provides the error, I am fine leaving the error out from the message. In that case fmt.Sprintf() should only be used in cases where we trying to show some other variable for additional informational. But I am still a +1 on custom error messages.


+1

anik120: @aravindhp @awgreene updated PR to have custom error messages, while not repeating the same error message in the log. Let me know your thoughts.

test/testsuites/invalidopsrctests.go
```diff
@@ -78,9 +77,7 @@ func testOpSrcWithInvalidEndpoint(t *testing.T) {
 		}
 		return false, nil
 	})
-	if err != nil {
-		t.Errorf("OperatorSource never reached expected phase/message, expected %v: %v", expectedPhase, err)
-	}
+	assert.NoError(t, err, fmt.Sprintf("OperatorSource never reached expected phase/message, expected %v: %v", expectedPhase, err))
```
awgreene: This will never print the actual phase. Applies in all instances. I do not believe that it needs to be fixed in this PR.

anik120: @awgreene Not sure I follow. Why won't it print `OperatorSource never reached expected phase/message, expected Downloading: ....error message...` ?

aravindhp: +1. You can also drop the `err` as you have done elsewhere. Applies to all instances.

anik120: @awgreene @aravindhp dropped `err` from all similar instances.

PR #178, opened 2019-05-03
Fix tolerations

ravisantoshgudimetla:
In 4.1, we have taint-based evictions enabled. This changes logic in the node lifecycle controller in the kube-controller-manager significantly.

Historically, the node lifecycle controller would directly evict pods from node that had a Ready condition of False or Unknown after a pod eviction timeout set by the --pod-eviction-timeout flag on the kube-controller-manager. This setting applied to all pods cluster-wide. The default was 5m.

With taint-based evictions, all the node lifecycle controller does is taint the node with node.kubernetes.io/unreachable and/or node.kubernetes.io/not-ready taint with NoExecute effect. This would normally result in the immediate eviction of all pods that don't tolerate those taints, breaking the old behavior with pod eviction timeouts. Enter DefaultTolerationSeconds mutating admission plugin.

https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/plugin/pkg/admission/defaulttolerationseconds/admission.go

This plugin will add a default NoExecute toleration for node.kubernetes.io/unreachable and/or node.kubernetes.io/not-ready taints with a tolerationSeconds of 5m (300s) as long as no such toleration is already specified in the pod spec. This restores the old pod eviction tiemout behavior.

One of the intended effects of this change is the make the pod eviction timeout a pod-level property. Different applications require different timeouts depending on their design and having it controlled at a cluster level before was not optimal. The side-effect is that we won't allow pods to be scheduled onto nodes that have disk, memory, cpu pressure.

The DefaultTolerationSeconds plugin has a flag that allows adjusting the defaults for tolerationSeconds
https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/plugin/pkg/admission/defaulttolerationseconds/admission.go#L34-L40

We might expose this tunable for user control in the future and do not want the cluster control plane components to be subject to it.

Thus, this PR explictly defines a NoExecute toleration for node.kubernetes.io/unreachable and/or node.kubernetes.io/not-ready taints with a tolerationSeconds generically appropriate for cluster components.

Once these changes are in across all components, this e2e will enforce it in the future
openshift/origin#22752

Please refer to this doc, if you have questions around what tolerations you can have:

https://docs.google.com/document/d/1W449BfB5la9NC7pcDxkovgzlBlHMy5Lkj0-WXWjdiTU/edit#

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -18,7 +18,17 @@ spec:
         node-role.kubernetes.io/master: ""
       priorityClassName: "system-cluster-critical"
       tolerations:
-      - operator: Exists
+      - key: "node-role.kubernetes.io/master"
+        operator: Exists
+        effect: "NoSchedule"
+      - key: "node.kubernetes.io/unreachable"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 120
+      - key: "node.kubernetes.io/not-ready"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 120
       containers:
         - name: marketplace-operator
           image: quay.io/openshift/origin-operator-marketplace:latest
--- a/manifests/10_operator.yaml
+++ b/manifests/10_operator.yaml
@@ -18,7 +18,17 @@ spec:
         node-role.kubernetes.io/master: ""
       priorityClassName: "system-cluster-critical"
       tolerations:
-      - operator: Exists
+      - key: "node-role.kubernetes.io/master"
+        operator: Exists
+        effect: "NoSchedule"
+      - key: "node.kubernetes.io/unreachable"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 120
+      - key: "node.kubernetes.io/not-ready"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 120
       containers:
         - name: marketplace-operator
           image: quay.io/openshift/origin-operator-marketplace:latest


aravindhp:
@ravisantoshgudimetla please follow the 50/80 rule in the git commit message

deploy/operator.yaml
```diff
@@ -18,7 +18,14 @@ spec:
         node-role.kubernetes.io/master: ""
       priorityClassName: "system-cluster-critical"
       tolerations:
-      - operator: Exists
+      - key: "node.kubernetes.io/unreachable"
```
SamiSousa: Should these changes also be applied to the upstream manifests? Or are these tolerations only needed for OKD/OpenShift clusters?

kevinrizza:
@ravisantoshgudimetla what about pods created as operands? Should that apply to those as well?

ravisantoshgudimetla:
> @ravisantoshgudimetla what about pods created as operands? Should that apply to those as well?

Which operand are you talking about? Other operators managed by olm? 

These should be applied on a case-to-case basis.

aravindhp:
> > @ravisantoshgudimetla what about pods created as operands? Should that apply to those as well?
> 
> Which operand are you talking about? Other operators managed by olm?

@kevinrizza is asking about the operator-registry pods that the marketplace operator creates.

ravisantoshgudimetla:
> @kevinrizza is asking about the operator-registry pods that the marketplace operator creates.

Not sure about that, how are they run usually? on masters? on workers?

aravindhp:
> Not sure about that, how are they run usually? on masters? on workers?

We don't explicitly specify this. It is wherever the scheduler schedules those pods.

ravisantoshgudimetla:
> We don't explicitly specify this. It is wherever the scheduler schedules those pods.

Then it's not going to be behavioural change for those pods. So, where are the pod specs located for them? Perhaps we can create PRs for them in a follow-up PR?

kevinrizza:
> > We don't explicitly specify this. It is wherever the scheduler schedules those pods.
> 
> Then it's not going to be behavioural change for those pods. So, where are the pod specs located for them? Perhaps we can discuss about the behavioural change for them and make changes in a follow-up PR?

We create these pods in one of our controller's reconcilers

ravisantoshgudimetla:
> We create these pods in one of our controller's reconcilers

Grepping for tolerations gave me these 2 yamls. If you have code related to it, please feel to create a follow-up PR and tag me on it.

kevinrizza:
> > We create these pods in one of our controller's reconcilers
> 
> Grepping for tolerations gave me these 2 yamls. If you have code related to it, please feel to create a follow-up PR and tag me on it.

@ravisantoshgudimetla Hey sorry I think my question is more specific:

In addition to the yaml that creates our operator, we also create pods (in go code) when our operator responds to certain custom resources. So my question is, do we need to enforce this toleration on all of our operand pods? Or does this policy only need to affect the component operator pod itself?

Otherwise it seems like this change is fine in a vacuum, I just want to ensure that we are making all of the changes necessary. If adding this toleration to those pods is indeed necessary, we can address it in a followup pr.

/lgtm

aravindhp:
In other word @ravisantoshgudimetla, we are looking for guidance from you.

/approve

ravisantoshgudimetla:
> In other word @ravisantoshgudimetla, we are looking for guidance from you.

Would be more than happy to help, that's why, I asked to be tagged on the PR :).

aravindhp:
>Would be more than happy to help, that's why, I asked to be tagged on the PR :)

@ravisantoshgudimetla we would actually like some guidance before opening a PR :-) Somebody from the marketplace team will talk to you.

PR #179, opened 2019-05-06
Bug 1706867: ClusterOperator degrading on exit

awgreene:
Problem:
When the marketplace operator is stopped and restarted, it should not
implicitly mark the ClusterOperator status as degrading.

Solution:
When the Operator exits, simply log that the operator has exited and exit
gracefully.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1706867

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -67,24 +67,24 @@ func main() {
 	catalogsourceconfig.InitializeStaticSyncer(mgr.GetClient(), initialWait)
 	registrySyncer := operatorsource.NewRegistrySyncer(mgr.GetClient(), initialWait, resyncInterval, updateNotificationSendWait, catalogsourceconfig.Syncer, catalogsourceconfig.Syncer)
 
-	// monitorStopCh is used to send a signal to stop reporting cluteroperator status
+	// monitorStopCh is used to send a signal to stop reporting cluster operator status
 	monitorStopCh := make(chan struct{})
 	// monitorDoneCh will recieve a signal when threads have stopped updating cluster operator status
 	monitorDoneCh := status.StartReporting(cfg, mgr, namespace, os.Getenv("RELEASE_VERSION"), monitorStopCh)
 
 	// Setup Scheme for all defined resources
 	if err := apis.AddToScheme(mgr.GetScheme()); err != nil {
-		fatal(err, monitorDoneCh, monitorStopCh)
+		exit(err, monitorDoneCh, monitorStopCh)
 	}
 
 	// Add external resource to scheme
 	if err := olm.AddToScheme(mgr.GetScheme()); err != nil {
-		fatal(err, monitorDoneCh, monitorStopCh)
+		exit(err, monitorDoneCh, monitorStopCh)
 	}
 
 	// Setup all Controllers
 	if err := controller.AddToManager(mgr); err != nil {
-		fatal(err, monitorDoneCh, monitorStopCh)
+		exit(err, monitorDoneCh, monitorStopCh)
 	}
 
 	// Serve a health check.
@@ -100,18 +100,24 @@ func main() {
 	go catalogsourceconfig.Syncer.Sync(stopCh)
 
 	// Start the Cmd
-	// Update clusteroperator status and log a fatal error
-	fatal(mgr.Start(stopCh), monitorDoneCh, monitorStopCh)
+	// Update ClusterOperator status and log a fatal error
+	exit(mgr.Start(stopCh), monitorDoneCh, monitorStopCh)
 }
 
-// fatal updates clusteroperator status and logs a fatal error
-func fatal(err error, monitorDoneCh <-chan struct{}, monitorStopCh chan struct{}) {
-	// Stop reporting clusterOperator status
+// exit updates the ClusterOperator status and exits with the correct error code.
+func exit(err error, monitorDoneCh <-chan struct{}, monitorStopCh chan struct{}) {
+	// Stop reporting ClusterOperator status
 	close(monitorStopCh)
 
-	// Wait for clusterOperator status reporting to stop
+	// Wait for ClusterOperator status reporting to stop
 	<-monitorDoneCh
 
-	// exit
-	log.Fatal(err)
+	// If an error exists then exit with status set to 1
+	if err != nil {
+		log.Fatalf("The operator encountered an error, exit code 1: %v", err)
+	}
+
+	// No error, graceful termination
+	log.Info("The operator exited gracefully, exit code 0")
+	os.Exit(0)
 }
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -289,8 +289,8 @@ func (s *status) monitorClusterStatus() {
 			// be set to degraded.
 			conditionListBuilder := clusterStatusListBuilder()
 			conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionFalse, "")
-			conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionFalse, "")
-			statusConditions := conditionListBuilder(configv1.OperatorDegraded, configv1.ConditionTrue, "Operator exited")
+			conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionFalse, "The operator has exited and is no longer reporting status.")
+			statusConditions := conditionListBuilder(configv1.OperatorDegraded, configv1.ConditionFalse, "")
 			statusErr := s.setStatus(statusConditions)
 			if statusErr != nil {
 				log.Error("[status] " + statusErr.Error())


pkg/status/status.go
```diff
@@ -285,16 +285,7 @@ func (s *status) monitorClusterStatus() {
 	for {
 		select {
 		case <-s.stopCh:
-			// If the stopCh is closed, the operator will exit and CO should
-			// be set to degraded.
-			conditionListBuilder := clusterStatusListBuilder()
-			conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionFalse, "")
-			conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionFalse, "")
-			statusConditions := conditionListBuilder(configv1.OperatorDegraded, configv1.ConditionTrue, "Operator exited")
```
aravindhp: It will be useful if we updated the `Message` to reflect what happened. Example: `Operator exited with status...`. I am also worried that we are not looking at the exit values here and an actual issue will be ignored.

awgreene: @aravindhp
>It will be useful if we updated the Message to reflect what happened. Example: Operator exited with status...

That's possible, but we will need to update how we listen for `SIGTERM` and `SIGINT`. 

We currently [start the manager](https://github.com/operator-framework/operator-marketplace/blob/master/cmd/manager/main.go#L104) using OLM's [signals package](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/lib/signals/signals.go) to and will need to reproduce the function if we want to modify it. Thoughts?

awgreene: Also - where should we track the message - are you suggesting that we report `degraded` as false but update its message?

aravindhp: >We currently start the manager using OLM's signals package to and will need to reproduce the function if we want to modify it. Thoughts?

If changes are needed we can copy the function over and modify it. You can also talk to the OLM folks about it to check if they are facing a similar issue.

>Also - where should we track the message - are you suggesting that we report degraded as false but update its message?

If I look at the cluster version page on the UI it should be reflected there and from the CLI. I leave it to you as to where it needs to be updated.

awgreene: > If changes are needed we can copy the function over and modify it. You can also talk to the OLM folks about it to check if they are facing a similar issue.

After investigating how to report the exit code to the cluster operator I don't believe that we should report the exit code in the cluster operator status. Instead, we can simply log the exit code in the fatal function and reflect that the operator has exited in the cluster operator status like so:

Updated fatal code:
```
// exit updates the clusteroperator status and exits with the correct error code.
func exit(err error, monitorDoneCh <-chan struct{}, monitorStopCh chan struct{}) {
	// Stop reporting clusterOperator status
	close(monitorStopCh)

	// Wait for clusterOperator status reporting to stop
	<-monitorDoneCh

	// If an error exists then exit with status set to 1
	if err != nil {
		log.Fatalf("The operator encountered an error, exit code 1: %v", err)
	}

	// No error, graceful termination
	log.Info("The operator exited gracefully, exit code 0")
	os.Exit(0)
}
```

The `CO` status should be updated so degraded is not set to true and to update the available message to let the admin know that it is no longer available:
```
Available = false, "The operator has exited and is no longer reporting status."
Progressing = false, ""
Degraded = false, ""
```
This approach will let CVO know that marketplace is no longer available and give admins a full idea of what happen in the logs while introducing minimal changes.

aravindhp: >Example updated fatal code

What is the point in just logging? You are still not checking to see what the signal was. Moreover you are giving a false negative (in the error message) in the case of SIGTERM and false positive (in the CO) in the case of an actual bad exit.

>This approach will let CVO know that marketplace is no longer available

How will updating the message inform the CVO that the operator is not available?

kevinrizza: > > After investigating how to report the exit code to the cluster operator I don't believe that we should report the exit code in the cluster operator status.
> 
> Please explain why.
> 
> > Example updated fatal code
> 
> What is the point in just logging? You are still not checking to see what the signal was. Moreover you are giving a false negative (in the error message) in the case of SIGTERM and false positive (in the CO) in the case of an actual bad exit.
> 
> > This approach will let CVO know that marketplace is no longer available
> 
> How will updating the message inform the CVO that the operator is not available?

I'm not sure I totally agree here. It seems like the purpose of sending ClusterOperator status is to describe the current state of the operator. Can we define these terms to maybe better help come up with what is "accurate"? My imagination of these terms:

`Available`: The operator is responding to requests in the controller
`Progressing`: The operator is coming up and is not yet ready to respond to requests
`Degraded`: The operator, due to some present issue, is unable to successfully respond to requests.

I don't think any of these fields describe the state of the operator that is "not running". It seems like the correct solution, in that case, is on exit just mark the cluster status field `Available` to false. I don't think that an exit code of `0` vs an exit code of `1` would result in a different combination here. And, if that is the case, then simply logging the exit code in the marketplace operator should be sufficient here. What value do we get by adding something to the cluster operator message that can't be gained by just looking at the marketplace operator's logs? What is the goal of the cluster status other than a high level aggregation of state?

I'm not sure I see what "false positive" we would be sending using Alex's implementation description above. Is there something I'm mising here?

awgreene: >Please explain why.

Kevin seems to have beaten me to it.

>Moreover you are giving a false negative (in the error message) in the case of SIGTERM

`SIGTERM` wouldn't enter the `if statement` because the fatal function would be called with a `nil` error, so we would update the cluster operator as unavailable and exit with code 0. I tested this by creating a deployment with the changes and killing the pod.

I do not see the false positive case either.

aravindhp: >I do not see the false positive case either.

```
Available = false, "The operator has exited and is no longer reporting status."
Progressing = false, ""
Degraded = false, ""
```
In the case of a real failure shouldn't `degraded` be set to `true`?
Also, is setting all conditions to `false` allowed?

>What value do we get by adding something to the cluster operator message that can't be gained by just looking at the marketplace operator's logs?

We don't want the cluster admin to have to go look at the operator logs to figure out what the problem is in the case of failure. That is exactly what the ClusterOperator CR is for.

awgreene: > In the case of a real failure shouldn't degraded be set to true?

I don't believe this is the case given these descriptions found [here](https://godoc.org/github.com/openshift/api/config/v1#ClusterStatusConditionType)

```
Available indicates that the binary maintained by the operator 
(eg: openshift-apiserver for the openshift-apiserver-operator), is 
functional and available in the cluster.
```
VS.
```
Degraded indicates that the operand is not functioning completely. 
An example of a degraded state would be if there should be 5 copies 
of the operand running but only 4 are running. It may still be available,
but it is degraded
```

> Also, is setting all conditions to false allowed?

Based on the descriptions available in the link  above and the [official ClusterOperator docs](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md) there is no reason why they can't given that they each describe separate states. I can reach out to confirm.

> We don't want the cluster admin to have to go look at the operator logs to figure out what the problem is in the case of failure. That is exactly what the ClusterOperator CR is for.

I agree that this is the intent of the CR. We can add this to the CR message but it will require that we convert the [monitorStopCh](https://github.com/operator-framework/operator-marketplace/blob/master/pkg/status/status.go#L287) to take a variable. This isn't a bad thing - but I had wanted to remove the monitorStopCh and replace it with a single channel that our operator could use to close all child processes. It's a nice to have but not breaking.

awgreene: I also don't believe that asking a cluster administrator to inspect the logs is an unreasonable ask.

aravindhp: > I also don't believe that asking a cluster administrator to inspect the logs of an unavailable operator is an unreasonable ask.

Not if it can be helped. It is just a bad user experience.

kevinrizza: I just think that Available=False describes the state of the cluster. I don't believe that some combination of these states accurately describes the difference between "stopped because we asked nicely" and "stopped rudely". That distinction seems arbitrary from the perspective of cluster status, so why spend significant engineering time trying to solve it? That's my take here.

aravindhp: I am okay with having all three conditions set to false with an appropriate message as long as that's allowed.

awgreene: There seems to be no reason why we cannot after reviewing the following document [1] [2] and checking on Slack to confirm.

[1] https://godoc.org/github.com/openshift/api/config/v1#ClusterStatusConditionType
[2] https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md
[3] https://coreos.slack.com/archives/CEKNRGF25/p1557174280220900

cmd/manager/main.go
```diff
@@ -101,17 +101,23 @@ func main() {
 
 	// Start the Cmd
 	// Update clusteroperator status and log a fatal error
-	fatal(mgr.Start(stopCh), monitorDoneCh, monitorStopCh)
+	exit(mgr.Start(stopCh), monitorDoneCh, monitorStopCh)
 }
 
-// fatal updates clusteroperator status and logs a fatal error
-func fatal(err error, monitorDoneCh <-chan struct{}, monitorStopCh chan struct{}) {
+// exit updates the clusteroperator status and exits with the correct error code.
```
aravindhp: Nit: please be consistent about how you write `clusteroperator`. You say `clusterOperator` in some places and `clusterOperator` in others. Pick one and stick to it everywhere.
PS: I am not saying you need to fix all places in this PR but just don't introduce more inconsistencies.

awgreene: Good catch - I had used `cluster operator` for comments in the past and will update these comments to use the same naming convention.

aravindhp: Please don't use `cluster operator`. It is one word. I would prefer `ClusterOperator` but am okay with `clusteroperator`.

awgreene: Okay - I'll use `ClusterOperator` and create a PR to update the other files to follow that convention.

awgreene:
ci/prow/e2e-aws-upgrade failed due to the following:
```
Installing from initial release registry.svc.ci.openshift.org/ci-op-5s0dj2rj/release@sha256:326980c818b9c252c388f8d70b46ae94911ecde2ddd16ef9b2429d6f38bc61d5
level=warning msg="Found override for ReleaseImage. Please be warned, this is not advised"
level=info msg="Consuming \"Install Config\" from target directory"
level=info msg="Creating infrastructure resources..."
level=info msg="Waiting up to 30m0s for the Kubernetes API at https://api.ci-op-5s0dj2rj-ff81b.origin-ci-int-aws.dev.rhcloud.com:6443..."
level=info msg="API v1.13.4+1ce9c2b up"
level=info msg="Waiting up to 30m0s for bootstrapping to complete..."
level=info msg="Destroying the bootstrap resources..."
level=info msg="Waiting up to 30m0s for the cluster at https://api.ci-op-5s0dj2rj-ff81b.origin-ci-int-aws.dev.rhcloud.com:6443 to initialize..."
level=fatal msg="failed to initialize the cluster: Some cluster operators are still updating: authentication, console: timed out waiting for the condition"
```

/test e2e-aws-upgrade

awgreene:
Same issue as earlier:
```
level=fatal msg="failed to initialize the cluster: Some cluster operators are still updating: authentication, console: timed out waiting for the condition"
```

/test e2e-aws-upgrade

PR #181, opened 2019-05-06
Fix registry pod tolerations

kevinrizza:
Due to a change in the way that OpenShift 4.1 handles taint based evictions, add a set of default tolerations to operator-registry pods.

See https://github.com/operator-framework/operator-marketplace/pull/178 for more details.


pkg/catalogsourceconfig/registry.go
```diff
@@ -298,13 +298,33 @@ func (r *registry) newDeployment(registryCommand []string, needServiceAccount bo
...
 func (r *registry) newPodTemplateSpec(registryCommand []string, needServiceAccount bool) core.PodTemplateSpec {
+	tolerationSecondSpec := int64(120)
 	podTemplateSpec := core.PodTemplateSpec{
 		ObjectMeta: meta.ObjectMeta{
 			Name:      r.csc.GetName(),
 			Namespace: r.csc.GetNamespace(),
 			Labels:    r.getLabel(),
 		},
 		Spec: core.PodSpec{
+			Tolerations: []core.Toleration{
```
aravindhp: Do we need to make this openshift only?

PR #182, opened 2019-05-08
[e2e] OperatorSource with conflicting packages

awgreene:
Introduce a test that ensures that an OperatorSource and its child resources
are successfully rolled out regardless of whether or not there is a
package conflict with an exiting OperatorSource.

### DIFF
--- a/test/testgroups/operatorsourcetests.go
+++ b/test/testgroups/operatorsourcetests.go
@@ -20,7 +20,7 @@ func OperatorSourceTestGroup(t *testing.T) {
 	require.NoError(t, err, "Could not get namespace")
 
 	// Create the OperatorSource.
-	err = helpers.CreateRuntimeObject(test.Global.Client, ctx, helpers.CreateOperatorSourceDefinition(namespace))
+	err = helpers.CreateRuntimeObject(test.Global.Client, ctx, helpers.CreateOperatorSourceDefinition(helpers.TestOperatorSourceName, namespace))
 	require.NoError(t, err, "Could not create OperatorSource")
 
 	// Run the test suites.
--- a/test/testgroups/operatorsourcetests.go
+++ b/test/testgroups/operatorsourcetests.go
@@ -26,5 +26,5 @@ func OperatorSourceTestGroup(t *testing.T) {
 	// Run the test suites.
 	t.Run("opsrc-creation-test-suite", testsuites.OpSrcCreation)
 	t.Run("csc-target-namespace-test-suite", testsuites.CscTargetNamespace)
-	t.Run("csc-packages-test-suite", testsuites.TestCscWithNonExistingPackage)
+	t.Run("packages-test-suite", testsuites.PackageTests)
 }
--- a/test/testsuites/packagestests.go
+++ b/test/testsuites/packagestests.go
@@ -13,10 +13,16 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// TestCscWithNonExistingPackage tests that a csc with a non-existing package
+// PackageTests is a test suite that ensures that package behave as intended
+func PackageTests(t *testing.T) {
+	t.Run("csc-with-non-existing-package", testCscWithNonExistingPackage)
+	t.Run("opsrc-with-conflicting-packages", testOpSrcWithConflictingPackages)
+}
+
+// testCscWithNonExistingPackage tests that a csc with a non-existing package
 // is handled correctly by the marketplace operator and its child resources are not
 // created.
-func TestCscWithNonExistingPackage(t *testing.T) {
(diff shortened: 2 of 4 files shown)


test/testsuites/packagestests.go
```diff
@@ -82,3 +89,33 @@ func childResourcesNotCreated(t *testing.T) {
...
+	// Get test namespace.
+	namespace, err := ctx.GetNamespace()
+	if err != nil {
+		t.Errorf("Could not get namespace: %v", err)
+	}
+
+	// Create the OperatorSource.
+	opSrcName := "conflicting-operators"
+	err = helpers.CreateRuntimeObject(test.Global.Client, ctx, helpers.CreateOperatorSourceDefinition(opSrcName, namespace))
+	if err != nil {
```
anik120: @awgreene Please use `assert`/`require` here and everywhere else you're checking for errors. 

Otherwise looks good to me.

test/testsuites/packagestests.go
```diff
@@ -10,10 +10,17 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// TestCscWithNonExistingPackage tests that a csc with a non-existing package
+// InvalidPackages is a test suite that ensures that package related issues are handled
+// correctly.
+func InvalidPackages(t *testing.T) {
+	t.Run("csc-with-non-existing-package", testCscWithNonExistingPackage)
+	t.Run("opsrc-with-conflicting-packages", testConflictingPackages)
```
aravindhp: I don't see a function called `testConflictingPackages()`. How is this working?

test/testsuites/packagestests.go
```diff
@@ -82,3 +89,33 @@ func childResourcesNotCreated(t *testing.T) {
...
+	// Get global framework variables.
+	client := test.Global.Client
+
+	// Get test namespace.
+	namespace, err := ctx.GetNamespace()
+	if err != nil {
+		t.Errorf("Could not get namespace: %v", err)
+	}
+
+	// Create the OperatorSource.
```
aravindhp: Please call out why creating this OperatorSource will result in two OperatorSources being present on the cluster with conflicting packages.

test/testsuites/packagestests.go
awgreene: Will do!

test/testsuites/packagestests.go
awgreene: Thanks for pointing this out @aravindhp, I had noticed after I made the original PR but decided not to update the PR and kick off an additional test because CI/CD was failing due to high load on the infrastructure at the time. I should have added a comment after making the change.

test/testsuites/packagestests.go
awgreene: Will do!

test/testsuites/packagestests.go
awgreene: I am going to wait until your changes in #177 go through before implementing these changes.

awgreene:
Ready for review @aravindhp @kevinrizza @galletti94 @anik120

test/testgroups/operatorsourcetests.go
```diff
@@ -26,5 +26,5 @@ func OperatorSourceTestGroup(t *testing.T) {
 	// Run the test suites.
 	t.Run("opsrc-creation-test-suite", testsuites.OpSrcCreation)
 	t.Run("csc-target-namespace-test-suite", testsuites.CscTargetNamespace)
-	t.Run("csc-packages-test-suite", testsuites.TestCscWithNonExistingPackage)
+	t.Run("invalid-packages-test-suite", testsuites.InvalidPackages)
```
aravindhp: It would be more appropriate to name this `packages-test-suite` as we could potentially add other tests cases here that deal with valid packages. Or are you thinking of creating a separate test suite in that case?

test/testsuites/packagestests.go
```diff
@@ -13,10 +13,17 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// TestCscWithNonExistingPackage tests that a csc with a non-existing package
+// InvalidPackages is a test suite that ensures that package related issues are handled
```
aravindhp: You should say `invalid packages` if this test suite only deals with invalid package scenarios.

test/testgroups/operatorsourcetests.go
awgreene: I was not planning to create another test suite to verify intended package behavior. I will update the name of this test suite to `packages-test-suite`.

test/testsuites/packagestests.go
awgreene: I will rename the test suite to `Packages`.

PR #184, opened 2019-05-08
[status] Rely on stopCh to stop reporting status 

awgreene:
Problem:
The routine responsible for ClusterOperator status reporting is stopped when the monitorStopCh is closed. The monitorStopCh is only closed when the stopCh is closed and is not used for any other purposes.

Solution:
Rather than creating a channel that is closed when the stopCh channel is closed, we can simply wait for the stopCh channel to close.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -67,24 +67,19 @@ func main() {
 	catalogsourceconfig.InitializeStaticSyncer(mgr.GetClient(), initialWait)
 	registrySyncer := operatorsource.NewRegistrySyncer(mgr.GetClient(), initialWait, resyncInterval, updateNotificationSendWait, catalogsourceconfig.Syncer, catalogsourceconfig.Syncer)
 
-	// monitorStopCh is used to send a signal to stop reporting ClusterOperator status
-	monitorStopCh := make(chan struct{})
-	// monitorDoneCh will recieve a signal when threads have stopped updating ClusterOperator status
-	monitorDoneCh := status.StartReporting(cfg, mgr, namespace, os.Getenv("RELEASE_VERSION"), monitorStopCh)
-
 	// Setup Scheme for all defined resources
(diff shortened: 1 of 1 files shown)


awgreene:
Not for 4.1

/hold

cmd/manager/main.go
```diff
@@ -96,22 +91,23 @@ func main() {
 	log.Print("Starting the Cmd.")
 	stopCh := signals.SetupSignalHandler()
 
+	// monitorDoneCh will recieve a signal when threads have stopped updating ClusterOperator status
```
awgreene: `monitorDoneCh!=statusReportingDoneCh`... will address later.

aravindhp: I don't understand this comment and the naming of the variables. Please elaborate.

awgreene:
@aravindhp 
>I don't understand you commit message. How are you stopping child process with a channel? :-)

Fair - I will update the commit message to better portray how this channel is used to stop the operator from reporting status.

cmd/manager/main.go
awgreene: Sure.

1. I named the channel `statusReportingDoneCh` because the channel is closed when the go routine responsible for reporting `ClusterOperator` Status successfully updates the operator and exits.
2. The comment should be updated to better capture the intent. Perhaps something like:
`// statusReportingDoneCh will be closed after the operator has successfully stopped reporting ClusterOperator status.`

aravindhp: Please reflect this and this PR should be good to be merged. You just need to clearly explain what you are doing now in the code and describe how and why you are moving away from multiple channels in the commit message.

awgreene: Will do!

awgreene:
I have addressed your comments @aravindhp. Please let me know if you would like any additional changes.

awgreene:
Had a merge conflict involving a comment, no changes.

awgreene:
/test e2e-aws
--

PR #186, opened 2019-05-13
[e2e] Check ClusterOperator Status upon initialization

galletti94:
This test checks each of the ClusterOperator Conditions in its Status and
ensures that these match the expected Conditions when the marketplace operator
is available and healthy.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"testing"
 
+	configv1 "github.com/openshift/api/config/v1"
 	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/operator-framework/operator-marketplace/pkg/apis"
 	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
@@ -23,6 +24,7 @@ func TestMarketplace(t *testing.T) {
 	initTestingFramework(t)
 
 	// Run Test Groups
+	t.Run("cluster-operator-status-test-group", testgroups.ClusterOperatorTestGroup)
 	t.Run("operator-source-test-group", testgroups.OperatorSourceTestGroup)
 	t.Run("no-setup-test-group", testgroups.NoSetupTestGroup)
 }
@@ -63,5 +65,17 @@ func initTestingFramework(t *testing.T) {
 	err = test.AddToFrameworkScheme(olm.AddToScheme, catalogSource)
 	if err != nil {
 		t.Fatalf("failed to add CatalogSource custom resource scheme to framework: %v", err)
+	}
+	// Add (configv1) ClusterOperator to framework scheme
+	clusterOperator := &configv1.ClusterOperator{
+		TypeMeta: metav1.TypeMeta{
+			Kind: "ClusterOperator",
+			APIVersion: fmt.Sprintf("%s/%s",
+				configv1.SchemeGroupVersion.Group, configv1.SchemeGroupVersion.Version),
+		},
+	}
+	err = test.AddToFrameworkScheme(configv1.Install, clusterOperator)
+	if err != nil {
+		t.Fatalf("failed to add ClusterOperator custom resource scheme to framework: %v", err)
 	}
 }
--- a/test/testgroups/clusteroperatortests.go
+++ b/test/testgroups/clusteroperatortests.go
@@ -1,0 +1,13 @@
+package testgroups
+
+import (
+	"testing"
+
+	"github.com/operator-framework/operator-marketplace/test/testsuites"
+)
+
+// ClusterOperatorTestGroup runs test suites that check the status of the Cluster Operator
+func ClusterOperatorTestGroup(t *testing.T) {
+	// Run the test suites.
+	t.Run("cluster-operator-status-on-startup-test-suite", testsuites.ClusterOperatorStatusOnStartup)
+}
--- a/test/testsuites/clusteroperatorstatustests.go
+++ b/test/testsuites/clusteroperatorstatustests.go
@@ -1,0 +1,56 @@
+package testsuites
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/operator-framework/operator-sdk/pkg/test"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+// ClusterOperatorStatusOnStartup is a test suite that ensures the ClusterOperator resource which
+// defines the status of the marketplace operator has the correct status upon initialization
+func ClusterOperatorStatusOnStartup(t *testing.T) {
+	ctx := test.NewTestCtx(t)
+	defer ctx.Cleanup()
+
+	// Get global framework variables.
+	client := test.Global.Client
+
+	// Get namespace
+	namespace, err := test.NewTestCtx(t).GetNamespace()
+	require.NoError(t, err, "Could not get namespace")
+
+	// Check that the ClusterOperator resource has the correct status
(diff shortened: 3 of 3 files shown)


test/testgroups/clusteroperatortests.go
```diff
@@ -0,0 +1,13 @@
...
+	"testing"
+
+	"github.com/operator-framework/operator-marketplace/test/testsuites"
+)
+
+// ClusterOperatorTestGroup runs test suites that check the status of the Cluster Operator
+func ClusterOperatorTestGroup(t *testing.T) {
+	// Run the test suites.
+	t.Run("cluster-operator-status-on-startup-test-suite", testsuites.ClusterOperatorStatusOnStartup)
+}
```
aravindhp: Please add new line

test/testsuites/clusteroperatorstatustests.go
```diff
@@ -0,0 +1,57 @@
...
+	ctx := test.NewTestCtx(t)
+	defer ctx.Cleanup()
+
+	// Get global framework variables.
+	client := test.Global.Client
+
+	// Get namespace
+	namespace, err := test.NewTestCtx(t).GetNamespace()
+	if err != nil {
+		t.Fatalf("Could not get namespace: %v", err)
```
aravindhp: Please use `require` or `assert` appropriately as per #177. This comment applies to all places where you are checking for errors.

test/testsuites/clusteroperatorstatustests.go
```diff
@@ -0,0 +1,57 @@
...
+		t.Fatalf("ClusterOperator never reached expected status: %v", err)
+	}
+}
+
+// clusterOperatorCheckTypeStatus checks each condition (type & status) against expectedTypeStatus
+func clusterOperatorCheckTypeStatus(client test.FrameworkClient, clusterOperatorName string, namespace string, expectedTypeStatus map[configv1.ClusterStatusConditionType]configv1.ConditionStatus) error {
+	// Get ClusterOperator resource with clusterOperatorName name
+	namespacedName := types.NamespacedName{Name: clusterOperatorName, Namespace: namespace}
+	result := &configv1.ClusterOperator{}
+	// No need to poll here since the ClusterOperator resource should be there when tests are launched
```
aravindhp: Why shouldn't you poll? It is possible the tests run and for some reason the operator is coming up slowly on the cluster.

test/testsuites/clusteroperatorstatustests.go
```diff
@@ -0,0 +1,57 @@
...
+	// Get namespace
+	namespace, err := test.NewTestCtx(t).GetNamespace()
+	if err != nil {
+		t.Fatalf("Could not get namespace: %v", err)
+	}
+
+	// Check that the ClusterOperator resource has the correct status
+	clusterOperatorName := "marketplace"
+	expectedTypeStatus := map[configv1.ClusterStatusConditionType]configv1.ConditionStatus{
+		"Progressing": "False",
```
aravindhp: Aren't there constants that you can use here from `configv1`?

galletti94: Great point!

test/testsuites/clusteroperatorstatustests.go
galletti94: ok I agree - what timeout length would you advise here?

aravindhp: I would say at least 5-10 minutes. But others should chime in here too.

galletti94: Thanks! @awgreene @kevinrizza @anik120 what do you think?

test/testsuites/clusteroperatorstatustests.go
```diff
@@ -0,0 +1,70 @@
...
+	require.NoError(t, err, "ClusterOperator never reached expected status")
+}
+
+// clusterOperatorCheckTypeStatus checks each condition (type & status) against expectedTypeStatus
+func clusterOperatorCheckTypeStatus(client test.FrameworkClient, clusterOperatorName string, namespace string, expectedTypeStatus map[configv1.ClusterStatusConditionType]configv1.ConditionStatus) error {
+	// Get ClusterOperator resource with clusterOperatorName name
+	namespacedName := types.NamespacedName{Name: clusterOperatorName, Namespace: namespace}
+	result := &configv1.ClusterOperator{}
+
+	// Poll in case the ClusterOperator resource is not immediately present on the cluster
```
aravindhp: It is not enough to poll for the presence. It is possible that the `ClusterOperator` is present but not in the state we expect it to be. That is what you should be polling for.

test/testsuites/clusteroperatorstatustests.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+
+		return true, nil
+	})
+	if err != nil {
+		return err
+	}
+	// Check that each ClusterOperator ConditionType has a ConditionStatus matching expectedTypeStatus
+	for _, condition := range result.Status.Conditions {
+		if expectedTypeStatus[condition.Type] != condition.Status {
```
aravindhp: This should be an `assert`.

test/testsuites/clusteroperatorstatustests.go
```diff
@@ -0,0 +1,70 @@
...
+// clusterOperatorCheckTypeStatus checks each condition (type & status) against expectedTypeStatus
+func clusterOperatorCheckTypeStatus(client test.FrameworkClient, clusterOperatorName string, namespace string, expectedTypeStatus map[configv1.ClusterStatusConditionType]configv1.ConditionStatus) error {
+	// Get ClusterOperator resource with clusterOperatorName name
+	namespacedName := types.NamespacedName{Name: clusterOperatorName, Namespace: namespace}
+	result := &configv1.ClusterOperator{}
+
+	// Poll in case the ClusterOperator resource is not immediately present on the cluster
+	RetryInterval := time.Second * 5
+	Timeout := time.Minute * 5
+	err := wait.Poll(RetryInterval, Timeout, func() (done bool, err error) {
```
kevinrizza: Why can't you just use the helper method here?

`helpers.WaitForResult(client, result, clusterOperatorName, namespace)`

test/testsuites/clusteroperatorstatustests.go
```diff
@@ -0,0 +1,70 @@
...
+		if err != nil {
+			if errors.IsNotFound(err) {
+				return false, nil
+			}
+			return false, err
+		}
+
+		return true, nil
+	})
+	if err != nil {
```
aravindhp: This should be `require`. (I think :-))

test/e2e/main_test.go
```diff
@@ -64,4 +66,16 @@ func initTestingFramework(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to add CatalogSource custom resource scheme to framework: %v", err)
 	}
+	// Add (configv1) ClusterOperator to framework scheme
+	clusterOperator := &configv1.ClusterOperator{
+		TypeMeta: metav1.TypeMeta{
+			Kind: "ClusterOperator",
```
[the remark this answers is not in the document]
galletti94: I get `ClusterOperator` undefined if I remove the quotes. And `configv1` doesn't seem to have `CluterOperatorKind` defined.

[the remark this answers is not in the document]
aravindhp: Duh wrong suggestion on my part. Works only in YAML :-)

test/testsuites/clusteroperatorstatustests.go
```diff
@@ -0,0 +1,55 @@
...
+			return false, err
+		}
+		for _, condition := range result.Status.Conditions {
+			if expectedTypeStatus[condition.Type] != condition.Status {
+				return false, fmt.Errorf("Expecting condition type %v of status %v but got %v", condition.Type, expectedTypeStatus[condition.Type], condition.Status)
+			}
+		}
+		return true, nil
+	})
+	require.NoError(t, err, "ClusterOperator never reached expected status")
```
aravindhp: This should be an `assert`

galletti94:
@aravindhp @kevinrizza I addressed your comments. Please let me know if it lgty

PR #187, opened 2019-05-16
[status] Handle unset RELEASE_VERSION env var

awgreene:
Problem: If the marketplace operator is started in a container where the RELEASE_VERSION environment variable is not set, the ClusterOperator version will be set to an empty string.

Solution: If the RELEASE_VERSION is not set, warn the user by setting the ClusterOperator version to "OpenShift Independant Version".

### DIFF
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -139,6 +139,11 @@ func new(cfg *rest.Config, mgr manager.Manager, namespace string, version string
 		log.Fatal("[status] " + err.Error())
 	}
 
+	// If version is an empty string, warn that the operator is not a part of the OpenShift release payload.
+	if version == "" {
+		version = "OpenShift Independent Version"
+	}
+
 	return &status{
 		configClient:    configClient,
 		coAPINotPresent: coAPINotPresent,
@@ -311,7 +316,7 @@ func (s *status) monitorClusterStatus() {
 			// or if it is the first report.
 			if s.clusterOperator == nil {
 				conditionListBuilder := clusterStatusListBuilder()
-				conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionTrue, fmt.Sprintf("Progressing towards %s", s.version))
+				conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionTrue, fmt.Sprintf("Progressing towards release version: %s", s.version))
 				conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionFalse, "")
 				statusConditions := conditionListBuilder(configv1.OperatorDegraded, configv1.ConditionFalse, "")
 				statusErr = s.setStatus(statusConditions)
@@ -329,7 +334,7 @@ func (s *status) monitorClusterStatus() {
 			if cohelpers.IsStatusConditionFalse(s.clusterOperator.Status.Conditions, configv1.OperatorAvailable) {
 				conditionListBuilder := clusterStatusListBuilder()
 				conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionFalse, "")
-				statusConditions := conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionTrue, fmt.Sprintf("%s is available", s.version))
+				statusConditions := conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionTrue, fmt.Sprintf("Available release version: %s", s.version))
 				statusErr = s.setStatus(statusConditions)
 				break
 			}


pkg/status/status.go
```diff
@@ -329,7 +334,7 @@ func (s *status) monitorClusterStatus() {
 			if cohelpers.IsStatusConditionFalse(s.clusterOperator.Status.Conditions, configv1.OperatorAvailable) {
 				conditionListBuilder := clusterStatusListBuilder()
 				conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionFalse, "")
-				statusConditions := conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionTrue, fmt.Sprintf("%s is available", s.version))
+				statusConditions := conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionTrue, fmt.Sprintf("%s is available", s.releaseVersion))
```
anik120: Looks good to me. 

Just a `nit`: Currently the message would print as `<RELEASE_VERSION unavailable> is available`. Can we think of better phrasing for the message? Something like `Release version: <Unknown RELEASE_VERSION> is available`. In the ideal situation the message would read `Release version: 2.0.5 is available`.

awgreene: I support this change.

pkg/status/status.go
```diff
@@ -55,7 +55,7 @@ type status struct {
 	coAPINotPresent bool
 	namespace       string
 	clusterOperator *configv1.ClusterOperator
-	version         string
+	releaseVersion  string
```
aravindhp: Isn't version here being the release version obvious? Why do you need to call it out and make the code verbose?

pkg/status/status.go
aravindhp: Or we can use the github commit hash and show that as the release version in the case the environment variable is not present. OLM does this or at lease used to do this in the past.

pkg/status/status.go
anik120: I had the same thought at first, but then realized 'releaseVersion` makes the purpose of the variable more clear rather than just `version`. `version` isn't exactly obvious for someone reading the code for the first time and has to dig around a little extra to understand the meaning of the variable. 
Besides, if we have variables names `coAPINotPresent` or `clusterOperator`, `releaseVersion` doesn't seem that verbose in comparison.

aravindhp: Having other verbose variables doesn't mean we can keep adding more. 

What other purpose does version have? If you want to be technically correct then it should be `currentVersion`. I would rather leave it as `version` because it obviously means the version of the operator that you are using.

kevinrizza: > Having other verbose variables doesn't mean we can keep adding more.

+1,000,000

pkg/status/status.go
awgreene: I am not confident that we should set the release version to the marketplace commit hash.

The `RELEASE_VERSION` environment variable should always be set when on a cluster managed with CVO. It represents the version of the cluster and by extent the version of the operator. If the `RELEASE_VARIABLE` environment variable was not set the install/upgrade would fail.

This bug should only occur when the marketplace operator is deployed manually. I believe that having the release version set to the marketplace sha could misrepresent it's intended purpose.

If we want to provide users with a way to identify the version marketplace apart from the CVO Release Version, I believe we should create a new task where we add the commit has of the marketplace operator as a separate key/value pair within versions.

awgreene: @aravindhp ^

pkg/status/status.go
awgreene: I believe that the release version of CVO and the version of marketplace are conceptually different things. Example: CVO could have multiple release version that reference the same version of marketplace.

pkg/status/status.go
aravindhp: This will help if someone is using marketplace in upstream kubernetes cluster and not just from the CLI. I also think it would be immensely useful from a CLI perspective.

pkg/status/status.go
aravindhp: In this context it is the version of marketplace. It just so happens we have decided to use the release version specified by CVO. PS: this is not the release version of CVO but the version specified by it.

pkg/status/status.go
awgreene: @aravindhp could you update this comment based on the off line conversation that we had.

pkg/status/status.go
awgreene: Understood - I will undo the `releaseVersion` changes.

pkg/status/status.go
aravindhp: This should state something to the effect that we are running the operator that is not part of the payload. For upstream we should use the git has to show the version elsewhere as part of an another PR.

awgreene: @aravindhp
>This should state something to the effect that we are running the operator that is not part of the payload.

Thanks!

> For upstream we should use the git has to show the version elsewhere as part of an another PR.

The status routine won't start on upstream.

awgreene: @kevinrizza and I fell on setting the `ClusterOperator` version to `OpenShift Independent Version` if `RELEASE_VERSION` is not set.

aravindhp: >The status routine won't start on upstream.

My statement still stands: " For upstream we should use the git hash to show the version **elsewhere** as part of another PR."

awgreene: Ah - missed that elsewhere, thank you.

awgreene:
Ready for feedback @aravindhp @kevinrizza

PR #190, opened 2019-05-22
[e2e] Update error reporting for e2e pkg tests

galletti94:
This updates the e2e package test based on the new error reporting
format that uses require/assert.

### DIFF
--- a/test/testsuites/cscpackagetests.go
+++ b/test/testsuites/cscpackagetests.go
@@ -4,9 +4,12 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+
 	operator "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
 	"github.com/operator-framework/operator-marketplace/test/helpers"
 	"github.com/operator-framework/operator-sdk/pkg/test"
+	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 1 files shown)


test/testsuites/cscpackagetests.go
```diff
@@ -54,31 +53,24 @@ func TestCscWithNonExistingPackage(t *testing.T) {
...
-		t.Fatalf("Could not get namespace: %v", err)
-	}
+	require.NoError(t, err, "Could not get namespace")
+
 	// Check that the CatalogSourceConfig's child resources were not created.
 	err = helpers.CheckCscChildResourcesDeleted(test.Global.Client, cscName, namespace, namespace)
-	if err != nil {
-		t.Fatal(err)
-	}
+	assert.NoError(t, err, "Child Resources of CatalogSourceConfig were created")
```
anik120: Is it possible to frame this error message in a way that sounds like an error more? I'm thinking if this message shows up during CI testing it might not sound like an error and can get confusing. 
How about `Child resources of CatalogSourceConfig were unexpectedly created`? 
Also `nit`: `resources` with `r`.

aravindhp: This message is a tad misleading. How about `Child resources of the CatalogSourceConfig were not deleted`?

galletti94: The goal of this test is to show that the child resources are not created as a result of creating a csc with a non-existing package - there is no "deletion" happening here.

aravindhp: +1 for @anik120's suggested error message

aravindhp: -1 for my suggestion. Please go with @anik120's suggestion.

galletti94:
@aravindhp @anik120 I updated based on your feedback. Let me know if it lgty

PR #191, opened 2019-05-22
Combine Downloading and Configuring Phases

galletti94:
To solve a race condition bug during upgrade, we want to combine the Downloading and Configuring phase of the operatorsource into a single phase


galletti94:
@aravindhp @kevinrizza @anik120 I have tested the operator upgrade multiple times and have not run into https://bugzilla.redhat.com/show_bug.cgi?id=1695550. Please review when you get a chance

pkg/operatorsource/factory.go
```diff
@@ -59,11 +59,8 @@ func (s *phaseReconcilerFactory) GetPhaseReconciler(logger *log.Entry, opsrc *ma
...
 
-	case phase.OperatorSourceDownloading:
-		return NewDownloadingReconciler(logger, s.registryClientFactory, s.datastore, s.client, s.refresher), nil
-
 	case phase.Configuring:
-		return NewConfiguringReconciler(logger, s.datastore, s.client), nil
+		return NewConfiguringReconciler(logger, s.registryClientFactory, s.datastore, s.client, s.refresher), nil
 
 	case phase.OperatorSourcePurging:
 		return NewPurgingReconciler(logger, s.datastore, s.client), nil
```
kevinrizza: Higher up in this file there is a diagram that refers to the Downloading phase. We should update it to remove that reference.

pkg/operatorsource/validating.go
```diff
@@ -62,6 +62,6 @@ func (r *validatingReconciler) Reconcile(ctx context.Context, in *marketplace.Op
 
 	r.logger.Info("Scheduling for download")
 
-	nextPhase = phase.GetNext(phase.OperatorSourceDownloading)
+	nextPhase = phase.GetNext(phase.Configuring)
 	return
 }
```
kevinrizza: There's a comment at the top of this reconciler that mentions the Downloading phase.

test/testsuites/invalidopsrctests.go
```diff
@@ -65,7 +65,7 @@ func testOpSrcWithInvalidEndpoint(t *testing.T) {
 
 	// Check that OperatorSource is in "Downloading" state with appropriate message
```
kevinrizza: This should say `Check that OperatorSource is in "Configuring" state` instead of `Downloading`

test/testsuites/invalidopsrctests.go
```diff
@@ -65,7 +65,7 @@ func testOpSrcWithInvalidEndpoint(t *testing.T) {
...
 	resultOperatorSource := &operator.OperatorSource{}
-	expectedPhase := "Downloading"
+	expectedPhase := "Configuring"
 	err = wait.Poll(helpers.RetryInterval, helpers.Timeout, func() (bool, error) {
 		err = helpers.WaitForResult(client, resultOperatorSource, namespace, opSrcName)
 		if err != nil {
```
kevinrizza: A comment further up in this file references the Downloading phase

pkg/operatorsource/configuring.go
```diff
@@ -70,12 +76,74 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+	if preUpdateDatastorePackageList == "" {
+		r.logger.Info("New opsrc detected. Refreshing catalogsourceconfigs.")
+		r.refresher.SendRefresh()
+	}
+
+	packages := r.datastore.GetPackageIDsByOperatorSource(out.GetUID())
+	out.Status.Packages = packages
+
+	r.logger.Infof("Successfully stored %d operator manifest(s)", count)
+	r.logger.Info("Download complete, scheduling for configuration")
```
kevinrizza: Is it possible for us to abstract some of this code into functions? As is this method is really long and hard to read.

pkg/operatorsource/appregistryhelper.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
 	"github.com/operator-framework/operator-marketplace/pkg/appregistry"
+	interface_client "github.com/operator-framework/operator-marketplace/pkg/client"
```
aravindhp: If combining the phases fixes the upgrade issue, we would also need to backport #147 to 4.1.z. Is this desired?

pkg/operatorsource/configuring.go
```diff
@@ -16,16 +16,16 @@ import (
...
-// It uses the Client interface which is a wrapper to the raw client
-// provided by the operator-sdk, instead of the raw client itself.
+// NewConfiguringReconcilerWithInterfaceClient returns a configuring
+// Reconciler that reconciles an OperatorSource object in "Configuring"
+// phase. It uses the Client interface which is a wrapper to the raw
+// client provided by the operator-sdk, instead of the raw client itself.
 // Using this interface facilitates mocking of kube client interaction
 // with the cluster, while using fakeclient during unit testing.
-func NewReconcilerWithInterfaceClient(logger *log.Entry, datastore datastore.Writer, client interface_client.Client) Reconciler {
+func NewConfiguringReconcilerWithInterfaceClient(logger *log.Entry, datastore datastore.Writer, client interface_client.Client) Reconciler {
```
aravindhp: Is `New` needed? Can't it be `new` given it is not called from the outside?

pkg/operatorsource/downloading.go
```diff
@@ -15,6 +16,16 @@ import (
...
+	return NewDownloadingReconcilerWithInterfaceClient(logger, factory, datastore, interface_client.NewClient(client), refresher)
+}
+
+// NewDownloadingReconcilerWithInterfaceClient returns a downloading
+// Reconciler that reconciles an OperatorSource object in "Downloading"
+// phase. It uses the Client interface which is a wrapper to the raw
+// client provided by the operator-sdk, instead of the raw client itself.
+// Using this interface facilitates mocking of kube client interaction
+// with the cluster, while using fakeclient during unit testing.
+func NewDownloadingReconcilerWithInterfaceClient(logger *log.Entry, factory appregistry.ClientFactory, datastore datastore.Writer, client interface_client.Client, refresher PackageRefreshNotificationSender) Reconciler {
```
aravindhp: Is `New` needed? Can't it be `new` given it is not called from the outside?

pkg/operatorsource/configuring.go
```diff
@@ -70,12 +76,74 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+
+		// Moving it to 'Failed' phase since human intervention is required to
+		// resolve this situation. As soon as the user pushes new operator
+		// manifest(s) registry sync will detect a new release and will trigger
+		// a new reconciliation.
+		nextPhase = phase.GetNextWithMessage(phase.Failed, err.Error())
+		return
+	}
+
+	r.logger.Infof("Downloaded %d manifest(s) from the operator source endpoint", len(manifests))
```
aravindhp: I suggest we change this message to reflect that these manifests are scheduled to be downloaded by the `operator-registry` pod.

pkg/operatorsource/configuring.go
aravindhp: +1

pkg/operatorsource/configuring_test.go
```diff
@@ -9,12 +9,108 @@ import (
...
 	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	"github.com/operator-framework/operator-marketplace/pkg/appregistry"
+	"github.com/operator-framework/operator-marketplace/pkg/datastore"
 	mocks "github.com/operator-framework/operator-marketplace/pkg/mocks/operatorsource_mocks"
 	"github.com/operator-framework/operator-marketplace/pkg/operatorsource"
 	"github.com/operator-framework/operator-marketplace/pkg/phase"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 )
 
+func TestReconcile_ScheduledForDownload_Success(t *testing.T) {
```
aravindhp: I realize the original did not have a doc. But it would help if a doc was added here describing the test case. I also think the name should be changed.

pkg/operatorsource/configuring.go
anik120: +1

pkg/operatorsource/appregistryhelper.go
galletti94: @kevinrizza any thoughts?

galletti94: @aravindhp thanks for noticing this - it completely escaped me! Will it be harder/riskier to get this PR in 4.1.z if it includes #147 ? I am ok with applying the changes for combining the phases on top of the 4.1 changes and re-writing the unit tests to minimize potential bugs and accelerate the merge process. I can get that done this afternoon. What do you think?

pkg/operatorsource/configuring.go
galletti94: Don't we need this for the call in `configuring_test.go`?

aravindhp: Ah yes, I forgot about the unit tests.

pkg/operatorsource/appregistryhelper.go
aravindhp: I was looking at the code and given #147 has changed configuring.go it will be hard to separate the two. We might need to fix this separately for 4.1.z if needed. Lets leave it as it is for now.

pkg/operatorsource/configuring.go
```diff
@@ -70,12 +76,74 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+
+	// Before we write to the datastore, lets check to see if there are any packages.
```
anik120: nit: `Lets` is the third-person singular form of the verb `Let`. Maybe `Before we write to the datastore, check to see if there are any packages` is enough?

pkg/operatorsource/configuring.go
```diff
@@ -70,12 +76,74 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+	// If there are not, we are assuming this is a new operator source. We should force
+	// all catalog source configs to compare their versions to what's in the datastore
```
anik120: If this is the `CatalogSourceConfig` object, shouldn't it be a single word?

pkg/operatorsource/configuring.go
```diff
@@ -70,12 +76,74 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+			nextPhase = phase.GetNextWithMessage(phase.Failed, err.Error())
+			return
+		}
+
+		r.logger.Infof("There were some faulty operator manifest(s), errors - %v", err)
+		err = nil
+	}
+
+	// Now that we have updated the datastore, lets check if the opsrc is new.
+	// If it is, lets for a resync for catalogsourceconfigs.
```
anik120: Similar `nit` about `lets` vs `let's`

Also, is this suppose to be `force a resync`?

pkg/operatorsource/configuring.go
kevinrizza: `Lets` is incorrect here, but `Let's` (the contraction for `Let us`) seems appropriate.

pkg/operatorsource/configuring.go
aravindhp: Let us be consistent and use `CatalogSourceConfig` in proper case everywhere

pkg/operatorsource/configuring.go
```diff
@@ -70,12 +76,43 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+	if err != nil {
+		// No operator manifest was written, move to Failed phase.
+		nextPhase = phase.GetNextWithMessage(phase.Failed, err.Error())
+	}
+
+	packages := r.datastore.GetPackageIDsByOperatorSource(out.GetUID())
+	out.Status.Packages = packages
+
+	r.logger.Infof("Successfully stored %d operator manifest(s)", count)
+	r.logger.Info("Download complete, scheduling for configuration")
```
aravindhp: These two log messages are inaccurate. Please fix.

pkg/operatorsource/configuring.go
```diff
@@ -70,12 +76,43 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+
+		// Moving it to 'Failed' phase since human intervention is required to
+		// resolve this situation. As soon as the user pushes new operator
+		// manifest(s) registry sync will detect a new release and will trigger
+		// a new reconciliation.
+		nextPhase = phase.GetNextWithMessage(phase.Failed, err.Error())
+		return
+	}
+
+	r.logger.Infof("%d manifest(s) scheduled for download by the operator-registry pod from the operator source endpoint", len(manifests))
```
aravindhp: `%d manifest(s) scheduled for download in the operator-registry pod` will suffice

pkg/operatorsource/configuring.go
```diff
@@ -70,12 +76,43 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
 
 	out = in
 
-	manifests := r.datastore.GetPackageIDsByOperatorSource(in.GetUID())
+	r.logger.Infof("Downloading from [%s]", in.Spec.Endpoint)
```
aravindhp: It will help if you mention the Quay namespace in addition to the endpoint.

pkg/operatorsource/configuring.go
```diff
@@ -95,20 +132,86 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
 		return
 	}
 
+	r.logger.Info("CatalogSourceConfig object has been updated successfully")
+
+	nextPhase = phase.GetNext(phase.Succeeded)
+	return
+}
+
+// downloadManifests abstracts away the download of package metadata from the operatorsource endpoint
```
aravindhp: - You are not downloading the manifests here.
- Let us be consistent and use `OperatorSource`.
- You don't need to say `abstract away`. Rather state what the function is doing.

pkg/operatorsource/configuring.go
```diff
@@ -95,20 +132,86 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+	manifests, err = registry.ListPackages(spec.RegistryNamespace)
+	if err != nil {
+		return manifests, err
+	}
+
+	return manifests, nil
+}
+
+// checkDatastoreExistingPackages checks to see if there are any packages before we write to the datastore
+// If there are not, we are assuming this is a new operator source. We should force all catalogsourceconfigs
```
aravindhp: Please us `OperatorSource` and `CatalogSourceConfig`. Comment applies to all places.

pkg/operatorsource/configuring.go
```diff
@@ -95,20 +132,86 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+// to compare their versions to what's in the datastore after we update it.
+// the function returns the count of the packages in the datastore and an error
+func (r *configuringReconciler) checkDatastoreExistingPackages(in *marketplace.OperatorSource, out *marketplace.OperatorSource, manifests []*datastore.RegistryMetadata) (int, error) {
```
aravindhp: `checkDatastoreForExistingPackages`

pkg/operatorsource/configuring.go
```diff
@@ -95,20 +132,86 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+	// If it is, let's force a resync for catalogsourceconfigs.
+	if preUpdateDatastorePackageList == "" {
+		r.logger.Info("New opsrc detected. Refreshing catalogsourceconfigs.")
+		r.refresher.SendRefresh()
+	}
+
+	return count, err
+}
+
+// updateExistingCatalogSourceConfig abstracts away the logic of updating an existing catalogsourceconfig
```
aravindhp: Sane comment about `abstracts away`

pkg/operatorsource/factory.go
aravindhp: I don't see this update.

pkg/operatorsource/configuring.go
```diff
@@ -70,12 +76,43 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+	r.logger.Infof("Downloading from [%s]", in.Spec.Endpoint)
+
+	manifests, err := r.downloadManifests(&in.Spec, in.Namespace)
+	if err != nil {
+		nextPhase = phase.GetNextWithMessage(phase.Configuring, err.Error())
+		return
+	}
+
+	if len(manifests) == 0 {
+		err = errors.New("The operator source endpoint returned an empty manifest list")
```
aravindhp: `OperatorSource`

pkg/operatorsource/configuring.go
```diff
@@ -95,20 +132,86 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+	if preUpdateDatastorePackageList == "" {
+		r.logger.Info("New opsrc detected. Refreshing catalogsourceconfigs.")
+		r.refresher.SendRefresh()
+	}
+
+	return count, err
+}
+
+// updateExistingCatalogSourceConfig abstracts away the logic of updating an existing catalogsourceconfig
+// when the operatorsource that owns it updates in any way.
```
aravindhp: `is updated in any way.

pkg/operatorsource/configuring_test.go
```diff
@@ -9,12 +9,110 @@ import (
...
+	"github.com/operator-framework/operator-marketplace/pkg/datastore"
 	mocks "github.com/operator-framework/operator-marketplace/pkg/mocks/operatorsource_mocks"
 	"github.com/operator-framework/operator-marketplace/pkg/operatorsource"
 	"github.com/operator-framework/operator-marketplace/pkg/phase"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 )
 
+// Use Case: Registry returns a non-empty list of metadata
+// Expected Result: Next phase is set to "Succeeded"
+func TestReconcile_ScheduledForDownload_Succeeded(t *testing.T) {
```
aravindhp: Shouldn't it be `ScheduledForConfiguring`?

pkg/operatorsource/configuring.go
```diff
@@ -95,20 +132,86 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
 	}
 
 	// If we are here, the given CatalogSourceConfig object already exists.
-	cscNamespacedName := types.NamespacedName{Name: in.Name, Namespace: in.Namespace}
-	cscExisting := marketplace.CatalogSourceConfig{}
-	err = r.client.Get(ctx, cscNamespacedName, &cscExisting)
+	err = r.updateExistingCatalogSourceConfig(ctx, in, packages)
 	if err != nil {
 		r.logger.Errorf("Unexpected error while getting CatalogSourceConfig: %s", err.Error())
```
aravindhp: Error message needs to be updated.

pkg/operatorsource/appregistryhelper.go
kevinrizza: sorry one more nit here: Shouldn't this be camelCase instead of snake_case? i.e. `interfaceClient` ?

galletti94: I just copied what was being used previously but happy to change it across

galletti94: addressed

aravindhp: The `interface_client` usage was introduced in #147. Changing that across the board should not be done in this PR.

pkg/operatorsource/configuring.go
```diff
@@ -95,20 +131,86 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+	if err != nil {
+		if count == 0 {
+			return count, err
+		}
+		// Ignore faulty operator manifests
+		r.logger.Infof("There were some faulty operator manifest(s), errors - %v", err)
+		err = nil
+	}
+
+	// Now that we have updated the datastore, let's check if the opsrc is new.
```
anik120: The function is called `checkDataStoreForExistingPackages`, which it does, but also does the job of forcing a resync of the `CatalogSourceConfig`. Maybe we can separate the two jobs or name the function something else that capture the purpose of the function in it's entirety?

pkg/operatorsource/appregistryhelper.go
```diff
@@ -5,7 +5,7 @@ import (
 
 	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
 	"github.com/operator-framework/operator-marketplace/pkg/appregistry"
-	interface_client "github.com/operator-framework/operator-marketplace/pkg/client"
+	interfaceClient "github.com/operator-framework/operator-marketplace/pkg/client"
```
aravindhp: This change should be made in the previous commit which is where you introduced it.

pkg/operatorsource/appregistryhelper.go
galletti94: ok reverting

pkg/operatorsource/configuring_test.go
```diff
@@ -9,12 +9,110 @@ import (
...
+	opsrcIn := helperNewOperatorSourceWithPhase("marketplace", "foo", phase.Configuring)
+
+	registryClient := mocks.NewAppRegistryClient(controller)
+
+	optionsWant := appregistry.Options{Source: opsrcIn.Spec.Endpoint}
+	factory.EXPECT().New(optionsWant).Return(registryClient, nil).Times(1)
+
+	// We expect the registry to return an empty manifest list.
+	manifests := []*datastore.RegistryMetadata{}
+	registryClient.EXPECT().ListPackages(opsrcIn.Spec.RegistryNamespace).Return(manifests, nil).Times(1)
```
anik120: I don't think the `Times(1)` is not needed. By default `EXPECT()` expects a single call to that function.

pkg/operatorsource/configuring.go
```diff
@@ -15,21 +17,23 @@ import (
...
-// It uses the Client interface which is a wrapper to the raw client
-// provided by the operator-sdk, instead of the raw client itself.
+// NewConfiguringReconcilerWithInterfaceClient returns a configuring
+// Reconciler that reconciles an OperatorSource object in "Configuring"
+// phase. It uses the Client interface which is a wrapper to the raw
+// client provided by the operator-sdk, instead of the raw client itself.
 // Using this interface facilitates mocking of kube client interaction
 // with the cluster, while using fakeclient during unit testing.
-func NewReconcilerWithInterfaceClient(logger *log.Entry, datastore datastore.Writer, client interface_client.Client) Reconciler {
+func NewConfiguringReconcilerWithInterfaceClient(logger *log.Entry, factory appregistry.ClientFactory, datastore datastore.Writer, client interfaceClient.Client, refresher PackageRefreshNotificationSender) Reconciler {
```
anik120: Why is it necessary to rename this function and add the word `Configuring` in it? Isn't that making the function name bigger than needed? Besides, it's an internal function, `NewConfiguringReconciler` is the function that is exposed to the factory.

pkg/operatorsource/configuring.go
```diff
@@ -70,12 +76,42 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+		nextPhase = phase.GetNextWithMessage(phase.Failed, err.Error())
+		return
+	}
+
+	r.logger.Infof("%d manifest(s) scheduled for download in the operator-registry pod", len(manifests))
+
+	count, err := r.checkDatastoreForExistingPackages(in, out, manifests)
+	if err != nil {
+		// No operator manifest was written, move to Failed phase.
+		nextPhase = phase.GetNextWithMessage(phase.Failed, err.Error())
```
aravindhp: Missing a `return` here

pkg/operatorsource/configuring.go
```diff
@@ -95,20 +131,86 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
 	}
 
+	r.logger.Info("CatalogSourceConfig object has been updated successfully")
+
+	nextPhase = phase.GetNext(phase.Succeeded)
+	return
+}
+
+// getManifestMetadata gets the package metadata from the OperatorSource endpoint
+// it returns the list of packages to be written to the OperatorSource status and an error
```
anik120: `And an error` sounds confusing. Is it returning the list of the packages AND and error every time? Or is it returning the list of packages, or an error if something goes wrong, which is the pattern used throughout the code base?

pkg/operatorsource/configuring.go
```diff
@@ -95,20 +131,86 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+func (r *configuringReconciler) getManifestMetadata(spec *marketplace.OperatorSourceSpec, namespace string) ([]*datastore.RegistryMetadata, error) {
+
+	manifests := make([]*datastore.RegistryMetadata, 0)
```
aravindhp: Like I mentioned before, this variable also needs to be renamed

pkg/operatorsource/validating.go
```diff
@@ -62,6 +62,6 @@ func (r *validatingReconciler) Reconcile(ctx context.Context, in *marketplace.Op
 
 	r.logger.Info("Scheduling for download")
```
aravindhp: Please fix

pkg/phase/phase.go
```diff
@@ -31,10 +31,6 @@ const (
 	// In this phase we validate the OperatorSource object.
 	OperatorSourceValidating = "Validating"
 
-	// In this phase, we connect to the specified registry, download available
-	// manifest(s) and save them to the underlying datastore.
```
aravindhp: This doc should be updated and moved to the `Configuring` doc

pkg/operatorsource/configuring_test.go
galletti94: I didn't write this as part of this PR. Not sure this is a refactor that should go in here

pkg/operatorsource/configuring.go
galletti94: It was necessary as part of the refactor in the previous commit to distinguish between the NewReconcilerWithInterfaceClient of the Downloading phase.

pkg/operatorsource/configuring.go
galletti94: The first one. The error will be `nil` if there was no error.

Not sure about your statement that this is a pattern used throughout the code base. How would you return an error `or` a result (both being of different types) in go?

pkg/operatorsource/configuring.go
galletti94: what do we want to call it? `metadata`? `manifestMetadata`?

pkg/operatorsource/configuring.go
galletti94: I see your point but I think it makes sense to have all the syncing done in this function. Maybe @aravindhp or @kevinrizza have an opinion? Or we could change the name to show this is happening here?

pkg/operatorsource/configuring.go
galletti94: @aravindhp or @kevinrizza is there a pattern regarding the returned value/error that we are following for these types of helper methods?

aravindhp: I think @anik120 was asking clarification regarding when an error is returned. 
```
// getManifestMetadata gets the package metadata from the OperatorSource endpoint.
// It returns the list of packages to be written to the OperatorSource status. error is set
// when there is an issue downloading the metadata. In that case the list of packages
// will be empty.
```
Does this make sense? Or are you questioning the pattern itself?

anik120: Yup I was asking for a clearer clarification in the comment about when an error is returned. @aravindhp's snippet provides that clarification.

galletti94: Gotcha thanks for helping clarify!

pkg/operatorsource/configuring_test.go
anik120: Interesting, All of these test cases are being shown as additions in this PR. Maybe @tkashem added the `Times(1)` and can help us understand if my understanding is correct about it.

pkg/operatorsource/configuring.go
aravindhp: `metadata` works for me

pkg/operatorsource/configuring.go
```diff
@@ -95,20 +131,86 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+	}
+
+	return manifests, nil
+}
+
+// checkDatastoreForExistingPackages checks to see if there are any packages before we write to the datastore
+// If there are not, we are assuming this is a new operator source. We should force all CatalogSourceConfig
+// to compare their versions to what's in the datastore after we update it.
+// the function returns the count of the packages in the datastore and an error
+func (r *configuringReconciler) checkDatastoreForExistingPackages(in *marketplace.OperatorSource, out *marketplace.OperatorSource, manifests []*datastore.RegistryMetadata) (int, error) {
```
aravindhp: Please rename `manifests` here too

pkg/operatorsource/configuring.go
aravindhp: I see @anik120's point. This function is actually writing to the datastore and deciding if a sync is needed. We could separate those concerns.

galletti94: That makes sense - thanks for noticing this @anik120 I'll push a change

galletti94:
@aravindhp @kevinrizza @anik120 Updated based on your comments - please let me know if it lgty

pkg/operatorsource/configuring.go
```diff
@@ -70,12 +76,48 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+		// resolve this situation. As soon as the user pushes new operator
+		// manifest(s) registry sync will detect a new release and will trigger
+		// a new reconciliation.
+		nextPhase = phase.GetNextWithMessage(phase.Failed, err.Error())
+		return
+	}
+
+	r.logger.Infof("%d manifest(s) scheduled for download in the operator-registry pod", len(manifests))
+
+	isNeedResync, err := r.writeDatastoreManifests(in, out, manifests)
```
aravindhp: Nit: `isResyncNeeded`

pkg/operatorsource/configuring.go
```diff
@@ -95,20 +137,83 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+
+	metadata, err = registry.ListPackages(spec.RegistryNamespace)
+	if err != nil {
+		return metadata, err
+	}
+
+	return metadata, nil
+}
+
+// writeDatastoreManifests checks to see if there are any manifests
```
aravindhp: Again, we are not writing manifests to the datastore but just the metadata. Please fix all instances of the places where you are using `manifests`.

pkg/operatorsource/configuring.go
```diff
@@ -95,20 +137,83 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+// before we write to the datastore. If there are not, we are assuming this is
+// a new operator source. We should force all CatalogSourceConfig to compare
```
aravindhp: `OperatorSource`

pkg/operatorsource/configuring_test.go
aravindhp: These test cases were moved from `downloading_test.go`

pkg/phase/phase.go
```diff
@@ -13,10 +13,11 @@ const (
 	// attribute is empty.
 	Initial = ""
 
-	// In this phase, for OperatorSource objects we ensure that a corresponding
-	// CatalogSourceConfig object is created. For CatalogSourceConfig objects,
-	// we ensure that an Operator-Registry pod is created and is associated with a
-	// CatalogSource.
+	// In this phase, we connect to the specified registry, download available
+	// manifest(s) and save them to the underlying datastore. For OperatorSource
```
aravindhp: `manifest metadata`

pkg/operatorsource/configuring.go
galletti94: does this include renaming the function to `writeDatastoreMetadata`?

aravindhp: Yes. Suggestion: `writeMetadataToDatastore`

galletti94:
@aravindhp @kevinrizza @anik120 @awgreene I addressed your comments please take another look when you get a chance

PR #195, opened 2019-05-24
[status] Use builder pattern in status

kevinrizza:
Problem:
Currently seeing an issue on some cluster installs where it seems like
the status is not being set properly. This seems to be due to a bug in
the status reporting package where we were relying on a function to
create an object in memory when it was called multiple times, which
is not happening on some environments.

Solution:
Use a builder that creates the object and references it in a specific
memory location.
Additionally, rename the new() function in the status package to
newStatus so that the builtin new is available.
Also fixes a log formatting issue.


kevinrizza:
Closing as this will be addressed in another PR.

PR #196, opened 2019-05-29
[status] Ensure CO status messages are not empty

awgreene:
Problem:
Marketplace recently encountered a bug that was difficult to diagnose
due to ClusterOperator.Status.Messages being set to an empty strings
in multiple location. This bug would have been easier to diagnose
if the messages were set/unique.

Solution:
Make sure that ClusterOperator.Status.Messages are not empty strings.

### DIFF
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -247,15 +247,17 @@ func (s *status) updateStatus(previousStatus *configv1.ClusterOperatorStatus) er
 			}
 		}
 
+		// Log Conditions
+		log.Infof("[status] Attempting to set the ClusterOperator status conditions to:")
+		for _, statusCondition := range s.clusterOperator.Status.Conditions {
+			log.Infof("[status] ConditionType: %v ConditionStatus: ConditionMessage: %v", statusCondition.Type, statusCondition.Status, statusCondition.Message)
+		}
+
 		_, err := s.configClient.ClusterOperators().UpdateStatus(s.clusterOperator)
 		if err != nil {
 			return fmt.Errorf("Error %v updating ClusterOperator", err)
 		}
-		// Log Conditions
-		log.Infof("[status] Current ClusterOperator conditions:")
-		for _, statusCondition := range s.clusterOperator.Status.Conditions {
-			log.Infof("[status] ConditionType: %v\tConditionStatus: %v\tConditionMessage: %v", statusCondition.Type, statusCondition.Status, statusCondition.Message)
-		}
+		log.Info("[status] ClusterOperator status conditions updated.")
 	}
 	return err
 }
@@ -292,10 +294,11 @@ func (s *status) monitorClusterStatus() {
 		case <-s.stopCh:
 			// If the stopCh is closed, the operator will exit and CO should
 			// be set to degraded.
+			msg := "The operator has exited and is no longer reporting status."
 			conditionListBuilder := clusterStatusListBuilder()
-			conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionFalse, "")
-			conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionFalse, "The operator has exited and is no longer reporting status.")
-			statusConditions := conditionListBuilder(configv1.OperatorDegraded, configv1.ConditionFalse, "")
+			conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionFalse, msg)
+			conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionFalse, msg)
+			statusConditions := conditionListBuilder(configv1.OperatorDegraded, configv1.ConditionFalse, msg)
 			statusErr := s.setStatus(statusConditions)
 			if statusErr != nil {
 				log.Error("[status] " + statusErr.Error())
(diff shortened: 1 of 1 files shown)


pkg/status/status.go
```diff
@@ -315,10 +316,11 @@ func (s *status) monitorClusterStatus() {
 			// Create the ClusterOperator in the progressing state if it does not exist
 			// or if it is the first report.
 			if s.clusterOperator == nil {
+				msg := fmt.Sprintf("Progressing towards release version: %s", s.version)
```
aravindhp: This should only be set for the `OperatorProgressing` condition. The other conditions should have some other generic message.

awgreene: Okay, perhaps `OperatorAvailable` and `OperatorDegraded` messages could be:
>"Determining status: a new release version is being deployed."

awgreene: Or better yet:
>"Determining status: The Marketplace operator is upgrading to a new release version."

awgreene: In fact, we may just want to use `Determining status` given that it takes a bit of time to determine if the operator is degraded after the new version is rolled out.

awgreene: We may want to change the status to `ConditionUnknown` as well.

aravindhp: Why would the `status` be `ConditionUnknown`?

awgreene: Marketplace uses`OperatorDegraded` to convey whether or not it is having an issue. I suggest setting the `OperatorDegraded` status to `ConditionUnknown` given that no tests have been ran to validate that the operator is or isn't encountering issues.

awgreene: We could leave the status of `OperatorAvailable` as `false` given that the new version isn't rolled out yet.

aravindhp: Your PR was to ensure the messages were not empty. Lets stick to that in this PR.

awgreene:
> High level thought, can we also just add more verbose info logs so that we know what we are attempting to set in the ClusterOperatorStatus whenever we set it?

I had considered that approach as well but decided not to pursue it. Updating a status suggests that we know enough about the state of an operator to provide an update to an admin. I believe that it makes sense to include a message to tell the admin why were updating the status.

kevinrizza:
> > High level thought, can we also just add more verbose info logs so that we know what we are attempting to set in the ClusterOperatorStatus whenever we set it?
> 
> I had considered that approach as well but decided not to pursue it. Updating a status suggests that we know enough about the state of an operator to provide an update to an admin. I believe that it makes sense to include a message to tell the admin why were updating the status.

But what about the case where we don't update it successfully, or there is something wrong with the update we are trying to make? I think in this case it would be useful, even if it's just in the debug logs.

awgreene:
> > > High level thought, can we also just add more verbose info logs so that we know what we are attempting to set in the ClusterOperatorStatus whenever we set it?
> > 
> > 
> > I had considered that approach as well but decided not to pursue it. Updating a status suggests that we know enough about the state of an operator to provide an update to an admin. I believe that it makes sense to include a message to tell the admin why were updating the status.
> 
> But what about the case where we don't update it successfully, or there is something wrong with the update we are trying to make? I think in this case it would be useful, even if it's just in the debug logs.

This is a good point. How about I create a second commit where the status is logged?

awgreene:
Ready for another review @aravindhp @kevinrizza

pkg/status/status.go
```diff
@@ -247,15 +247,17 @@ func (s *status) updateStatus(previousStatus *configv1.ClusterOperatorStatus) er
...
 
-		_, err := s.configClient.ClusterOperators().UpdateStatus(s.clusterOperator)
-		if err != nil {
-			return fmt.Errorf("Error %v updating ClusterOperator", err)
-		}
 		// Log Conditions
-		log.Infof("[status] Current ClusterOperator conditions:")
+		log.Infof("[status] Attempting to set the ClusterOperator status conditions to:")
 		for _, statusCondition := range s.clusterOperator.Status.Conditions {
 			log.Infof("[status] ConditionType: %v\tConditionStatus: %v\tConditionMessage: %v", statusCondition.Type, statusCondition.Status, statusCondition.Message)
```
aravindhp: While you are doing this, please remove the `\t`. They don't insert tabs in the installer output.

awgreene: Good to know!

awgreene:
Can this get a LGTM @kevinrizza?

pkg/status/status.go
```diff
@@ -247,15 +247,17 @@ func (s *status) updateStatus(previousStatus *configv1.ClusterOperatorStatus) er
 			}
 		}
 
+		// Log Conditions
+		log.Infof("[status] Attempting to set the ClusterOperator status conditions to:")
+		for _, statusCondition := range s.clusterOperator.Status.Conditions {
+			log.Infof("[status] ConditionType: %v ConditionStatus: ConditionMessage: %v", statusCondition.Type, statusCondition.Status, statusCondition.Message)
```
aravindhp: You introduced a bug here that none of us caught :-(. `Infof call needs 2 args but has 3 args` Please fix and also use `go-vet` in VScode.

aravindhp: Or I can fix it in my PR

awgreene: Ack - I'll address it now.

aravindhp: I already fixed it in my PR. Leave it be.

awgreene: Understood, thanks!

PR #198, opened 2019-05-31
OperatorSource configuring reconciler directly creates datastore CatalogSource

galletti94:
TODO:

- [x] Refactor to use interface client in both OperatorSource and CatalogSourceConfig Configuring Reconcilers
- [x] Refactor to move common code to registry package. Also changing method signatures of functions that assume the creation of a CatalogSourceConfig (which is not the case in the new OperatorSource reconciler)
- [x] Remove CatalogSourceConfig creation from OperatorSource configuring reconciler
- [x] Fix unit tests
- [x] Fix e2e tests


pkg/registry/registrydeployer.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	return nil
+}
+
+// EnsurePackagesInStatus makes sure that the csc's status.PackageRepositioryVersions
+// field is updated at the end of the configuring phase if successful. It iterates
+// over the list of packages and creates a new map of PackageName:Version for each
+// package in the spec.
+func (r *RegistryDeployer) EnsurePackagesInStatus(csc *marketplace.CatalogSourceConfig) {
```
aravindhp: This should not be in the `CatalogSourceConfig` package

pkg/registry/registrydeployer.go
```diff
@@ -0,0 +1,139 @@
...
+
+		newPackageRepositioryVersions[packageID] = version
+	}
+
+	csc.Status.PackageRepositioryVersions = newPackageRepositioryVersions
+}
+
+// checkPackages returns an error if there are packages missing from the
+// datastore but listed in the spec.
+func (r *RegistryDeployer) checkPackages(packageIDs []string) error {
```
aravindhp: This should be in the `datastore` package

pkg/registry/registrydestroyer.go
```diff
@@ -0,0 +1,135 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	apps "k8s.io/api/apps/v1"
+	core "k8s.io/api/core/v1"
+	rbac "k8s.io/api/rbac/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+)
+
+// NewRegistryDestroyer returns a new GrpcDestroyer built from the provided parameters.
```
aravindhp: Again this removes `CatalogSource` resources

pkg/operatorsource/configuring.go
```diff
@@ -114,38 +113,18 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
 	packages := r.datastore.GetPackageIDsByOperatorSource(out.GetUID())
 	out.Status.Packages = packages
 
-	cscCreate := new(CatalogSourceConfigBuilder).WithTypeMeta().
-		WithNamespacedName(in.Namespace, in.Name).
-		WithLabels(in.GetLabels()).
-		WithSpec(in.Namespace, packages, in.Spec.DisplayName, in.Spec.Publisher).
-		WithOwnerLabel(in).
-		CatalogSourceConfig()
+	registryDeployer := registry.NewRegistryDeployer(r.logger, r.reader, r.client)
```
aravindhp: Please maintain the same pattern here i.e. creating a `CatalogSource` instead of a `CatalogSourceConfig`.

pkg/operatorsource/configuring.go
```diff
@@ -114,38 +113,18 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
-	if err == nil {
-		nextPhase = phase.GetNext(phase.Succeeded)
-		r.logger.Info("CatalogSourceConfig object has been created successfully")
-
-		return
-	}
-
-	// If we are here, the given CatalogSourceConfig object already exists.
-	err = r.updateExistingCatalogSourceConfig(ctx, in, packages)
+	err = registryDeployer.CreateRegistryResources(key, in.Spec.DisplayName, in.Spec.Publisher, in.Namespace, packages, in.Labels)
```
aravindhp: You are updating an existing `CatalogSource` here.

pkg/operatorsource/deleted.go
```diff
@@ -49,9 +49,11 @@ func (r *deletedReconciler) Reconcile(ctx context.Context, in *marketplace.Opera
 
 	// Delete the operator source manifests.
 	r.datastore.RemoveOperatorSource(out.UID)
+	registryDestroyer := registry.NewRegistryDestroyer(r.logger, r.client)
```
aravindhp: You are removing a `CatalogSource` here

pkg/registry/registrydeployer.go
```diff
@@ -77,26 +76,6 @@ func (r *RegistryDeployer) CreateRegistryResources(key types.NamespacedName, dis
 	return nil
 }
 
-// EnsurePackagesInStatus makes sure that the csc's status.PackageRepositioryVersions
-// field is updated at the end of the configuring phase if successful. It iterates
-// over the list of packages and creates a new map of PackageName:Version for each
-// package in the spec.
-func (r *RegistryDeployer) EnsurePackagesInStatus(csc *marketplace.CatalogSourceConfig) {
```
aravindhp: Not sure why this is being removed in this commit

pkg/catalogsourceconfig/configuring.go
```diff
@@ -5,11 +5,11 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/operator-framework/operator-marketplace/pkg/operatorsource"
-
 	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	interface_client "github.com/operator-framework/operator-marketplace/pkg/client"
```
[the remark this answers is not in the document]
galletti94: I agree with fixing this is a follow up PR. But won't it be a bit more confusing to have only this part using `interface`?

pkg/registry/registrydeployer.go
galletti94: In this commit the `OperatorSource` configuring reconciler is still creating a `catalogsourceconfig` so this had to go in the common package since it was being used by both. In the next commit since `OperatorSource` is no longer using this method, we decided to move it back.

pkg/registry/registrydeployer.go
galletti94: responded here https://github.com/operator-framework/operator-marketplace/pull/198#discussion_r289464571

pkg/catalogsourceconfig/configuring.go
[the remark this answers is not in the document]
aravindhp: I suggested fixing up **other** places in a follow up PR. There are way too many being introduced in this PR to keep holding off.

>But won't it be a bit more confusing to have only this part use interface?

See the usage `interface.Client` vs `interface_client.Client`.

pkg/registry/registrydeployer.go
aravindhp: This sort of code churn is not advisable and makes it confusing. You should try to leave it where it is then.

pkg/catalogsourceconfig/configuring.go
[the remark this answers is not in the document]
galletti94: That's fair - there is an overwhelming majority in this PR and the confusion isn't really an issue. Will fix

pkg/registry/registrydeployer.go
galletti94: But that would mean duplicating it so that it's available in the operatorsource package

pkg/catalogsourceconfig/configuring.go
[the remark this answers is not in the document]
aravindhp: Please make these changes in the commit where you are introducing the change. Not in a separate or in the final commit.

pkg/registry/registrydeployer.go
aravindhp: Then it is pointing at that your abstractions are incorrect. The registry package should not have any `CatalogSourceConfig` specific code.

pkg/catalogsourceconfig/configuring.go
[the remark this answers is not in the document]
galletti94: I can't use `interface` because it's a key word in go. What do you think of `wrapper.Client` or `fake.Client`. @kevinrizza @awgreene @anik120 please let me know your thoughts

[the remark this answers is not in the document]
anik120: Neither `interface` nor `wrapper` or `fake` seem to capture what the interface is for. Wasn't the decision made to use `interfaceClient` instead?

pkg/registry/registrydeployer.go
galletti94: Ok good point. Will revisit this

pkg/catalogsourceconfig/configuring.go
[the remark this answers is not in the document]
galletti94: that was suggested but not decided. `interfaceClient.Client` is a bit of a long and redundant name IMO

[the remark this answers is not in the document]
aravindhp: @anik120 `intefaceClient` was not decided upon. I don't see why `wrapper` doesn't describe what it is. Here is the doc:
```
// Client is a wrapper around the raw kube client provided
// by operator-sdk. Using the wrapper facilitates mocking of client
// interactions with the cluster, while using fakeclient during unit testing.
```
as written by you :-)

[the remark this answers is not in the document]
anik120: Maybe `interfaceClient` was more natural to me since I thought of it that way all the while I was designing it. But if you guys think `wrapper` is natural too then I'm stepping off of this :p

pkg/registry/catalogsourcebuilder.go
```diff
@@ -1,21 +1,10 @@
-package catalogsourceconfig
+package registry
```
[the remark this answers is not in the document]
aravindhp: We might also want to consider using a separate `builders` package given `CatalogSourceBuilder` has nothing to do with the registry.

pkg/registry/catalogsourcebuilder.go
```diff
@@ -73,10 +62,10 @@ func (b *CatalogSourceBuilder) WithOLMLabels(cscLabels map[string]string) *Catal
 }
 
 // WithOwnerLabel sets the owner label of the CatalogSource object to the given owner.
-func (b *CatalogSourceBuilder) WithOwnerLabel(owner *marketplace.CatalogSourceConfig) *CatalogSourceBuilder {
+func (b *CatalogSourceBuilder) WithOwnerLabel(name, namespace string) *CatalogSourceBuilder {
 	labels := map[string]string{
-		CscOwnerNameLabel:      owner.Name,
-		CscOwnerNamespaceLabel: owner.Namespace,
+		RegistryOwnerNameLabel:      name,
+		RegistryOwnerNamespaceLabel: namespace,
```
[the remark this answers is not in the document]
aravindhp: Or even make it `Owner...`

pkg/registry/catalogsourcebuilder.go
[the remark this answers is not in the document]
galletti94: Yeah a separate `builders` package might be cleaner in the long run. Do we want to do this refactor as part of the registry package refactor or in a separate commit?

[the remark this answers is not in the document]
aravindhp: You are moving the files anyhow in this PR. So why not into their own package?

pkg/catalogsourceconfig/configuring.go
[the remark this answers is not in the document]
kevinrizza: ... if we really want to be pedantic ( :) ) I think that the real problem is the word ordering. It's not a client, it's an interface or a wrapper for a type. So it seems to me that the right thing would be something like `clientWrapper`.

pkg/registry/registrydeployer.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	interface_client "github.com/operator-framework/operator-marketplace/pkg/client"
+	"github.com/operator-framework/operator-marketplace/pkg/datastore"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// NewRegistryDeployer returns a new GrpcBuilder.
```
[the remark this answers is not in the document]
galletti94: what do you think of `NewCatalogSourceReconciler`?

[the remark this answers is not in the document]
aravindhp: Sure but it should not live in the `registry` package.

galletti94:
@aravindhp @anik120 @kevinrizza @awgreene Updated based on feedback, please review when you get a chance

pkg/catalogsource/deletion.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// DeletionReconciler encapsulates the logic used to remove all resources related to a registry.
+type DeletionReconciler struct {
+	log    *logrus.Entry
+	client client.Client
+}
+
+// DeleteRegistryResources deletes all resources that make up a registry.
+func (r *DeletionReconciler) DeleteRegistryResources(ctx context.Context, name, namespace, targetNamespace string) (err error) {
```
aravindhp: Same comment as before. You are deleting a `CatalogSource` and all its resources.

pkg/catalogsourceconfig/deleted.go
```diff
@@ -55,7 +50,9 @@ func (r *deletedReconciler) Reconcile(ctx context.Context, in *marketplace.Catal
 	r.cache.Evict(out)
 
 	// Delete all created resources
-	err = r.deleteCreatedResources(ctx, in.Name, in.Namespace, in.Spec.TargetNamespace)
+	registryDestroyer := catalogsource.NewDeletionReconciler(r.logger, r.client)
```
aravindhp: `registryDestroyer`???

pkg/operatorsource/operatorsource.go
```diff
@@ -0,0 +1,16 @@
...
+
+// OpsrcOwnerNameLabel is the label used to mark ownership over resources
+// that are owned by the OperatorSource. When this label is set, the reconciler
+// should handle these resources when the OperatorSource is deleted.
+const OpsrcOwnerNameLabel string = "opsrc-owner-name"
+
+// OpsrcOwnerNamespaceLabel is the label used to mark ownership over resources
+// that are owned by the OperatorSource. When this label is set, the reconciler
+// should handle these resources when the OperatorSource is deleted.
+const OpsrcOwnerNamespaceLabel string = "opsrc-owner-namespace"
```
aravindhp: Are `OpsrcOwnerNameLabel` and `OpsrcOwnerNamespaceLabel` still being used anywhere?

pkg/builders/builders.go
```diff
@@ -0,0 +1,11 @@
+package builders
+
+// OwnerNameLabel is the label used to mark ownership over a given resources.
+// When this label is set, the reconciler should handle these resources when the owner
+// is deleted.
+const OwnerNameLabel string = "owner-name"
```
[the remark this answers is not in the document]
galletti94: Hmm that's a good point. I'll try it out an let you know

pkg/catalogsource/creation.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	return nil
+}
+
+// EnsurePackagesInStatus makes sure that the csc's status.PackageRepositioryVersions
+// field is updated at the end of the configuring phase if successful. It iterates
+// over the list of packages and creates a new map of PackageName:Version for each
+// package in the spec.
+func (r *CreationReconciler) EnsurePackagesInStatus(csc *marketplace.CatalogSourceConfig) {
```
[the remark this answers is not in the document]
galletti94: Yeah this slipped through the cracks this should be in the csc package

pkg/operatorsource/operatorsource.go
galletti94: They are only used in the unit/e2e tests now actually. What was their original purpose?

aravindhp: Same purpose as the other labels. They acted as `OwnerReferences` for objects owned by `OperatorSources`.

aravindhp: That should be happening given how you have abstracted it but it will be using the new label. I think you might have to put different labels depending on if it is a `CatalogSourceConfig` or `OperatorSource` otherwise deletion of the resources won't happen. Unfortunate these labels (except for CatalogSources created in  namespaces other than the watched one) will be going away once we go back to using `OwnerReferences`. Will it make it easier for you if open a PR that does that?

pkg/catalogsource/creation.go
```diff
@@ -0,0 +1,120 @@
...
+
+// CreationReconciler creates the resources required to deploy a registry and enable operators
+// on a cluster managed by OLM.
+type CreationReconciler struct {
+	log    *logrus.Entry
+	reader datastore.Reader
+	client wrapper.Client
+}
+
+// CreateRegistryResources ensures a CatalogSource exists with all the
```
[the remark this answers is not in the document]
galletti94: Maybe I should change the description. This function actually can create a catalogsource and the resources needed.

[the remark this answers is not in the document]
aravindhp: >This function actually can create a catalogsource and the resources needed.

That is what my comment was implying. You need rename the function and update the description.

galletti94:
@aravindhp @kevinrizza @awgreene @anik120 Updated based on your comments - please let me know if it lgty

pkg/catalogsource/creation.go
```diff
@@ -0,0 +1,99 @@
+package catalogsource
```
kevinrizza: Why are you putting this in a separate package from your `builders` package? And why is there a separate file for `creation` and `deletion`?

pkg/catalogsourceconfig/configuring.go
```diff
@@ -63,74 +61,31 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.C
 	// in case of a failure.
 	r.cache.Set(out)
 
-	err = r.reconcileCatalogSource(in)
+	csReconciler := catalogsource.NewResourcesReconciler(r.log, r.reader, r.client)
```
kevinrizza: This is a warping of the `reconciler` pattern. A `reconciler` is a type that takes an object associated with a particular controller and handles it by trying to reconcile between `spec` and `status`. We have reconcilers for objects that our operator handles (one per phase per custom resource). We don't `reconcile` these additional resources, we just create, update, or delete them.

galletti94:
@aravindhp @kevinrizza @anik120 @awgreene I addressed your comments. Please let me know

pkg/builders/catalogsourceconfigbuilder.go
```diff
@@ -1,40 +1,26 @@
...
-
-// OpsrcOwnerNamespaceLabel is the label used to mark ownership over resources
-// that are owned by the OperatorSource. When this label is set, the reconciler
-// should handle these resources when the OperatorSource is deleted.
-const OpsrcOwnerNamespaceLabel string = "opsrc-owner-namespace"
-
 // CatalogSourceConfigBuilder builds a new CatalogSourceConfig type object.
 type CatalogSourceConfigBuilder struct {
-	object marketplace.CatalogSourceConfig
+	Object marketplace.CatalogSourceConfig
```
aravindhp: What is the reason for this change?

pkg/operatorsource/configuring.go
```diff
@@ -115,38 +113,24 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+		Namespace: in.Namespace,
 	}
-
-	if err == nil {
-		nextPhase = phase.GetNext(phase.Succeeded)
-		r.logger.Info("CatalogSourceConfig object has been created successfully")
-
-		return
+	// Get all labels from OperatorSource and add the DatastoreLabel to the map
+	labels := map[string]string{
```
aravindhp: This should be done inside the builder.

pkg/operatorsource/configuring.go
```diff
@@ -115,38 +113,24 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *marketplace.O
...
+		datastore.DatastoreLabel: "true"}
+	for key, value := range in.Labels {
+		labels[key] = value
 	}
-
-	// If we are here, the given CatalogSourceConfig object already exists.
-	err = r.updateExistingCatalogSourceConfig(ctx, in, packages)
+	err = grpcCatalogEnsurer.EnsureResources(key, in.Spec.DisplayName, in.Spec.Publisher, in.Namespace, packages, labels)
```
aravindhp: This call should indicate using a `bool` that you are creating a datastore `CatalogSource` and not pass in labels.

pkg/operatorsource/operatorsource.go
```diff
@@ -0,0 +1,11 @@
...
```
aravindhp: Are OpsrcOwnerNameLabel and OpsrcOwnerNamespaceLabel still being used anywhere?

pkg/client/client.go
```diff
@@ -18,6 +18,8 @@ type Client interface {
 	Create(ctx context.Context, obj runtime.Object) error
 	Get(ctx context.Context, key ObjectKey, objExisting runtime.Object) error
 	Update(ctx context.Context, obj runtime.Object) error
+	Delete(ctx context.Context, obj runtime.Object, opts ...client.DeleteOptionFunc) error
+	List(ctx context.Context, opts *client.ListOptions, list runtime.Object) error
```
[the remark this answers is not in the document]
galletti94: This is a side-effect of needing to use the client interface in both sets of reconcilers. The catalogsourceconfig's delete reconciler needs to list and delete resources. Would it help for me to mention that I updated the client interface in the 1st commit message?

pkg/operatorsource/configuring.go
galletti94: Previously we were passing the entire catalogsourceconfig around in order to propagate labels. Now that we no longer create the in-memory catalogsourceconfig as part of the operatorsource configuring reconciler, we need to pass along these labels.

pkg/operatorsource/configuring.go
galletti94: From https://github.com/operator-framework/operator-marketplace/pull/198#discussion_r290855167 since we need to pass in labels, we can just append to it the datastore label that needs to be propagated.

pkg/operatorsource/operatorsource.go
galletti94: only in the tests afaik

pkg/client/client.go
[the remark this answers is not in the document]
aravindhp: Then shouldn't you be introducing this change in the commit where you start using the two new additions?

pkg/operatorsource/operatorsource.go
aravindhp: Then there is an issue that is not being uncovered. There should be no constants in the main code base that is only being used by the test code.

kevinrizza: +1

pkg/client/client.go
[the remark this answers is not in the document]
galletti94: These are used in this commit. Because we pass in this client interface in each of these methods these new Delete and LIst methods will be used [here](https://github.com/operator-framework/operator-marketplace/pull/198/commits/d59caf31fca43b6ee33649a8da90d55e69a4f615#diff-b0cc6bb0c0e2413e3d6903a35fca00f4R116) and [here](https://github.com/operator-framework/operator-marketplace/pull/198/commits/d59caf31fca43b6ee33649a8da90d55e69a4f615#diff-b0cc6bb0c0e2413e3d6903a35fca00f4R109) for example. The code that uses them didn't need to be updated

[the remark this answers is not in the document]
aravindhp: Ah I see. Please mention the reason for doing so in the commit message.

pkg/builders/catalogsourceconfigbuilder.go
galletti94: Otherwise we cannot set the object here https://github.com/operator-framework/operator-marketplace/pull/198/files/d59caf31fca43b6ee33649a8da90d55e69a4f615..c8dabbe50a7b43dcef2e54d7bf3792fbe1daeb61#diff-20a79a6a9296f6b57de79782f072fbb6R216. Previously since the catalogsourceconfigbuilder was in the same package as the operatorsource this wasn't an issue.

aravindhp: Please mention that in the commit message

pkg/operatorsource/operatorsource.go
galletti94: Agreed. Do we want to remove the part of the test that checks for these labels? Or do we want to just move these constants to the testing package?

aravindhp: We definitely don't want to move these constants to the testing package. That would just be moving the problem. You need to figure out what these constants were used for in the first place and how you have modified or maintaining that behavior.

galletti94: Because the catalogsourceconfig and operatorsource both delete the same child resources, the oprscOwnerLabel is no longer used to find child resources during deletion. These labels were used in an older version of the tests where we cared about this particular label. Now this test just checks that any label will be correctly propagated. During upgrade, because we want to delete the orphaned catalogsourceconfigs, we will need that label to identify which catalogsourceconfigs to delete so I will leave it in the builders package (this file `operatorsource.go` was created by mistake)

galletti94:
@aravindhp @kevinrizza @awgreene @anik120 I addressed your comments. Please take a look when you have time

pkg/registry/registry.go
```diff
@@ -56,13 +58,16 @@ type Registry interface {
...
-		log:    log,
-		client: client,
-		reader: reader,
-		csc:    catalogSourceConfigWrapper{csc},
-		image:  image,
+		log:       log,
+		client:    client,
+		reader:    reader,
+		name:      key.Name,
+		namespace: key.Namespace,
```
aravindhp: `name` and `namespace` is redundant here. Please remove.

pkg/grpccatalog/grpccatalog.go
```diff
@@ -0,0 +1,215 @@
...
+	log    *logrus.Entry
+	reader datastore.Reader
+	client wrapper.Client
+}
+
+// EnsureResources creates a GRPC CatalogSource if one does not already
+// exists otherwise it updates an existing one. It then creates/updates all the
+// resources it requires.
+func (r *GrpcCatalog) EnsureResources(key types.NamespacedName, displayName, publisher, targetNamespace, packages string, labels map[string]string) error {
+	// Ensure that the packages in the spec are available in the datastore
```
aravindhp: You need a guard here that checks if `reader` is not nil

galletti94:
@aravindhp I addressed your comments

pkg/operatorsource/configuring_test.go
```diff
@@ -298,11 +324,15 @@ func TestReconcile_UpdateError_MovedToFailedPhase(t *testing.T) {
...
+
+	// Then we expect to read the packages
+	reader.EXPECT().CheckPackages(gomock.Any()).Return(nil)
 
-	createErr := k8s_errors.NewAlreadyExists(schema.GroupResource{}, "CatalogSourceConfig already exists")
+	// Then we expect a read to the datastore
+	reader.EXPECT().Read(gomock.Any()).Return(&datastore.OpsrcRef{}, nil).AnyTimes()
 
-	kubeclient.EXPECT().Create(context.TODO(), gomock.Any()).Return(createErr)
+	// createErr := k8s_errors.NewAlreadyExists(schema.GroupResource{}, "CatalogSourceConfig already exists")
```
aravindhp: Please remove instead of commenting the line out if it is not required

pkg/operatorsource/configuring_test.go
```diff
@@ -238,10 +257,16 @@ func TestReconcile_CatalogSourceConfigAlreadyExists_Updated(t *testing.T) {
...
-	csc := helperNewCatalogSourceConfigWithLabels(namespace, name, labelsWant)
-	fakeclient := NewFakeClientWithCSC(csc)
+	// Then we expect to read the packages
+	reader.EXPECT().CheckPackages(gomock.Any()).Return(nil)
+
+	// Then we expect a read to the datastore
+	reader.EXPECT().Read(gomock.Any()).Return(&datastore.OpsrcRef{}, nil).AnyTimes()
 
-	reconciler := operatorsource.NewConfiguringReconciler(helperGetContextLogger(), factory, writer, fakeclient, refresher)
+	// csc := helperNewCatalogSourceConfigWithLabels(namespace, name, labelsWant)
```
aravindhp: Please remove if not needed

test/testsuites/invalidopsrctests.go
```diff
@@ -183,4 +191,8 @@ func testOpSrcWithNonexistentRegistryNamespace(t *testing.T) {
 		return false, nil
 	})
 	assert.NoError(t, err, fmt.Sprintf("OperatorSource never reached expected phase/message, expected %v", expectedPhase))
+
+	// // Delete the OperatorSource
+	// err = helpers.DeleteRuntimeObject(client, nonexistentRegistryNamespaceOperatorSource)
+	// require.NoError(t, err, "Could not delete OperatorSource")
```
aravindhp: ???

galletti94:
@aravindhp I rebased on top of master and addressed your comments.

kevinrizza:
@galletti94 I didn't comb over all of our docs, but we should make sure the `Readme` and any other markdown files don't mention `datastore catalogsourceconfigs`

aravindhp:
@galletti94, please test this e2e on an OpenShift cluster from the UI after applying CVO overrides for the operator that uses the image you build from this PR.

aravindhp:
@galletti94 please be aware of https://github.com/operator-framework/operator-lifecycle-manager/pull/883. Make sure changes here doesn't affect that PR or if it does then @alecmerdler needs to be made aware.

galletti94:
Alec is trying to make catalogsources created by users visible in operatorhub and is trying to figure out whether existing labels can be used to filter appropriately in the UI (or if a new label/value needs to be set). I commented to propose a solution but this PR doesn't affect/is not affected by https://github.com/operator-framework/operator-lifecycle-manager/pull/883.

galletti94:
@aravindhp @kevinrizza @awgreene @anik120 Updated according to your feedback and the investigation on the `watch` errors (you'll find a description of the issue in the commit message). Please let me know if it lgty

README.md
```diff
@@ -74,22 +74,7 @@ upstream-community-operators   federationv2,svcat,metering,etcd,prometheus,autom
...
-spec:
-  targetNamespace: local-operators
-  packages: descheduler,jaeger
-  csDisplayName: "Upstream Community Operators"
-  csPublisher: "Red Hat"
-```
-Note, that in the example above, `local-operators` is a namespace that OLM is watching. Deployment of this CR will cause a `CatalogSource` called `installed-upstream-community-operators` to be created in the `local-operators` namespace. This can be confirmed by `oc get catalogsource installed-upstream-community-operators -n local-operators`. Note that you can reuse the same `CatalogSourceConfig` for adding more operators.
-
-Now you can create OLM [`Subscriptions`](https://github.com/operator-framework/operator-lifecycle-manager/tree/274df58592c2ffd1d8ea56156c73c7746f57efc0#discovery-catalogs-and-automated-upgrades) for `desheduler` and `jaeger`.
+Now if you want to install the `descheduler` and `jaeger` operators, create OLM [`Subscriptions`](https://github.com/operator-framework/operator-lifecycle-manager/tree/274df58592c2ffd1d8ea56156c73c7746f57efc0#discovery-catalogs-and-automated-upgrades) for `desheduler` and `jaeger`.
```
aravindhp: - Please update the section about `CatalogSourceConfig`. `CatalogSourceConfig` is used to create `CatalogSource` subsets from an `OperatorSource`. 

Should we call out that `OperatorSources` result in the creation of a `CatalogSource` that can then be used for creating OLM `Subscriptions` in the `OperatorSource` section?

README.md
```diff
@@ -99,11 +84,11 @@ metadata:
...
+Note, that in the example above, `local-operators` is a namespace that OLM is watching. For OLM to act on your subscription please note that an [`OperatorGroup`](https://github.com/operator-framework/operator-lifecycle-manager/blob/274df58592c2ffd1d8ea56156c73c7746f57efc0/Documentation/design/architecture.md#operator-group-design) that matches the [`InstallMode(s)`](https://github.com/operator-framework/operator-lifecycle-manager/blob/274df58592c2ffd1d8ea56156c73c7746f57efc0/Documentation/design/building-your-csv.md#operator-metadata) in your [`CSV`](https://github.com/operator-framework/operator-lifecycle-manager/blob/274df58592c2ffd1d8ea56156c73c7746f57efc0/Documentation/design/building-your-csv.md#what-is-a-cluster-service-version-csv) needs to be present in the subscription namespace (here `local-operators`).
```
aravindhp: I don't think you need to mention `local-operators` is a namespace that OLM is watching but we might want to call out that `marketplace` needs to be a global OLM namespace. Without that the `Subscription` won't get picked up. So we might need to rethink how our stuff is deployed for upstream.

README.md
galletti94: Yes we probably should

galletti94: For your first suggestion, should that not go in #197 ? Or I can change it to be a `CatalogSourceConfig` is used to create combination of `CatalogSources` because there is no subsets restriction yet

README.md
galletti94: Good point. Is there a way to query whether a namespace is a global OLM namespace?

README.md
aravindhp: Sure. It creates a CatalogSource from one or more `OperatorSources` not a combination of `CatalogSources`.

README.md
galletti94: If this is possible, on start up we could check this and just set the status to failing if this isn't the case

aravindhp: Or rather mention that the `Subscription` needs to be in the `marketplace` namespace

galletti94:
@aravindhp @kevinrizza I added some changes to the readme. Please let me know

README.md
```diff
@@ -31,9 +31,11 @@ If you want an `OperatorSource` to work with private app-registry repositories,
 
 On adding an `OperatorSource` to an OKD cluster, operators will be visible in the [OperatorHub UI](https://github.com/openshift/console/tree/master/frontend/public/components/operator-hub) in the OKD console. There is no equivalent UI in the Kubernetes console.
 
+The creation of an `OperatorSource` results in the creation of an OLM `CatalogSource` in the same namespace the marketplace operator is running in. This `CatalogSource` will be populated with operators from the `OperatorSource` ready to be managed by OLM.
+
 #### CatalogSourceConfig
 
-`CatalogSourceConfig` is used to enable an operator present in the `OperatorSource` to your cluster. Behind the scenes, it will configure an OLM `CatalogSource` so that the operator can then be managed by OLM.
+`CatalogSourceConfig` is used to create OLM `CatalogSources` of sets of operators from one or more `OperatorSources` so that these operators can then be managed by OLM.
```
aravindhp: Nit: `consisting of operators`

galletti94: addressed

PR #200, opened 2019-05-31
Add watches for deletion of CatalogSourceConfig child resources

aravindhp:
- Watch all namespaces
Make the operator watch all namespaces but only allow reconciliation of resources in the namespace where it is running. 
- [csc] Add watches for deletion of child resources 
Watch for deletion of child resources and recreate them if they belonged to a CatalogSourceConfig.


aravindhp:
This is an early preview of adding watches for resources created by the operator. Please take a look and provide feedback about the general direction.

pkg/catalogsourceconfig/succeeded.go
```diff
@@ -37,7 +43,24 @@ func (r *succeededReconciler) Reconcile(ctx context.Context, in *marketplace.Cat
...
+		Name:      in.Name,
+		Namespace: in.Spec.TargetNamespace,
+	}
+
+	if err := r.client.Get(context.TODO(), key, catalogSourceGet); err != nil {
+		// The CatalogSource has been deleted. Drop the existing Status field so
+		// that reconciliation can start anew.
+		out.Status = marketplace.CatalogSourceConfigStatus{}
+		nextPhase = phase.GetNext(phase.Configuring)
+		msg = "CatalogSource has been deleted, scheduling for configuring"
```
galletti94: Do we care about the distinction between deleted and not found?

aravindhp: I don't think so as the implication is the same. The resulting configuring shouldn't hurt either.

aravindhp:
@tkashem @awgreene @galletti94 please review

pkg/apis/operators/shared/shared.go
```diff
@@ -35,3 +36,17 @@ func RemoveFinalizer(objectMeta *metav1.ObjectMeta, deletingFinalizer string) {
 
 	return
 }
+
+// IsObjectInOtherNamespace returns true if the namespace is not the watched
+// namespace of the operator
+func IsObjectInOtherNamespace(namespace string) bool {
+	watchNamespace, err := k8sutil.GetWatchNamespace()
+	if err != nil {
+		return false
```
awgreene: Should we highlight in the function documentation that a failure to get get the watch namespace results in a false result?

aravindhp: Sure I can do that.

aravindhp: I am now wondering I should make this function return `bool, error` and return an error in the case where we cannot get the watch namespace as that indicates something going wrong fundamentally. At the call site, we would return `nil, error`. I will let others weigh in here too.

aravindhp: @shawn-hurley @awgreene I have addressed this. PTAL.

PR #202, opened 2019-06-05
[e2e] Modularize waiting for opsrc creation

aravindhp:
Add a helper function that waits for an OperatorSource to be created and reach the expected phase and message.

### DIFF
--- a/test/helpers/helpers.go
+++ b/test/helpers/helpers.go
@@ -2,6 +2,7 @@ package helpers
 
 import (
 	"context"
+	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


test/helpers/helpers.go
```diff
@@ -101,6 +102,27 @@ func WaitForExpectedPhaseAndMessage(client test.FrameworkClient, cscName string,
 	})
 }
 
+// WaitForOpSrcExpectedPhaseAndMessage checks if a OperatorSource with the given name exists in the namespace
```
tkashem: nit: `an OperatorSource`

aravindhp: I have a follow up PR where I am adding e2e tests. I will fix this in that PR. Too many of the tests have passed ;-)

PR #203, opened 2019-06-06
[csc] Update CSC version from v1 to v2

awgreene:


awgreene:
Please take a look @aravindhp @kevinrizza @galletti94

pkg/apis/operators/shared/phase_types.go
```diff
@@ -0,0 +1,60 @@
+package shared
```
aravindhp: I am assuming you did a `git mv` and git decided the file was new. Was that the case?

aravindhp:
This should go in after #198

pkg/apis/operators/shared/phase_types.go
awgreene: That is correct.

awgreene: It decided to create / delete the phase_type after adding the generated deepcopy functions.

kevinrizza:
@awgreene @aravindhp Hey so I think I am slightly out of the loop here, but what is the plan here? Are we just planning on supporting both versions in a cluster? Would it not be better to just migrate any existing objects from V1 to V2 and then not support older versions?

awgreene:
198 has closed, please review @aravindh @kevinrixza

PR #204, opened 2019-06-06
[status] Fixes logging bug

awgreene:
Updates an infof call to expect the correct number of args.


awgreene:
@aravindhp has already introduced a fix.

PR #205, opened 2019-06-07
Bug 1717439: [status] Populate RelatedObjects

aravindhp:
**Problem:**
The must-gather tool is not able to gather enough information about the marketplace operator.
    
**Solution:**
The must-gather tool requires a field, RelatedObjects, in the ClusterOperator CR to be populated with ObjectReferences of the resources associated with the operator. This PR populates this field with the operator's namespace and the default OperatorSource/CatalogSourceConfig/CatalogSource resources.

### DIFF
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -9,9 +9,12 @@ import (
 	configclient "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	cohelpers "github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
 	operatorhelpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v2"
 	log "github.com/sirupsen/logrus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
@@ -161,7 +164,7 @@ func new(cfg *rest.Config, mgr manager.Manager, namespace string, version string
 // cluster
 func (s *status) ensureClusterOperator() error {
 	var err error
-	s.clusterOperator, err = s.configClient.ClusterOperators().Get(clusterOperatorName, v1.GetOptions{})
+	s.clusterOperator, err = s.configClient.ClusterOperators().Get(clusterOperatorName, metav1.GetOptions{})
 
 	if err == nil {
 		log.Debug("[status] Found existing ClusterOperator")
@@ -172,12 +175,15 @@ func (s *status) ensureClusterOperator() error {
 		return fmt.Errorf("Error %v getting ClusterOperator", err)
 	}
 
-	s.clusterOperator, err = s.configClient.ClusterOperators().Create(&configv1.ClusterOperator{
-		ObjectMeta: v1.ObjectMeta{
+	clusterOperator := &configv1.ClusterOperator{
+		ObjectMeta: metav1.ObjectMeta{
 			Name:      clusterOperatorName,
 			Namespace: s.namespace,
 		},
-	})
+	}
+	s.setRelatedObjects()
+
+	s.clusterOperator, err = s.configClient.ClusterOperators().Create(clusterOperator)
 	if err != nil {
 		return fmt.Errorf("Error %v creating ClusterOperator", err)
 	}
@@ -250,8 +256,11 @@ func (s *status) updateStatus(previousStatus *configv1.ClusterOperatorStatus) er
 		// Log Conditions
 		log.Infof("[status] Attempting to set the ClusterOperator status conditions to:")
 		for _, statusCondition := range s.clusterOperator.Status.Conditions {
-			log.Infof("[status] ConditionType: %v ConditionStatus: ConditionMessage: %v", statusCondition.Type, statusCondition.Status, statusCondition.Message)
+			log.Infof("[status] ConditionType: %v ConditionStatus: %v ConditionMessage: %v", statusCondition.Type, statusCondition.Status, statusCondition.Message)
 		}
+
+		// Always update RelatedObjects to account for the upgrade case.
+		s.setRelatedObjects()
 
 		_, err := s.configClient.ClusterOperators().UpdateStatus(s.clusterOperator)
 		if err != nil {
@@ -260,6 +269,36 @@ func (s *status) updateStatus(previousStatus *configv1.ClusterOperatorStatus) er
 		log.Info("[status] ClusterOperator status conditions updated.")
 	}
 	return err
+}
+
+// setRelatedObjects populates RelatedObjects in the ClusterOperator.Status.
+// RelatedObjects are consumed by https://github.com/openshift/must-gather
+func (s *status) setRelatedObjects() {
+	objectReferences := []configv1.ObjectReference{
+		// Add the operator's namespace which will result in core resources
+		// being gathered
+		{
+			Resource: "namespaces",
+			Name:     s.namespace,
+		},
+		// Add the non-core resources we care about
+		{
+			Group:     v1.SchemeGroupVersion.Group,
+			Resource:  v1.OperatorSourceKind,
+			Namespace: s.namespace,
+		},
+		{
+			Group:     v2.SchemeGroupVersion.Group,
+			Resource:  v2.CatalogSourceConfigKind,
+			Namespace: s.namespace,
+		},
+		{
+			Group:     olm.GroupName,
+			Resource:  olm.CatalogSourceKind,
+			Namespace: s.namespace,
+		},
+	}
+	s.clusterOperator.Status.RelatedObjects = objectReferences
 }
 
 // syncChannelReceiver will listen on the sync channel and update the status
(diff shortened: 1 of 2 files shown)


pkg/status/status.go
```diff
@@ -262,6 +269,40 @@ func (s *status) updateStatus(previousStatus *configv1.ClusterOperatorStatus) er
...
+	// Add the operator's namespace which will result in core resources being gathered
+	objectReferences := []configv1.ObjectReference{
+		{
+			Resource: "namespace",
+			Name:     s.namespace,
+		},
+	}
+
+	// Add the default OperatorSources and it's child resources so that their info is gathered too
+	sources := []string{"redhat-operators", "certified-operators", "community-operators"}
```
anik120: Is there a way we can avoid using hard-coded string values here?

pkg/status/status.go
```diff
@@ -178,6 +180,8 @@ func (s *status) ensureClusterOperator() error {
 			Namespace: s.namespace,
 		},
 	})
+	s.setRelatedObjects()
```
anik120: There's a comment [here](https://github.com/operator-framework/operator-marketplace/pull/205/files#diff-a5666326aefdf43a8e89fd4871449fa0R260) that might be a good idea to include above this particular line too for consistency.

pkg/status/status.go
```diff
@@ -262,6 +269,40 @@ func (s *status) updateStatus(previousStatus *configv1.ClusterOperatorStatus) er
...
+		}, configv1.ObjectReference{
+			Group:     olm.GroupName,
+			Resource:  olm.CatalogSourceKind,
+			Namespace: s.namespace,
+			Name:      source,
+		})
+	}
+	s.clusterOperator.Status.RelatedObjects = objectReferences
+}
+
```
kevinrizza: What about deployments, services, roles, etc? Are we not including those here?

pkg/status/status.go
aravindhp: Not easily. One way is to read the `manifest` folder and figure out the default `OperatorSources`. Alternatively we update the `ClusterOperator` CR every time a new marketplace CR is added. I prefer not going down either of these paths as this is a candidate for 4.1.z.

pkg/status/status.go
aravindhp: Reason to add a comment should be to explain something that is not obvious and not for consistency. The reason for the comment in the other place is to convey to the user why we are always updating `RelatedObjects` even though it is static.

pkg/status/status.go
aravindhp: We don't add `Services` and `Roles` for non-private `OperatorSources`. I thought I saw `Deployments` being collected as part of specifying the namespace. Let me check and if it is not I will add that too.

aravindhp:
> @aravindhp Also a nit:
> 
> `Problem: The support is not able to gather enough information about the marketplace operator using the must-gather tool.`
> 
> I think this should either say `Support is not able to` or `The support team is not able to`
> 
> I'm not sure which you meant here.

@anik120, @kevinrizza I will update the commit message to not reference support. It will be better to just refer to the `must-gather` tool.

pkg/status/status.go
aravindhp: `Deployments` are collected by default. Here is an example of the data that is collected:
```
must-gather/
├── cluster-scoped-resources
│   ├── config.openshift.io
│   │   ├── apiservers.yaml
│   │   ├── authentications.yaml
│   │   ├── builds.yaml
│   │   ├── clusteroperators
│   │   │   └── marketplace.yaml
│   │   ├── clusteroperators.yaml
│   │   ├── clusterversions.yaml
│   │   ├── consoles.yaml
│   │   ├── dnses.yaml
│   │   ├── featuregates.yaml
│   │   ├── images.yaml
│   │   ├── infrastructures.yaml
│   │   ├── ingresses.yaml
│   │   ├── networks.yaml
│   │   ├── oauths.yaml
│   │   ├── projects.yaml
│   │   ├── schedulers.yaml
│   │   └── support.yaml
│   └── operator.openshift.io
└── namespaces
    └── openshift-marketplace
        ├── apps
        │   ├── daemonsets.yaml
        │   ├── deployments.yaml
        │   ├── replicasets.yaml
        │   └── statefulsets.yaml
        ├── apps.openshift.io
        │   └── deploymentconfigs.yaml
        ├── autoscaling
        │   └── horizontalpodautoscalers.yaml
        ├── batch
        │   ├── cronjobs.yaml
        │   └── jobs.yaml
        ├── build.openshift.io
        │   ├── buildconfigs.yaml
        │   └── builds.yaml
        ├── core
        │   ├── configmaps.yaml
        │   ├── events.yaml
        │   ├── pods.yaml
        │   ├── replicationcontrollers.yaml
        │   ├── secrets.yaml
        │   └── services.yaml
        ├── image.openshift.io
        │   └── imagestreams.yaml
        ├── openshift-marketplace.yaml
        ├── operators.coreos.com
        │   ├── catalogsourceconfigs
        │   │   ├── certified-operators.yaml
        │   │   ├── community-operators.yaml
        │   │   └── redhat-operators.yaml
        │   ├── catalogsources
        │   │   ├── certified-operators.yaml
        │   │   ├── community-operators.yaml
        │   │   └── redhat-operators.yaml
        │   └── operatorsources
        │       ├── certified-operators.yaml
        │       ├── community-operators.yaml
        │       └── redhat-operators.yaml
        ├── pods
        │   ├── certified-operators-7755565bd7-5hnzn
        │   │   ├── certified-operators
        │   │   │   └── certified-operators
        │   │   │       ├── healthz
        │   │   │       └── logs
        │   │   │           ├── current.log
        │   │   │           └── previous.log
        │   │   └── certified-operators-7755565bd7-5hnzn.yaml
        │   ├── community-operators-c944ffc4f-w9r6l
        │   │   ├── community-operators
        │   │   │   └── community-operators
        │   │   │       ├── healthz
        │   │   │       └── logs
        │   │   │           ├── current.log
        │   │   │           └── previous.log
        │   │   └── community-operators-c944ffc4f-w9r6l.yaml
        │   ├── marketplace-operator-67775478d7-vcjxv
        │   │   ├── marketplace-operator
        │   │   │   └── marketplace-operator
        │   │   │       ├── healthz
        │   │   │       └── logs
        │   │   │           ├── current.log
        │   │   │           └── previous.log
        │   │   └── marketplace-operator-67775478d7-vcjxv.yaml
        │   └── redhat-operators-6f4b66f8d-vdwv9
        │       ├── redhat-operators
        │       │   └── redhat-operators
        │       │       ├── healthz
        │       │       └── logs
        │       │           ├── current.log
        │       │           └── previous.log
        │       └── redhat-operators-6f4b66f8d-vdwv9.yaml
        └── route.openshift.io
            └── routes.yaml
```

`healthz` data is not collected as `must-gather` assumes HTTP liveness while `operator-registry` uses a command.

aravindhp:
@kevinrizza, @anik120, I have updated the commit message. PTAL.

pkg/status/status.go
awgreene: I don't believe that `setRelatedObjects` needs to be called given that it is called in the `updateStatus` function. Am I missing a usecase?

aravindhp: What if `must-gather` is executed in the window before the `updateStatus()` call?

aravindhp: Plus this the place where we are creating the object on a fresh cluster, so it should have all the required fields and not depend on it being done elsewhere.

awgreene: > What if must-gather is executed in the window before the updateStatus() call?
The call to the client to create the object on cluster occurs before the  `setRelatedObjects()` function so this would still be an issue.

> Plus this the place where we are creating the object on a fresh cluster, so it should have all the required fields and not depend on it being done elsewhere.

I believe that we should always ensure that required static fields like this one are set in update calls.

aravindhp: >The call to the client to create the object on cluster occurs before the setRelatedObjects() function so this would still be an issue.

Good point. I will fix that.

>I believe that we should always ensure that required static fields like this one are set in update calls.

And I am doing that. But we should talk more about this when you are doing the refactor.

awgreene: >And I am doing that. But we should talk more about this when you are doing the refactor.

Agreed!

aravindhp:
@awgreene  I have addressed your comments. PTAL.

aravindhp:
Waiting for outcome on https://bugzilla.redhat.com/show_bug.cgi?id=1719454

aravindhp:
Updated to reflect https://github.com/openshift/must-gather/pull/100

@tkashem @kevinrizza @awgreene @galletti94 @anik120 please review again

pkg/status/status.go
```diff
@@ -262,6 +270,36 @@ func (s *status) updateStatus(previousStatus *configv1.ClusterOperatorStatus) er
...
 }
 
+// setRelatedObjects populates RelatedObjects in the ClusterOperator.Status.
+// RelatedObjects are consumed by https://github.com/openshift/must-gather
+func (s *status) setRelatedObjects() {
+	objectReferences := []configv1.ObjectReference{
+		// Add the operator's namespace which will result in core resources
+		// being gathered
+		{
+			Resource: "namespace",
```
tkashem: I had to use `namespaces` to satisfy CI - https://github.com/openshift/origin/blob/master/test/extended/operators/clusteroperators.go#L74

tkashem: @aravindhp do you want to `retest`, otherwise I will put an `lgtm`

aravindhp: Interesting. Moreover I am not sure if the test is correct as `namespace` should also be valid. @sanchezl please weigh in hear. Does it need to `namespaces`?

tkashem: I originally put `namespace` and then ran into the ci error. Also, looks like all `clusteroperator` resources are using `namespaces`.

sanchezl: @aravindhp Yes, the ci test is more restrictive than it should be. The plural name is the "correct" form, but the singular alias (or any other alias) should also work.

aravindhp: No worries. I will make this plural for now. You can update the test cases whenever you get the chance.

aravindhp:
@tkashem @kevinrizza I changed it to `namespaces` and rebased. PTAL.

PR #206, opened 2019-06-14
[docs] Update uninstall flow to exclude csc

galletti94:
We no longer create CatalogSources during install and instead we
reference existing CatalogSources on the cluster. This commit fixes
the uninstall documentation to describe the new uninstall flow where
interacting with a CatalogSourceConfig is no longer needed because none
is created during install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,8 @@ upstream-community-operators   federationv2,svcat,metering,etcd,prometheus,autom
 ```
 **_Note_**: Please do not install [upstream-community-operators] and [community-operators] `OperatorSources` on the same cluster. The rule of thumb is to install [community-operators] on OpenShift clusters and [upstream-community-operators] on upstream Kubernetes clusters.
 
-Now if you want to install the `descheduler` and `jaeger` operators, create OLM [`Subscriptions`](https://github.com/operator-framework/operator-lifecycle-manager/tree/274df58592c2ffd1d8ea56156c73c7746f57efc0#discovery-catalogs-and-automated-upgrades) for `desheduler` and `jaeger`.
+Now if you want to install the `descheduler` and `jaeger` operators, create OLM [`Subscriptions`](https://github.com/operator-framework/operator-lifecycle-manager/tree/274df58592c2ffd1d8ea56156c73c7746f57efc0#discovery-catalogs-and-automated-upgrades) for `desheduler` and `jaeger` in the appropriate namespace. For upstream Kubernetes, this will be `marketplace` (i.e. the same namespace the `CatalogSource` created by the `OperatorSource` is in). This is because `marketplace` is not a global catalog namespace in upstream Kubernetes.
+
 ```
 apiVersion: operators.coreos.com/v1alpha1
 kind: Subscription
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -96,51 +96,18 @@ Note, that in the example above, the subscription must be in the `marketplace` (
 
 After an operator has been installed, to uninstall the operator you need to delete the following resources. Below we uninstall the `jaeger` operator as an example.
 
-Delete the `Subscription` in the `targetNamespace` that the operator was installed into. If created via the OpenShift OperatorHub UI, it will be named after the operator's packageName.
+Delete the `Subscription` in the namespace that the operator was installed into. For upstream, this is the `marketplace` namespace, for OCP, this is the `openshift-marketplace` namespace. If the install was done via the OpenShift OperatorHub UI, it will be named after the operator's packageName so you can run the following command to delete it from the command line:
```
aravindhp: >If the install was done via the OpenShift OperatorHub UI, it will be named after the operator's packageName so you can run the following command to delete it from the command line:

How will that command work in OpenShift given there is no `local-operators` namespace? And please call it OKD for consistency.

galletti94:
@aravindhp @kevinrizza @awgreene @anik120 Please take another look when you get a chance

README.md
```diff
@@ -96,49 +96,18 @@ Note, that in the example above, the subscription must be in the `marketplace` (
 
 After an operator has been installed, to uninstall the operator you need to delete the following resources. Below we uninstall the `jaeger` operator as an example.
 
-Delete the `Subscription` in the `targetNamespace` that the operator was installed into. If created via the OpenShift OperatorHub UI, it will be named after the operator's packageName.
+Delete the `Subscription` in the namespace that the operator was installed into. For OKD, this is the `marketplace` namespace. Keeping to the above example subscription `jaeger`, we can run the following command to delete it from the command line:
```
aravindhp: For OKD it is not `marketplace` but `openshift-marketplace`. i think the problem might be with the fact that the install example is creating the `Subscription` in the `marketplace` namespace. Now the `Subscription` can be anywhere. So changing that might make things easier.

Also, `Note, that in the example above, the subscription must be in the marketplace namespace` in line 93 no longer holds true, so that should be fixed too.

kevinrizza: That's not right, is it? For both OCP and OKD the namespace should be `openshift-marketplace`. It's just for upstream kube that it should be `marketplace`

kevinrizza: That's not right, is it? For both OCP and OKD the namespace should be `openshift-marketplace`. It's just for upstream kube that it should be `marketplace`

galletti94: Yeah sorry that was a typo on my end. Pushed a fix just now

README.md
```diff
@@ -90,55 +91,26 @@ spec:
...
+Note, for OLM to act on your subscription please note that an [`OperatorGroup`](https://github.com/operator-framework/operator-lifecycle-manager/blob/274df58592c2ffd1d8ea56156c73c7746f57efc0/Documentation/design/architecture.md#operator-group-design) that matches the [`InstallMode(s)`](https://github.com/operator-framework/operator-lifecycle-manager/blob/274df58592c2ffd1d8ea56156c73c7746f57efc0/Documentation/design/building-your-csv.md#operator-metadata) in your [`CSV`](https://github.com/operator-framework/operator-lifecycle-manager/blob/274df58592c2ffd1d8ea56156c73c7746f57efc0/Documentation/design/building-your-csv.md#what-is-a-cluster-service-version-csv) needs to be present in the subscription namespace (here `marketplace`).
```
kevinrizza: Probably no need to use the word `note` twice in that sentence. Maybe just `For OLM to act on your subscription please note that`

aravindhp: Please say `which is marketplace in this example` instead of `(hear marketplace)`

galletti94:
@aravindhp @kevinrizza I addressed your comments. PTAL

PR #207, opened 2019-06-17
[migration] Update Subs, remove stale CSC in live clusters

anik120:
With #198 introducing a change in a way that a datastore CatalogSource
is directly created by an OperatorSource instead of a CatalogSourceConfig,
this PR performs some upgrade logic to update/cleanup existing/stale
resources in existing clusters. Specifically:
* Subscriptions with lables "csc-owner-name" and "csc-owner-namespace" are
updated to contain information about dataStore CatalogSources
* Stale datastore CatalogSourceConfigs created before #198 are deleted.
* Stale installed CatalogSourceConfigs created during operator installation
in old clusters are deleted.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -12,6 +12,7 @@ import (
 	"github.com/operator-framework/operator-marketplace/pkg/apis"
 	"github.com/operator-framework/operator-marketplace/pkg/catalogsourceconfig"
 	"github.com/operator-framework/operator-marketplace/pkg/controller"
+	"github.com/operator-framework/operator-marketplace/pkg/migrator"
 	"github.com/operator-framework/operator-marketplace/pkg/operatorsource"
 	"github.com/operator-framework/operator-marketplace/pkg/registry"
 	"github.com/operator-framework/operator-marketplace/pkg/status"
@@ -19,6 +20,7 @@ import (
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 	log "github.com/sirupsen/logrus"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
@@ -91,6 +93,18 @@ func main() {
 
 	log.Print("Starting the Cmd.")
 	stopCh := signals.SetupSignalHandler()
+
+	// set ClusterOperator status to report Migration
+	status.ReportMigration(cfg, mgr, namespace, os.Getenv("RELEASE_VERSION"), stopCh)
+
+	client, err := client.New(cfg, client.Options{})
+	if err != nil {
+		exit(err)
+	}
+
+	// Perform migration logic to upgrade cluster from 4.1.x to 4.2.0
+	migrator := migrator.NewMigrator(client)
+	migrator.Migrate(namespace)
 
 	// statusReportingDoneCh will be closed after the operator has successfully stopped reporting ClusterOperator status.
 	statusReportingDoneCh := status.StartReporting(cfg, mgr, namespace, os.Getenv("RELEASE_VERSION"), stopCh)
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -22,6 +22,14 @@ rules:
   - create
   - get
   - update
+- apiGroups:
+  - operators.coreos.com
+  resources:
+  - subscriptions
+  verbs:
+  - get
+  - update
+  - list  
 ---
 kind: Role
 apiVersion: rbac.authorization.k8s.io/v1
--- a/pkg/migrator/migrator.go
+++ b/pkg/migrator/migrator.go
@@ -1,0 +1,224 @@
+// Package migrator contains upgrade logic that's needed to upgrade a cluster from
+// openshift 4.1.x to openshift 4.2.0.
+package migrator
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/operator-framework/operator-marketplace/pkg/builders"
+
+	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v2"
+	"github.com/sirupsen/logrus"
+	log "github.com/sirupsen/logrus"
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// NewMigrator returns a migrator that updates/cleans up existing/stale
+// resources, when a Openshift 4.1.x cluster is migrated to version 4.2.x.
+// In an openshift 4.1.x cluster, CatalogSourceConfigs were used in the
+// install flow of an operator, and the Subscriptions created during
+// installing operators were referencing the CatalogSources created by the
+// CatalogSourceConfigs(named Installed CatalogSources). In an openshift 4.2 cluster,
+// the operator installation flow has been changed, and CatalogSourceConfigs
+// are no longer used in the install flow. Instead the CatalogSources are
+// directly created by the OperatorSources(named DataStore CatalogSources).
+// The migrator updates existing Subscriptions that referenced Installed CatalogSources
+// during operator installation in an Openshift 4.1.x cluster, to reference
+// Datastore CatalogSources instead. The migrator also deletes the stale
+// CatalogSourceConfigs that were created during operator installation, and
+// the datastore CatalogSourceConfigs created by OperatorSources
+// in an Openshift 4.1.x cluster.
+func NewMigrator(client client.Client) migrator {
+	return migrator{
+		logger: log.NewEntry(log.New()),
+		client: client,
+	}
+}
+
+type migrator struct {
+	logger *logrus.Entry
+	client client.Client
+}
+
+// Migrate updates existing Subscriptions, deletes the CSCs installed during
+// operator installation as well the stale datastore CSCs when the cluster is
+// migrating from Openshift 4.1.x to Openshift 4.2.x
+func (m *migrator) Migrate(operatorNamespace string) {
+	installedCscs := m.updateSubscriptions()
+	m.deleteInstalledCscs(installedCscs, operatorNamespace)
+	m.deleteDatastoreCscs(operatorNamespace)
+}
+
+// updateSubscriptions updates the existing Subscriptions'
+// spec.source and spec.sourcenamespace fields. Existing
+// Subscriptions referenced installed CatalogSources, which
+// are updated to reference datastore CatalogSources
+// instead.
+func (m *migrator) updateSubscriptions() []types.NamespacedName {
+	var installedCscs []types.NamespacedName
+	options := &client.ListOptions{}
+	options.SetLabelSelector(fmt.Sprintf(builders.OwnerNameLabel))
+
+	subscriptions := &olm.SubscriptionList{}
+	// Get the list of existing Subscriptions that have the label "csc-owner-name"
+	err := m.client.List(context.TODO(), options, subscriptions)
+	if err != nil {
+		m.logger.Errorf(fmt.Sprintf("Client error: %s", err.Error()))
+		return []types.NamespacedName{}
+	}
+	for _, instance := range subscriptions.Items {
+		installedCscs = append(installedCscs, types.NamespacedName{Name: instance.GetLabels()[builders.OwnerNameLabel], Namespace: instance.GetLabels()[builders.OwnerNamespaceLabel]})
+		// try to infer the datastore CatalogSource from the Subscription
+		datastoreCs, err := findCatalogSource(&instance, m.client)
+		if k8s_errors.IsNotFound(err) {
+			// infer the CatalogSource from the OperatorSource that has the package
+			datastoreCs, err = findCsFromOpsrc(&instance, m.client)
+			if err != nil {
+				m.logger.Errorf("[migration] Could not infer datastore CatalogSource for Subscription %s.", instance.GetName())
+				continue
+			}
+		}
+		// update the Subscription to reference the datastore CatalogSource
+		instance.Spec.CatalogSource = datastoreCs.GetName()
+		instance.Spec.CatalogSourceNamespace = datastoreCs.GetNamespace()
+		err = m.client.Update(context.TODO(), &instance)
+		if err != nil {
+			m.logger.Errorf("[migration] Error updating subscription %s. Error: %s", instance.GetName(), err.Error())
+		} else {
+			m.logger.Infof("[migration] Successfully updated Subscription %s", instance.GetName())
+		}
+	}
+	return installedCscs
+}
+
+// deleteInstalledCscs deletes the CSCs installed during operator installation.
+// The child resources of the CSCs are delete by the finalizer.
+func (m *migrator) deleteInstalledCscs(cscs []types.NamespacedName, operatorNamespace string) {
+	for _, cscInfo := range cscs {
+		// If the CatalogSourceConfig namespace information is missing, try and find the
+		// CatalogSourceConfig in the marketplace-operator's namespace
+		if cscInfo.Namespace == "" {
+			cscInfo.Namespace = operatorNamespace
+		}
+		csc := newCatalogSourceConfig(cscInfo.Namespace, cscInfo.Name)
+		err := m.client.Delete(context.TODO(), csc)
+		if err != nil {
+			m.logger.Errorf("[migration] Failed to delete installed CSC %s with error: ", cscInfo.Name, err.Error())
+		} else {
+			m.logger.Infof("[migration] Stale CSC %s scheduled for deletion.", cscInfo.Name)
+		}
+	}
+}
+
+// deleteDatastoreCscs deletes the datastore CSCs created by OperatorSources.
+// The child resources of the CSCs are deleted by the finalizer.
+func (m *migrator) deleteDatastoreCscs(operatorNamespace string) {
+	options := &client.ListOptions{}
+	options.SetLabelSelector(fmt.Sprintf("opsrc-datastore: \"true\""))
+	options.InNamespace(operatorNamespace)
+	cscs := &v2.CatalogSourceConfigList{}
+	// Get the list of existing cscs that have the label "opsrc-datastore: "true" "
+	err := m.client.List(context.TODO(), options, cscs)
+	if err != nil {
+		m.logger.Errorf("Client error: %s", err.Error())
+		return
+	}
+	for _, csc := range cscs.Items {
+		err = m.client.Delete(context.TODO(), &csc)
+		if err != nil {
+			m.logger.Errorf("[migration] Failed to delete CatalogSourceConfig %s. Error: %s", csc.GetName(), err.Error())
+		} else {
+			m.logger.Infof("[migration] Datastore CSC %s scheduled for deletion.", csc.GetName())
+		}
+	}
+}
+
+// findCatalogSource infers the datastore CatalogSource created by the OperatorSource
+// in an Openshift 4.2.0 cluster. The inferred datastore CatalogSource will then be
+// referenced from an existing Subscription.
+func findCatalogSource(subscription *olm.Subscription, client client.Client) (*olm.CatalogSource, error) {
+	associatedCscName := subscription.GetLabels()[builders.OwnerNameLabel]
+	possibleCsName := ExtractCsName(associatedCscName)
+	possibleCsNamespace := subscription.GetLabels()[builders.OwnerNamespaceLabel]
+	// try and fetch the CatalogSource
+	datastoreCs := &olm.CatalogSource{}
+	namespacedName := types.NamespacedName{Name: possibleCsName, Namespace: possibleCsNamespace}
+	err := client.Get(context.TODO(), namespacedName, datastoreCs)
+	if err != nil {
+		return nil, err
+	}
+	return datastoreCs, nil
+}
+
+// findCsFromOpsrc extracts the packageName from a Subscription,
+// and finds the corresponding CatalogSource that the package belongs to.
+func findCsFromOpsrc(subscription *olm.Subscription, kubeClient client.Client) (*olm.CatalogSource, error) {
+	packageName := subscription.Spec.Package
+	opsrcs := &v1.OperatorSourceList{}
+	err := kubeClient.List(context.TODO(), &client.ListOptions{}, opsrcs)
+	if err != nil {
+		return nil, err
+	}
+	for _, instance := range opsrcs.Items {
+		if !IsPackageInOpsrc(packageName, &instance) {
+			continue
+		}
+		// fetch the CatalogSource with the same name
+		datastoreCs := &olm.CatalogSource{}
+		namespacedName := types.NamespacedName{Name: instance.GetName(), Namespace: instance.GetNamespace()}
+		err = kubeClient.Get(context.TODO(), namespacedName, datastoreCs)
+		if err != nil {
+			return nil, err
+		}
+		return datastoreCs, nil
+	}
+	return nil, nil
+}
+
+// IsPackageInOpsrc takes the name of a package, and an OperatorSource
+// as input, and returns true if the package is present in the OperatorSource's
+// status.Packages field
+func IsPackageInOpsrc(packageName string, opsrc *v1.OperatorSource) bool {
+	packages := opsrc.GetPackages()
+	for _, pkg := range packages {
+		if pkg == packageName {
+			return true
+		}
+	}
+	return false
+}
+
+// ExtractCsName takes the name of a CatalogSourceConfig
+// as input, and infers the name of the corresponding
+// datastore CatalogSource. Installed CSCs follow the
+// following naming pattern: installed-publisher-namespace.
+// For example, for a CatalogSourceConfig named
+// `installed-community-openshift-operators`, extractCsName
+// extracts `community` off the name, appends `operators`
+// to it, and returns `community-operators` as output
+func ExtractCsName(cscName string) string {
+	possibleCsName := strings.Split(cscName, "-")[1]
+	return fmt.Sprintf("%s-%s", possibleCsName, "operators")
+}
+
+// newCatalogSourceConfig returns a newly built CatalogSourceConfig
+func newCatalogSourceConfig(namespace, name string) *v2.CatalogSourceConfig {
+	return &v2.CatalogSourceConfig{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: fmt.Sprintf("%s/%s",
+				v1.SchemeGroupVersion.Group, v1.SchemeGroupVersion.Version),
+			Kind: v2.CatalogSourceConfigKind,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+	}
+}
--- a/test/helpers/helpers.go
+++ b/test/helpers/helpers.go
@@ -2,12 +2,15 @@ package helpers
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"time"
 
 	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
-	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
-	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v2"
+	v1 "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	v2 "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v2"
+	"github.com/operator-framework/operator-marketplace/pkg/builders"
+	"github.com/operator-framework/operator-marketplace/pkg/datastore"
 	"github.com/operator-framework/operator-sdk/pkg/test"
 	apps "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -38,6 +41,19 @@ const (
 	// TestOperatorSourceLabelValue is a label value added to the opeator source returned
 	// by the CreateOperatorSource function.
 	TestOperatorSourceLabelValue string = "Community"
+
+	// TestDatastoreCscName is the name of a CatalogSourceConfig that is returned by
+	// the CreateDatastoreCsc function.
+	TestDatastoreCscName string = "test-operators"
+
+	// TestInstalledCscPublisherName is the publisher name part of a installed CatalogSourceConfig
+	// that is returned by the CreateDatastoreCsc function. This publisher name part should be
+	// apened with a namespace to generate the full installed CatalogSourceConfig name.
+	TestInstalledCscPublisherName string = "installed-test-"
+
+	// TestSubscriptionName is the name of a Subscription that is returned by
+	// the CreateSubscription function.
+	TestSubscriptionName string = "test-operators"
 )
 
 // WaitForResult polls the cluster for a particular resource name and namespace.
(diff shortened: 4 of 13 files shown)


cmd/manager/cleanup.go
```diff
@@ -0,0 +1,178 @@
...
+		// point the subscription to the datastore CatalogSource
+		instance.SetCatalogSource(datastoreCs.GetName())
+		instance.SetCatalogSourceNamespace("openshift-marketplace")
+		kubeClient.Update(context.TODO(), &instance)
+	}
+	return cscNames, nil
+}
+
+func deleteDatastoreCscs(cscNames []string, kubeClient client.Client) error {
+	for _, cscName := range cscNames {
```
aravindhp: Potential for null pointer dereference  given line 73 in `updateSubscriptions`

cmd/manager/cleanup.go
```diff
@@ -0,0 +1,178 @@
...
+			// guess the catalogsource from the opsrc that has the package
+			datastoreCs, err = findCsFromOpsrc(&instance, kubeClient)
+			if err != nil {
+				return nil, err
+			}
+		}
+		// point the subscription to the datastore CatalogSource
+		instance.SetCatalogSource(datastoreCs.GetName())
+		instance.SetCatalogSourceNamespace("openshift-marketplace")
+		kubeClient.Update(context.TODO(), &instance)
```
aravindhp: Please check for error and log it.

cmd/manager/cleanup.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	return cscNames, nil
+}
+
+func deleteDatastoreCscs(cscNames []string, kubeClient client.Client) error {
+	for _, cscName := range cscNames {
+		namespacedName := types.NamespacedName{Name: cscName, Namespace: "openshift-marketplace"}
+		csc := &marketplace.CatalogSourceConfig{}
+		err := kubeClient.Get(context.TODO(), namespacedName, csc)
```
aravindhp: Why do you need to do a `Get` here? Just attempt to delete it and log the error.

cmd/manager/cleanup.go
```diff
@@ -0,0 +1,178 @@
...
+	options.SetLabelSelector(fmt.Sprintf("opsrc-datastore: \"true\""))
+
+	cscs := &marketplace.CatalogSourceConfigList{}
+	// Get the list of existing cscs that have the lable "opsrc-datastore: "true" "
+	err := kubeClient.List(context.TODO(), options, cscs)
+	if err != nil {
+		return err
+	}
+	for _, csc := range cscs.Items {
+		kubeClient.Delete(context.TODO(), &csc)
```
aravindhp: Please check and log errors.

cmd/manager/cleanup.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func findCatalogSource(subscription *olm.Subscription, client client.Client) (*olm.CatalogSource, error) {
+	associatedCscName := subscription.GetLabels()["csc-owner-name"]
+	possibleCsName, err := extractCsName(associatedCscName)
+	if err != nil {
+		return nil, err
+	}
+	// try and fetch the CatalogSource
+	datastoreCs := &olm.CatalogSource{}
```
aravindhp: Why is this a datastore `CatalogSource`?

cmd/manager/cleanup.go
```diff
@@ -0,0 +1,178 @@
...
+			return err
+		}
+
+	}
+	return nil
+}
+
+func deleteOldCscs(kubeClient client.Client) error {
+	options := &client.ListOptions{}
+	options.SetLabelSelector(fmt.Sprintf("opsrc-datastore: \"true\""))
```
aravindhp: This seems to imply that you are looking for and deleting the datastore `CatalogSourceConfigs` here but you call this function `deleteOldCscs`. This is very confusing.

cmd/manager/cleanup.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+		// point the subscription to the datastore CatalogSource
+		instance.SetCatalogSource(datastoreCs.GetName())
+		instance.SetCatalogSourceNamespace("openshift-marketplace")
+		kubeClient.Update(context.TODO(), &instance)
+	}
+	return cscNames, nil
+}
+
+func deleteDatastoreCscs(cscNames []string, kubeClient client.Client) error {
```
aravindhp: Aren't you deleting the `CatalogSourceConfigs` created as part of an operator install here? But you call the function `deleteDatastoreCscs`?

cmd/manager/cleanup.go
```diff
@@ -0,0 +1,178 @@
...
+	datastoreCs := &olm.CatalogSource{}
+	namespacedName := types.NamespacedName{Name: possibleCsName, Namespace: "openshift-marketplace"}
+	err = client.Get(context.TODO(), namespacedName, datastoreCs)
+	if err != nil {
+		return nil, err
+	}
+	return datastoreCs, nil
+}
+
+func findCsFromOpsrc(subscription *olm.Subscription, clt client.Client) (*olm.CatalogSource, error) {
```
aravindhp: Please use `client` instead of `clt`

cmd/manager/cleanup.go
```diff
@@ -0,0 +1,178 @@
...
+	packageName := subscription.GetName()
```
aravindhp: The name of the package is in the spec of the `Subscription`

cmd/manager/cleanup.go
```diff
@@ -0,0 +1,178 @@
...
+	opsrcs := &marketplace.OperatorSourceList{}
+	err := clt.List(context.TODO(), &client.ListOptions{}, opsrcs)
+	if err != nil {
+		return nil, err
+	}
+	for _, instance := range opsrcs.Items {
+		if isPackageInOpsrc(packageName, &instance) {
```
aravindhp: Code is easier to read if you check the negation here. 
```
if !isPackageInOpsrc(packageName, &instance) {
    continue
}
```
This will result in less tabbed code below.

cmd/manager/cleanup.go
```diff
@@ -0,0 +1,178 @@
...
+			// fetch the CatalogSource with the same name
+			datastoreCs := &olm.CatalogSource{}
+			namespacedName := types.NamespacedName{Name: instance.GetName(), Namespace: "openshift-marketplace"}
+			err = clt.Get(context.TODO(), namespacedName, datastoreCs)
```
aravindhp: Why do you need to do a `Get` here?

cmd/manager/cleanup.go
```diff
@@ -0,0 +1,178 @@
...
+	packages := strings.Split(opsrc.GetPackages(), ",")
+	for _, pkg := range packages {
+		if pkg == packageName {
+			return true
+		}
+	}
+	return false
+}
+
+func extractCsName(cscName string) (string, error) {
```
aravindhp: Please document sort of pattern you are looking for here.

cmd/manager/main.go
```diff
@@ -98,6 +99,13 @@ func main() {
 	go registrySyncer.Sync(stopCh)
 	go catalogsourceconfig.Syncer.Sync(stopCh)
 
+	if os.Getenv("RELEASE_VERSION") != "" {
```
aravindhp: Why is this check required?

vendor/github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1/subscription_types.go
```diff
@@ -109,3 +109,13 @@ func NewInstallPlanReference(ref *corev1.ObjectReference) *InstallPlanReference
 		UID:        ref.UID,
 	}
 }
+
+// SetCatalogSource sets the Spec.source field
+func (s *Subscription) SetCatalogSource(csName string) {
```
aravindhp: You cannot modify vendor code like this. This needs a PR against OLM or we carry the code.

cmd/manager/main.go
```diff
@@ -98,6 +99,13 @@ func main() {
...
+		client, err := client.New(config.GetConfigOrDie(), client.Options{})
+		clean := NewCleanUpResources(log.NewEntry(log.New()), client)
```
aravindhp: You need to update the `ClusterOperator` CR during this process to show that we are performing some upgrade logic.

aravindhp:
Please add an e2e for this that asserts that no datastore `CatalogSourceConfigs` are present on the system.

cmd/manager/cleanup.go
```diff
@@ -0,0 +1,178 @@
...
+	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+	marketplace "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	"github.com/sirupsen/logrus"
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// NewCleanUpResources returns a CleanUpCrew that updates existing
+// subscriptions linked to old CSCs and links them to the datastore CSCs
```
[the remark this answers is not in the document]
anik120: @aravindhp Is `updates existing subscriptions pointed to stale CSCs through the field spec.source and points them to datastore CatalogSources` better?

cmd/manager/cleanup.go
```diff
@@ -0,0 +1,178 @@
...
+// in the openshift-marketplace namespace, and then deletes those CSCs
+// along with existing datastore CSCs.
+func NewCleanUpResources(logger *logrus.Entry, client client.Client) CleanUpCrew {
```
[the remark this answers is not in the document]
galletti94: +1 for `Upgrader`

cmd/manager/cleanup.go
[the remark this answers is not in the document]
aravindhp: `Subscriptions` don't point to `CatalogSourceConfigs`. They point to `CatalogSources`. This also gives rise to the question whether we delete these stale `CatalogSources` too.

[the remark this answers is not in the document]
anik120: So the `spec.source` field of a subscription points to a `catalogSource` which has the same name as the CSC that created it? In that case, won't just deleting the CSC delete the `CatalogSource` too as it's one of the CSC's child resource?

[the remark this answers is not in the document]
aravindhp: >So the spec.source field of a subscription points to a catalogSource which has the same name as the CSC that created it?

This is true for the `datastore` `CatalogSourceConfigs` but I am not sure about the ones that are created as part of the old operator install flow.

>In that case, won't just deleting the CSC delete the CatalogSource too as it's one of the CSC's child resource?

Yeah I think the finalizers should handle that.

cmd/manager/main.go
anik120: @aravindhp This is to make sure this code runs only on an OKD cluster. I noticed we set this env variable for OKD clusters and not when the operator is installed in a plain kube cluster. 
Sort of a hacky way, but the code assumes a lot of the resources are in the `openshift-marketplace` namespace. 
Although, based on your comment above about making a best effort to delete resources instead of erroring out and exiting, we could get rid of this check. The log will just look messy in a plain kube cluster.

aravindhp: We should do the same steps we are performing on an OKD cluster on a plain kuber cluster. We shouldn't be making assumptions on the namespace where the operator is running. We don't do this in other parts of the code. Not sure why you think you are making that assumption here.

cmd/manager/cleanup.go
```diff
@@ -0,0 +1,178 @@
...
+		if k8s_errors.IsNotFound(err) {
+			// guess the catalogsource from the opsrc that has the package
+			datastoreCs, err = findCsFromOpsrc(&instance, kubeClient)
+			if err != nil {
+				return nil, err
+			}
+		}
+		// point the subscription to the datastore CatalogSource
+		instance.SetCatalogSource(datastoreCs.GetName())
+		instance.SetCatalogSourceNamespace("openshift-marketplace")
```
aravindhp: I see where you are hard-coding the namespace. This is not required. You should be able to infer this. This applies to all instances where you are doing this.

cmd/manager/main.go
kevinrizza: Why can't we just use a different namespace if we find it's a non openshift cluster?

cmd/manager/cleanup.go
```diff
@@ -0,0 +1,178 @@
+package main
```
kevinrizza: In general, I would like to state that this seems like the wrong abstraction. It is unclear what is being "cleaned" and "why" we are cleaning. I think that instead that we would be better with a migrator of some kind, if we feel like an abstraction is needed. And that migrator should know about what specific migration is happening, not that just generically we need to "clean up".

Also, have we considered if this code will stick around forever? Or is this something we will remove after openshift 4.2 release? It seems like if (more likely, when) we need to do another such migration it wouldn't make sense to do these checks as well.

cmd/manager/cleanup.go
anik120: @aravindhp this is the `CatalogSource` that a subscription will have information about after it is updated. Existing Subscriptions have information about the CSCs installed during operator installation. For example, if a CSC name `installed-community-openshift-operators` was created during installation of an operator, the Subscription had the following information
```
spec:
     source: installed-community-openshift-operators
     sourceNamespace: openshift-operators
```

This has to be changed to contain information about the datastore CatalogSource instead. 

```
spec:
     source: community-operators
     sourceNamespace: openshift-marketplace
```
Given that, shouldn't it be `datastoreCs`?

aravindhp: No. Datastore `CatalogSourceConfigs` and `CatalogSources` were created by `OperatorSources` before #198.

cmd/manager/cleanup.go
anik120: @kevinrizza what if we moved all of these to a separate package called `migrator`, and document why we need this `migrator` to migrate from a openshift 4.1.1 release to a 4.1.2? (Hope I'm getting the versions correct). 
That way, when we need to do another migration, the whole abstraction can be removed.

cmd/manager/main.go
awgreene: We also want to make sure that the `ClusterOperator` is not set to available while resources are being migrated - If the Upgrade path is 4.1->4.2->4.3, we don't want the upgrade from 4.2->4.3 to kickoff before this completes.

deploy/role.yaml
```diff
@@ -22,6 +22,21 @@ rules:
   - create
   - get
   - update
+- apiGroups:
```
awgreene: This change needs to be applied to `upstream` as well.

cmd/manager/main.go
aravindhp: Each cluster upgrade is a distinct operation. I don't think the 4.2 --> 4.3 option will be available until 4.1 --> 4.2 is complete.

cmd/manager/cleanup.go
aravindhp: 4.1.x to 4.2.0 is what we are concerned about here.

cmd/manager/main.go
awgreene: I agree - but if marketplace reports that progression has finished we risk a race condition.

aravindhp: I don't understand how there is a race condition. If marketplace has not finished migration, then the cluster is not ready for the next upgrade.

cmd/manager/cleanup.go
anik120: @aravindhp what should the updated Subscription contain information about in it's `spec.Source` and `spec.SourceNamesapce` fields?

aravindhp: > arent datastore CatalogSources being created directly now by OperatorSources?

Correct

> Shouldn't the updated Subscription contain information about the datastore CatalogSources in it's spec.Source and spec.SourceNamesapce fields?

Yes when you update the `Subscription`, it should reference the new datastore `CatalogSource`. The problem here was with some of the function naming that was confusing me. Given you have already pushed changes, I am unable to clarify further. In future, please wait for such conversations to get resolved before you push changes.

cmd/manager/main.go
aravindhp: I don't want to make any assumptions on the name of the namespace.

cmd/manager/cleanup.go
aravindhp: I am +1 on `migrator` too.

cmd/manager/main.go
```diff
@@ -98,6 +100,19 @@ func main() {
 	go registrySyncer.Sync(stopCh)
 	go catalogsourceconfig.Syncer.Sync(stopCh)
 
+	client, err := client.New(config.GetConfigOrDie(), client.Options{})
+	if err != nil {
+		exit(err)
+	}
+	upgrader := migrator.NewUpgrader(log.NewEntry(log.New()), client)
```
kevinrizza: These seem like implementation details that should be obfuscated by the migrator. Why can't we just do all of this with one call? We should just do something like:

```
upgrader := migrator.NewUpgrader(log.NewEntry(log.New()), client)
err := upgrader.Migrate()
if err != nil {
	exit(err)
}
```

And you should implement a method on that migrator type that updates subscriptions and cleans up resources as implementation details.

If we start adding the details of _how_ we do the migration, especially if there are a lot of steps in future iterations, it will be really hard to read `main.go` -- which to me is already too long and complex.

kevinrizza: As an aside, this will also make it easier to write unit tests for the migrator package.

pkg/migrator/cleanup.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+// getSubsPackageName returns the packageName of a Subscription
+func getSubsPackageName(subscription *olm.Subscription) string {
+	return subscription.Spec.Package
+}
+
+// setSubsCatalogSource sets the Spec.source and Spec.sourceNamespace fields of a
+// Subscription, and returns the subscription
+func setSubsCatalogSource(subscription *olm.Subscription, csName string, csNamespace string) *olm.Subscription {
```
kevinrizza: If you used namespaced name here, it would be simpler. Also, what is the value of this method? It seems like we only call it in one place.

pkg/migrator/cleanup.go
```diff
@@ -0,0 +1,246 @@
...
+	subscription.Spec.CatalogSource = csName
+	subscription.Spec.CatalogSourceNamespace = csNamespace
+	return subscription
+}
+
+func helperNewCatalogSourceConfig(namespace, name string) *v2.CatalogSourceConfig {
```
kevinrizza: This function seems named poorly. We should, in general, avoid a noun like `helper` when we are just describing a function instead of a type. I would buy that a `helper` would implement a method like `GenerateCatalogSourceConfig` and be called as `helper.GenerateCatalogSourceConfig()`.  But since there is no object, we should just call it something like `func GenerateCatalogSourceConfig()`.


Also you created this `cscInfo` type which has the fields you use as function arguments, but you don't use it here.

pkg/migrator/cleanup.go
```diff
@@ -0,0 +1,246 @@
...
+
+// extractCsName takes the name of a CatalogSourceConfig
+// as input, and infers the name of the corresponding
+// CatalogSource datastore. Installed CSCs follow the
+// following naming pattern: installed-publisher-namespace.
+// For example, for a CatalogSourceConfig named
+// `installed-community-openshift-operators`,
+// extractCsName returns `community-operators` as output
+func extractCsName(cscName string) (string, error) {
+	var stringBuilder strings.Builder
```
kevinrizza: strings.Builder seems like slight overkill here, no? If we are just appending two strings why not just use `fmt.Sprintf("%s-%s", possibleCsName, "-operators")`?

pkg/migrator/cleanup.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+// getOpsrcPackages returns the list of packages from the Status block
+func getOpsrcPackages(opsrc *v1.OperatorSource) []string {
+	return strings.Split(opsrc.Status.Packages, ",")
+}
+
+// getSubsPackageName returns the packageName of a Subscription
+func getSubsPackageName(subscription *olm.Subscription) string {
+	return subscription.Spec.Package
```
kevinrizza: This function seems like overkill for a single liner that is called once and just gets the field. Why not just do it inline?

pkg/migrator/cleanup.go
```diff
@@ -0,0 +1,246 @@
...
+	err := kubeClient.List(context.TODO(), options, cscs)
+	if err != nil {
+		return err
+	}
+	for _, csc := range cscs.Items {
+		err = kubeClient.Delete(context.TODO(), &csc)
+		if err != nil {
+			logger.Errorf("[migration] Failed to delete CatalogSourceConfig %s. Error: %s", csc.GetName(), err.Error())
+		} else {
+			logger.Printf("[migration] Succesfully deleted datastore CSC %s", csc.GetName())
```
kevinrizza: This is technically not really true. It just means that the client returned a success and it was scheduled for deletion. We don't know that the `csc` was actually deleted yet. In fact, it won't be deleted until the finalizer is called and is executed, so there is other code in the operator that may have not been called yet. Since that's the case, I would argue that we shouldn't log this message. Instead, we could say something like "CSC %s scheduled for deletion"

pkg/migrator/cleanup.go
```diff
@@ -0,0 +1,246 @@
...
+// deleteInstalledCscs deletes the CSCs installed during opeartor installation.
+// The child resources of the CSCs are delete by the finalizer.
+func deleteInstalledCscs(cscs []cscInfo, kubeClient client.Client, logger *logrus.Entry) error {
+	for _, cscInfo := range cscs {
+		csc := helperNewCatalogSourceConfig(cscInfo.namespace, cscInfo.name)
+		err := kubeClient.Delete(context.TODO(), csc)
+		if err != nil {
+			logger.Errorf("[migration] Failed to delete installed CSC %s with error: ", cscInfo.name, err.Error())
+		} else {
+			logger.Printf("[migration] Succesfully deleted stale CSC %s,", cscInfo.name)
```
kevinrizza: Same argument as other deletion message.

pkg/migrator/cleanup.go
```diff
@@ -0,0 +1,246 @@
...
+
+	cscs := &v2.CatalogSourceConfigList{}
+	// Get the list of existing cscs that have the lable "opsrc-datastore: "true" "
+	err := kubeClient.List(context.TODO(), options, cscs)
+	if err != nil {
+		return err
+	}
+	for _, csc := range cscs.Items {
+		err = kubeClient.Delete(context.TODO(), &csc)
+		if err != nil {
```
kevinrizza: Why don't we ever try to return this error? I understand the desire to try on every item before we give up, but we could aggregate the error and return a list of errors at the end, or even aggregate the errors into a single error when returning.

pkg/migrator/cleanup.go
```diff
@@ -0,0 +1,246 @@
...
+	for _, cscInfo := range cscs {
+		csc := helperNewCatalogSourceConfig(cscInfo.namespace, cscInfo.name)
+		err := kubeClient.Delete(context.TODO(), csc)
+		if err != nil {
+			logger.Errorf("[migration] Failed to delete installed CSC %s with error: ", cscInfo.name, err.Error())
+		} else {
+			logger.Printf("[migration] Succesfully deleted stale CSC %s,", cscInfo.name)
+		}
+	}
+	return nil
```
kevinrizza: This method never actually returns an error, why always return nil? We should aggregate the kubeclient's errors and return those if they exist.

anik120:
@aravindhp @kevinrizza @awgreene addressed your comments. PTAL.

test/helpers/helpers.go
```diff
@@ -352,3 +369,135 @@ func CheckOpsrcChildResourcesDeleted(client test.FrameworkClient, opsrcName stri
...
+			Channel:                "alpha",
+		},
+	}
+}
+
+// CheckSubsciptionSuccessfulUpdation checks that a given Subscription
+// has been succesfully updated on cluster upgrade from openshift 4.1.x to
+// openshift 4.2.0 to contain information about the datastore CatalogSource
+// instead of the installed CatalogSource
+func CheckSubsciptionSuccessfulUpdation(client test.FrameworkClient, name, namespace string) error {
```
aravindhp: Subscription is misspelled. Moreover is this actually an helper? This seems like the test itself.

test/helpers/helpers.go
```diff
@@ -352,3 +369,135 @@ func CheckOpsrcChildResourcesDeleted(client test.FrameworkClient, opsrcName stri
...
+	subscription := &olm.Subscription{}
+	err := client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: namespace}, subscription)
+	if err != nil {
+		return err
+	}
+	// Check that subscription has infromation about the datastore CatalogSource,
```
aravindhp: Information is misspelled. Moreover please say references the datastore CatalogSource.

test/testgroups/migrationtests.go
```diff
@@ -0,0 +1,44 @@
...
+
+	// Create the installed CatalogSourceConfig.
+	err = helpers.CreateRuntimeObject(test.Global.Client, ctx, helpers.CreateInstalledCsc(namespace))
+	require.NoError(t, err, "Could not create installed CatalogSourceConfig")
+
+	// Create the datastore CatalogSourceConfig.
+	err = helpers.CreateRuntimeObject(test.Global.Client, ctx, helpers.CreateDatastoreCsc(helpers.TestDatastoreCscName, namespace))
+	require.NoError(t, err, "Could not create datastore CatalogSourceConfig")
+	// Wait for the corresponding datastore CatalogSource to deploy successfully
+	helpers.WaitForDeployment(helpers.TestDatastoreCscName, namespace, test.Global.Client)
```
aravindhp: You need to use `CheckCscChildResourcesCreated()` and ensure (`require.NoError`) that it succeeded.

test/testsuites/migrationtests.go
```diff
@@ -0,0 +1,51 @@
...
+	require.NoError(t, err)
+
+	// Check for successful deletion of installed CSC.
+	installedCscName := helpers.TestInstalledCscPublisherName + namespace
+	err = helpers.CheckCscSuccessfulDeletion(test.Global.Client, installedCscName, namespace, namespace)
+	require.NoError(t, err)
+}
+
+// testSubscriptionsUpdated ensures that Subscriptions created during operator installation
+// in a openshift 4.1.x cluster are updated to contain information about datastore CatalogSources
```
aravindhp: Please use reference instead of information. Moreover, clearly stating what you are expecting in the `Subscription` will be helpful.

aravindhp:
- Update the commit header to `[migration] Update existing Subscriptions and remove stale resources`
- Mention why you are updating the rules if they end up being required
- Mention the Subscriptions are created by the UI
- Please be consistent regarding using `CatalogSources` in all places instead of `catalogSources` or `catalogsources`

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+// Migrate updates existing subscriptions, deletes the CSCs installed during
+// operator installation as well the stale datastore CSCs when the cluster is
+// migrating from Openshift 4.1.x to Openshift 4.2.x
+func (m *migrator) Migrate() error {
+
+	installedCscs, err := UpdateSubscriptions(m.client, m.logger)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
anik120: @aravindhp `UpdateSubscriptions` does a best effort update of all the Subscriptions, and only returns an error when the `client.List` operation throws an error (say for RBAC related issues).
I have tried to classify errors in this PR into two categories: 
A) Errors encountered during updating/cleaning up some resources
B) Errors encountered otherwise (say RBAC related issues etc.) 
Type `A` errors are logged and not returned at the first error encountered. 
Type `B` errors are returned as soon as a first error is encountered.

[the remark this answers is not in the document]
aravindhp: By returning here you are not attempting to delete the stale datastore `CatalogSourceConfigs`. Which is why I am have mentioned later that `UpdateSubscription` shouldn't return an error. It should just return a list of installed `CatalogSourceConfigs` that need to be deleted. If you ran into an error and couldn't collect any installed `CatalogSourceConfigs` then the returned list would be empty.

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,231 @@
...
+// findCatalogSource infers the datastore CatalogSource that a given existing Subscription will
+// contain information about in it's spec.source and spec.sourceNamespace fields
+func findCatalogSource(subscription *olm.Subscription, client client.Client) (*olm.CatalogSource, error) {
+	associatedCscName := subscription.GetLabels()[builders.OwnerNameLabel]
+	possibleCsName := extractCsName(associatedCscName)
+	possibleCsNamespace := subscription.GetLabels()[builders.OwnerNamespaceLabel]
+	// try and fetch the CatalogSource
+	datastoreCs := &olm.CatalogSource{}
+	namespacedName := types.NamespacedName{Name: possibleCsName, Namespace: possibleCsNamespace}
+	err := client.Get(context.TODO(), namespacedName, datastoreCs)
```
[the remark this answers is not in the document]
anik120: @aravindhp trying to fetch the `CatalogSource` is more for creating a back up plan in case inferring the `CatalogSource` name from the `csc-owner-name` label fails.  
If the inferred `CatalogSource` does not exist, the `migrator` then moves onto the next strategy to find the corresponding `datastore` `CatalogSource`.

cmd/manager/main.go
```diff
@@ -92,6 +94,18 @@ func main() {
 	log.Print("Starting the Cmd.")
 	stopCh := signals.SetupSignalHandler()
 
+	// set clusterOperator status to report Migration
+	status.ReportMigration(cfg, mgr, namespace, os.Getenv("RELEASE_VERSION"), stopCh)
+	// Perform migration logic to upgrade cluster from 4.1.x to 4.2.0
+	client, err := client.New(config.GetConfigOrDie(), client.Options{})
+	if err != nil {
+		exit(err)
```
aravindhp: Do we really want to terminate the operator if we cannot migrate? Even if the CVO restarts the operator, we could end up in the same situation again. This will basically result in a cluster that doesn't have a marketplace operator. The same comment applies for the `Migrate()` call.

pkg/migrator/migrator.go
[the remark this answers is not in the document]
aravindhp: Sure but the answer in both cases should be the same. More over if you attempted the get when someone deleted the datastore `CatalogSource` and the watch which I am adding brings it back, it will still result in a `Subscription` not being updated.

cmd/manager/main.go
kevinrizza: It seems like the right thing to do would be to retry until it succeeds, given the normal pattern of eventual consistency.

aravindhp: What if it is stuck here for whatever reason? We will be migrating for ever. So either we do this in a background goroutine or do a best effort migration.

anik120: > What if it is stuck here for whatever reason? 

Exactly my concern. Do we want to keep consuming resources in an effort to migrate, if we are unable to migrate fully?

anik120: Also, the Subscriptions are updated by the migrator, so unless there's a switch which indicates if a Subscription has been updated or not, all the Subscriptions will keep getting updated, albeit just one field.

Keeping track of which Subscriptions were updated means consuming even more resources.

aravindhp: Then we should be removing the labels on the Subscriptions that we update

kevinrizza: I think that doing it every time we start is just part of the deal. I'm not sure how else we would handle it.

Are we okay with our migration failing and then moving on? Perhaps it depends on our definition of `failure`?

anik120: > Then we should be removing the labels on the Subscriptions that we update

@aravindhp Are we sure we can do that without producing any side effects? 

> Are we okay with our migration failing and then moving on? Perhaps it depends on our definition of failure?

Maybe clean up/ update as many resources possible, and report resources that could not be cleaned/updated for admin to act on them?

aravindhp: >Are we sure we can do that without producing any side effects?

Yes. Note that this does not mean removing *all* labels. Just the ones we care about.

>Maybe clean up/ update as many resources possible,

This is what we should be doing in general.

> report resources that could not be cleaned/updated for admin to act on them?

How will you report? Asking admins to look at logs is not acceptable. Please think of worst case scenarios where not updating a Subscription or not being able to delete the datastore/installed `CatalogSourceConfigs` will cause issues.

anik120: > Please think of worst case scenarios where not updating a Subscription or not being able to delete the datastore/installed CatalogSourceConfigs will cause issues.

Does that mean we're agreeing on creating a go routine that keeps making best effort to clean up resources until there's no errors?

aravindhp: You shouldn't jump  to that conclusion until you explore the scenarios.

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	deleteInstalledCscs(installedCscs, m.client, m.logger)
+	err = deleteDatastoreCscs(m.client, m.logger)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// UpdateSubscriptions updates the existing Subscriptions'
```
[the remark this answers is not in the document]
anik120: @aravindhp I was going for the plural `Subscriptions` as `updateSubscriptions` are updating multiple subscriptions.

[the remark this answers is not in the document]
aravindhp: OK. But please update the doc to say Subscriptions reference installed `CatalogSources`. Also use `Subscriptions` everywhere instead of `subscriptions` in some places.

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,231 @@
...
+// are updated with information about datastore CatalogSources
+// instead.
+func UpdateSubscriptions(kubeClient client.Client, logger *logrus.Entry) ([]types.NamespacedName, error) {
+	var installedCscs []types.NamespacedName
+	options := &client.ListOptions{}
+	options.SetLabelSelector(fmt.Sprintf(builders.OwnerNameLabel))
+
+	subscriptions := &olm.SubscriptionList{}
+	// Get the list of existing subscriptions that have the label "csc-owner-name"
+	err := kubeClient.List(context.TODO(), options, subscriptions)
```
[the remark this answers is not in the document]
anik120: > How does the "csc-owner-name" label end up in the Subscription object? Does OLM pass this along based on the referenced CatalogSource in the Subscription?

@aravindhp I remember the UI code adding the labels [here] (https://sourcegraph.com/github.com/openshift/console/-/blob/frontend/public/components/operator-hub/operator-hub-subscribe.tsx#L110), but looks like that's changed now. 

> You are also listing Subscriptions in all namespaces. Please confirm that this is required. I don't remember off the top of my head where the Subscriptions get created by the UI.

The UI creates the Subscriptions in the `openshift-operator` namespace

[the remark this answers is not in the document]
aravindhp: > I remember the UI code adding the labels here, but looks like that's changed now.

You should be looking at the 4.1 branch

>The UI creates the Subscriptions in the `openshift-operator` namespace

Then you do need the cluster role. It basically means that the `Subscriptions` could be anywhere as you are allowed to pick a namespace for an operator and the UI will create the `OperatorGroup` for that namespace if needed. In that case I would think the `Subscription` would be in the selected namespace.

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,231 @@
...
+	if err != nil {
+		return nil, err
+	}
+	for _, instance := range subscriptions.Items {
+		newCscInfo := types.NamespacedName{Name: instance.GetLabels()[builders.OwnerNameLabel], Namespace: instance.GetLabels()[builders.OwnerNamespaceLabel]}
```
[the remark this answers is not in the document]
anik120: > What if csc-owner-namespace is not present?

I don't see a scenario where `csc-owner-namespace` is not present, if `csc-owner-namespace` is present. That would be an inconsistency that'd need to be addressed wherever the `Subscription` object is being created 

> PS: I don't buy the readability argument that many people give for this. Add comments instead. ;-)

I guess you foresaw my argument :)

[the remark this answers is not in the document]
aravindhp: >I don't see a scenario where csc-owner-namespace is not present, if csc-owner-namespace is present. That would be an inconsistency that'd need to be addressed wherever the Subscription object is being created

There could be a potential scenario where an user or some other entity removes the label. You have to account for that.

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,231 @@
...
+
+// deleteDatastoreCscs deletes the datastore CSCs created by OperatorSources.
+// The child resources of the CSCs are delete by the finalizer.
+func deleteDatastoreCscs(kubeClient client.Client, logger *logrus.Entry) error {
+	options := &client.ListOptions{}
+	options.SetLabelSelector(fmt.Sprintf("opsrc-datastore: \"true\""))
+
+	cscs := &v2.CatalogSourceConfigList{}
+	// Get the list of existing cscs that have the lable "opsrc-datastore: "true" "
+	err := kubeClient.List(context.TODO(), options, cscs)
```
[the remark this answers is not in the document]
anik120: @aravindhp Problem is determining the namespace of the `marketplace-operator`. I don't suppose there's a clean way of determining the namespace of the operator (`openshift-marketplace` or `marketplace` etc). Listing the `CatalogSourceConfigs` in all namespaces seems like the cleanest way to do it.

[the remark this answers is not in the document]
aravindhp: https://github.com/operator-framework/operator-marketplace/blob/master/cmd/manager/main.go#L49

test/helpers/helpers.go
```diff
@@ -352,3 +369,135 @@ func CheckOpsrcChildResourcesDeleted(client test.FrameworkClient, opsrcName stri
...
+			TargetNamespace: namespace,
+			Packages:        "",
+		},
+	}
+}
+
+// CreateSubscription returns a newly built Subscription with the lables
+// `csc-owner-name` and `csc-owner-namespace`
+func CreateSubscription(name, namespace string) *olm.Subscription {
+	lables := make(map[string]string)
```
kevinrizza: this should be `labels`

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,231 @@
...
+func getOpsrcPackages(opsrc *v1.OperatorSource) []string {
+	return strings.Split(opsrc.Status.Packages, ",")
+}
+
+// setSubsCatalogSource sets the Spec.source and Spec.sourceNamespace fields of a
+// Subscription, and returns the subscription
+func setSubsCatalogSource(subscription *olm.Subscription, csName string, csNamespace string) *olm.Subscription {
+	subscription.Spec.CatalogSource = csName
+	subscription.Spec.CatalogSourceNamespace = csNamespace
+	return subscription
```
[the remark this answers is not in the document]
anik120: @aravindhp We want the subscription to be updated to the new values. 
Does that answer your question or am I misunderstanding the question?

pkg/migrator/migrator.go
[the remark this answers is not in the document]
anik120: @aravindhp there are two reasons for checking if the `CatalogSource` exists: 

A) If the datastore `CatalogSource` has been deleted then it shouldn't be referenced from the `Subscription`

Sure, with the watch you're adding it'll be brought back, but there's another scenario where:

B) The infered `CatalogSource` name is wrong. For example if the value of `csc-owner-label` is set to `installed-foo-openshift-marketplace`, and the `foo-operators` `CatalogSource` does not exist simply because `foo` in `installed-foo-openshift-marketplace` was misplaced in the first place, then the `Subscription` shouldn't reference a non-existent `CatalogSource`. Checking if the `CatalogSource` exists provides us the safety net.

pkg/migrator/migrator.go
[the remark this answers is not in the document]
aravindhp: What I am trying to say is `updatedSub := setSubsCatalogSource(&instance, datastoreCs.GetName(), datastoreCs.GetNamespace())`  is equivalent to `setSubsCatalogSource(&instance, datastoreCs.GetName(), datastoreCs.GetNamespace())`. You can use `instance` instead of `updatedSub` given you are using a pointer. 

You might also want to consider if this function is really required given you are only calling it in one place and the amount of work that is being done in it.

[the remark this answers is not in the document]
anik120: Ah, true. It made sense when I had accidentally modified the vendor code, but now that I'm not doing that the function doesn't make that much sense. I'll update the object at the call site.

pkg/migrator/migrator.go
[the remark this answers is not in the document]
aravindhp: Good point regarding B.

anik120:
@aravindhp @kevinrizza addressed your comments. PTAL.
Pending discussion: https://github.com/operator-framework/operator-marketplace/pull/207#discussion_r297678569

aravindhp:
Continued in #223

PR #208, opened 2019-06-20
Bug 1723672: Fix owner labels to specify Kind

galletti94:
Currently there is no way to determine the Kind that owns a given
resource the operator creates. This commit propagates information
about the Kind to the label creation function in the each resource
builder. It also update the deletion function to use these updated
labels as labelSelectors.

### DIFF
--- a/pkg/builders/builders.go
+++ b/pkg/builders/builders.go
@@ -1,21 +1,45 @@
 package builders
 
-// OwnerNameLabel is the label used to mark ownership over a given resources.
-// When this label is set, the reconciler should handle these resources when the owner
-// is deleted.
-const OwnerNameLabel string = "csc-owner-name"
+import (
+	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v2"
+)
 
-// OwnerNamespaceLabel is the label used to mark ownership over a given resources.
-// When this label is set, the reconciler should handle these resources when the owner
-// is deleted.
-const OwnerNamespaceLabel string = "csc-owner-namespace"
+// CscOwnerNameLabel is the label used to indicate the name of the CatalogSourceConfig
+// that owns this resources. When this label is set, the reconciler should handle these
+// resources when the CatalogSourceConfig is deleted.
+const CscOwnerNameLabel string = "csc-owner-name"
 
-// OpsrcOwnerNameLabel is the label used to mark ownership over resources
-// that are owned by the OperatorSource. When this label is set, the reconciler
-// should handle these resources when the OperatorSource is deleted.
+// CscOwnerNamespaceLabel is the label used to indicate the namespace of the CatalogSourceConfig
+// that owns this resources. When this label is set, the reconciler should handle these
+// resources when the CatalogSourceConfig is deleted.
+const CscOwnerNamespaceLabel string = "csc-owner-namespace"
+
+// OpsrcOwnerNameLabel is the label used to indicate the name of the OperatorSource
+// that owns this resources. When this label is set, the reconciler should handle these
+// resources when the OperatorSource is deleted.
 const OpsrcOwnerNameLabel string = "opsrc-owner-name"
 
-// OpsrcOwnerNamespaceLabel is the label used to mark ownership over resources
-// that are owned by the OperatorSource. When this label is set, the reconciler
-// should handle these resources when the OperatorSource is deleted.
+// OpsrcOwnerNamespaceLabel is the label used to indicate the namespace of the OperatorSource
+// that owns this resources. When this label is set, the reconciler should handle these
+// resources when the OperatorSource is deleted.
 const OpsrcOwnerNamespaceLabel string = "opsrc-owner-namespace"
+
+// GetOwnerLabel returns a map with either the CatalogSourceConfig or the OperatorSource owner
+// name and namespace labels depending what kind is the owner
+func GetOwnerLabel(name, namespace, owner string) map[string]string {
+	switch owner {
+	case v1.OperatorSourceKind:
+		return map[string]string{
+			OpsrcOwnerNameLabel:      name,
+			OpsrcOwnerNamespaceLabel: namespace,
+		}
+	case v2.CatalogSourceConfigKind:
+		return map[string]string{
+			CscOwnerNameLabel:      name,
+			CscOwnerNamespaceLabel: namespace,
+		}
+	default:
+		return map[string]string{}
+	}
+}
--- a/pkg/builders/catalogsourcebuilder.go
+++ b/pkg/builders/catalogsourcebuilder.go
@@ -61,13 +61,26 @@ func (b *CatalogSourceBuilder) WithOLMLabels(cscLabels map[string]string) *Catal
 	return b
 }
 
-// WithOwnerLabel sets the owner label of the CatalogSource object to the given owner.
-func (b *CatalogSourceBuilder) WithOwnerLabel(name, namespace string) *CatalogSourceBuilder {
+// WithOpsrcOwnerLabel sets the owner label of the CatalogSource object to the given owner.
+func (b *CatalogSourceBuilder) WithOpsrcOwnerLabel(name, namespace string) *CatalogSourceBuilder {
 	labels := map[string]string{
-		OwnerNameLabel:      name,
-		OwnerNamespaceLabel: namespace,
+		OpsrcOwnerNameLabel:      name,
+		OpsrcOwnerNamespaceLabel: namespace,
+	}
+	for key, value := range b.cs.GetLabels() {
+		labels[key] = value
 	}
 
+	b.cs.SetLabels(labels)
+	return b
+}
+
+// WithCscOwnerLabel sets the owner label of the CatalogSource object to the given owner.
+func (b *CatalogSourceBuilder) WithCscOwnerLabel(name, namespace string) *CatalogSourceBuilder {
+	labels := map[string]string{
+		CscOwnerNameLabel:      name,
+		CscOwnerNamespaceLabel: namespace,
+	}
 	for key, value := range b.cs.GetLabels() {
 		labels[key] = value
 	}
--- a/pkg/grpccatalog/grpccatalog.go
+++ b/pkg/grpccatalog/grpccatalog.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
 	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v2"
 	"github.com/operator-framework/operator-marketplace/pkg/builders"
 	wrapper "github.com/operator-framework/operator-marketplace/pkg/client"
@@ -41,7 +42,7 @@ type GrpcCatalog struct {
 // EnsureResources creates a GRPC CatalogSource if one does not already
 // exists otherwise it updates an existing one. It then creates/updates all the
 // resources it requires.
-func (r *GrpcCatalog) EnsureResources(key types.NamespacedName, displayName, publisher, targetNamespace, source, packages string, labels map[string]string) error {
+func (r *GrpcCatalog) EnsureResources(key types.NamespacedName, displayName, publisher, targetNamespace, source, packages, owner string, labels map[string]string) error {
 	// Ensure reader is not nil
 	if r.reader == nil {
 		return fmt.Errorf("GrpcCatalog.reader is not defined")
@@ -53,7 +54,7 @@ func (r *GrpcCatalog) EnsureResources(key types.NamespacedName, displayName, pub
 	}
 
 	// Ensure that a registry deployment is available
-	registry := registry.NewRegistry(r.log, r.client, r.reader, key, source, packages, registry.ServerImage)
+	registry := registry.NewRegistry(r.log, r.client, r.reader, key, source, packages, registry.ServerImage, owner)
 	err = registry.Ensure()
 	if err != nil {
 		return err
@@ -79,7 +80,7 @@ func (r *GrpcCatalog) EnsureResources(key types.NamespacedName, displayName, pub
 		r.log.Infof("Updated CatalogSource %s", catalogSourceGet.Name)
 	} else {
 		// Create the CatalogSource structure
-		catalogSource := newCatalogSource(labels, csKey, displayName, publisher, key.Namespace, registry.GetAddress())
+		catalogSource := newCatalogSource(labels, csKey, displayName, publisher, key.Namespace, registry.GetAddress(), owner)
 		r.log.Infof("Creating CatalogSource %s", catalogSource.Name)
 		err = r.client.Create(context.TODO(), catalogSource)
 		if err != nil && !errors.IsAlreadyExists(err) {
@@ -93,12 +94,15 @@ func (r *GrpcCatalog) EnsureResources(key types.NamespacedName, displayName, pub
 }
 
 // newCatalogSource returns a CatalogSource object.
-func newCatalogSource(labels map[string]string, key types.NamespacedName, displayName, publisher, namespace, address string) *olm.CatalogSource {
+func newCatalogSource(labels map[string]string, key types.NamespacedName, displayName, publisher, namespace, address, owner string) *olm.CatalogSource {
 	builder := new(builders.CatalogSourceBuilder).
-		WithOwnerLabel(key.Name, namespace).
 		WithMeta(key.Name, key.Namespace).
 		WithSpec(olm.SourceTypeGrpc, address, displayName, publisher)
-
+	if owner == v1.OperatorSourceKind {
+		builder.WithOpsrcOwnerLabel(key.Name, namespace)
+	} else if owner == v2.CatalogSourceConfigKind {
+		builder.WithCscOwnerLabel(key.Name, namespace)
+	}
 	// Check if the operatorsource.DatastoreLabel is "true" which indicates that
 	// the CatalogSource is the datastore for an OperatorSource. This is a hint
 	// for us to set the "olm-visibility" label in the CatalogSource so that it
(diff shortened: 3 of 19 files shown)


galletti94:
@aravindhp @kevinrizza @awgreene @anik120 Please review when you get a chance

pkg/builders/catalogsourcebuilder.go
```diff
@@ -62,10 +62,14 @@ func (b *CatalogSourceBuilder) WithOLMLabels(cscLabels map[string]string) *Catal
...
-		OwnerNameLabel:      name,
-		OwnerNamespaceLabel: namespace,
+func (b *CatalogSourceBuilder) WithOwnerLabel(name, namespace string, isOpsrcOwner bool) *CatalogSourceBuilder {
+	labels := map[string]string{}
+	if isOpsrcOwner {
+		labels[OpsrcOwnerNameLabel] = name
+		labels[OpsrcOwnerNamespaceLabel] = namespace
+	} else {
+		labels[OwnerNameLabel] = name
+		labels[OwnerNamespaceLabel] = namespace
```
awgreene: This code is reused in multiple places. Could we create `getOpsrcOwnerLables` and `getCscOwnerLabels` functions that return the labels map? That way, if the labels change in the future we can address this all in one spot.

pkg/operatorsource/deleted.go
```diff
@@ -63,8 +63,13 @@ func (r *deletedReconciler) Reconcile(ctx context.Context, in *v1.OperatorSource
 	r.datastore.RemoveOperatorSource(out.UID)
 	grpcCatalog := grpccatalog.New(r.logger, nil, r.client)
 
-	// Delete the owned registry resources.
-	err = grpcCatalog.DeleteResources(ctx, in.Name, in.Namespace, in.Namespace)
+	// Delete the owned registry resources. Need two calls here the first will
```
awgreene: Please use complete sentences when writing comments.

pkg/operatorsource/deleted.go
```diff
@@ -63,8 +63,13 @@ func (r *deletedReconciler) Reconcile(ctx context.Context, in *v1.OperatorSource
...
+	// delete all resources with the opsrc owner label, the second will delete
+	// all resources with csc owner label. This is needed for the upgrade case
+	// where services, deployments, etc. from 4.1 only have the csc owner label
+	// but should be deleted when the opsrc is deleted using the 4.2 operator
```
awgreene: Nit: Please end sentences with periods.

awgreene: This could introduce a bug where a user creates a `CSC` with the same name as an existing `OpSrc`. If we want to address this edgecase, we could make the second call if-and-only-if the  first call did not delete any resources.

pkg/status/status.go
```diff
@@ -250,7 +250,7 @@ func (s *status) updateStatus(previousStatus *configv1.ClusterOperatorStatus) er
 		// Log Conditions
 		log.Infof("[status] Attempting to set the ClusterOperator status conditions to:")
 		for _, statusCondition := range s.clusterOperator.Status.Conditions {
-			log.Infof("[status] ConditionType: %v ConditionStatus: ConditionMessage: %v", statusCondition.Type, statusCondition.Status, statusCondition.Message)
+			log.Infof("[status] ConditionType: %v ConditionStatus: %v ConditionMessage: %v", statusCondition.Type, statusCondition.Status, statusCondition.Message)
```
awgreene: Please call out that you are fixing a bug in your commit message if this is meant to be included in your PR.

pkg/builders/catalogsourcebuilder.go
galletti94: Good suggestion! What do you think @kevinrizza @aravindhp @anik120 ?

pkg/operatorsource/deleted.go
galletti94: The first call will always delete at least the `catalogsource` because it will have the opsrc labels. Wouldn't checking that the service/deployment etc. were deleted take a lot of time?

pkg/builders/catalogsourcebuilder.go
awgreene: Alternatively you could create a `getLabels` method that takes the `isOpsrcOwner` bool and returns the correct labels - that way you don't add  `if/else` statements everywhere.

pkg/status/status.go
galletti94: my bad I fixed this in passing because I was getting errors and forgot to remove it. Can put this in a separate commit

pkg/builders/catalogsourcebuilder.go
```diff
@@ -62,10 +62,14 @@ func (b *CatalogSourceBuilder) WithOLMLabels(cscLabels map[string]string) *Catal
 }
 
 // WithOwnerLabel sets the owner label of the CatalogSource object to the given owner.
-func (b *CatalogSourceBuilder) WithOwnerLabel(name, namespace string) *CatalogSourceBuilder {
-	labels := map[string]string{
-		OwnerNameLabel:      name,
-		OwnerNamespaceLabel: namespace,
+func (b *CatalogSourceBuilder) WithOwnerLabel(name, namespace string, isOpsrcOwner bool) *CatalogSourceBuilder {
```
kevinrizza: Using the boolean `isOpsrcOwner` here seems like an antipattern to me. One of the value adds to the builder pattern is that you can make these methods cumulative. I think adding the `OpsrcOwner` or `Owner` label should just be separated into separate methods so it is clearer and more explicit as to exactly what is being set.

kevinrizza: Also, apply this comment everywhere that you are using this pattern in the PR.

pkg/registry/registry.go
```diff
@@ -38,14 +38,15 @@ const DefaultServerImage = "quay.io/openshift/origin-operator-registry"
...
-	key      types.NamespacedName
-	image    string
-	address  string
+	log          *logrus.Entry
+	client       wrapper.Client
+	reader       datastore.Reader
+	source       string
+	packages     string
+	key          types.NamespacedName
+	isOpsrcOwner bool
```
kevinrizza: This variable name is confusing. What is its purpose, and why is it tied to an instance of the `registry` type? Is it because we happen to instantiate the `registry` separately for the `csc` controller and the `opsrc` controller? If so, it seems like that is a coincidental concern and we should be more respectful of the purpose of these objects and their composition.

pkg/operatorsource/deleted.go
galletti94: If you create a `csc` with the same name as the `opsrc` it will try to create resources that already exist. Is this a scenario we need to be concerned with?

pkg/builders/catalogsourcebuilder.go
galletti94: Yeah that makes sense

pkg/registry/registry.go
galletti94: How would you propagate this information to the builders though?

pkg/status/status.go
kevinrizza: This should be in a separate PR.

pkg/operatorsource/deleted.go
aravindhp: >This could introduce a bug where a user creates a CSC with the same name as an existing OpSrc.

This is actually a troublesome scenario that we have not addressed. Thanks for bringing that up, @awgreene. However addressing that is beyond the scope of this PR.

>If we want to address this edgecase, we could make the second call if-and-only-if the first call did not delete any resources.

Can't you decide which call to make based on the labels?

pkg/registry/registry.go
aravindhp: Fact of the matter is that the registry needs to know which label to populate the resources with. But I do agree the variable name is confusing? `forOpSrc` might be better but I don't think that will satisfy @kevinrizza.

>If so, it seems like that is a coincidental concern and we should be more respectful of the purpose of these objects and their composition.

What objects are you talking about here? The `registry` or the child objects?

pkg/status/status.go
aravindhp: +1. Moreover I fixed this in #205 since I was changing code there.

aravindhp:
Please prefix the commit and PR header with the BZ number and remove `[bug]`. That seems to imply you are introducing a bug ;-)

pkg/operatorsource/deleted.go
galletti94: Regardless of labels, what is the expected behavior in this case? If a `csc` were to be created with the same name as an existing `opsrc`, who should own the child resources? Should it be both (in which case an update to either the `csc` or the `opsrc` would affect theses child resources)?

aravindhp: I will repeat what I said... This is actually a troublesome scenario that we have not addressed. However addressing that is beyond the scope of this PR.

pkg/registry/registry.go
kevinrizza: I mean the `registry` object. I think the fact that we are setting a field that defines if "that" `registry` object is tagged to create resources from `opsrc`es or `csc`s is inherently bad. That sort of parameter is something that should be passed along as a function parameter, not tied to the object. The fact that those `registry` objects are initialized by each reconciler is an implementation detail, and one that could change.

galletti94:
@awgreene @kevinrizza @aravindhp @anik120 I addressed your comments. PTAL

pkg/builders/builders.go
```diff
@@ -19,3 +21,18 @@ const OpsrcOwnerNameLabel string = "opsrc-owner-name"
 // that are owned by the OperatorSource. When this label is set, the reconciler
 // should handle these resources when the OperatorSource is deleted.
 const OpsrcOwnerNamespaceLabel string = "opsrc-owner-namespace"
+
+// GetLabelMap returns a map with either the csc or opsrc owner name and
+// namespace labels depending what kind is the owner
+func GetLabelMap(name, namespace, owner string) map[string]string {
```
kevinrizza: I'm still wondering about why is it "opsrc or not". Why not also check for `v2.CatalogSourceConfigKind` and if it's neither just return an empty map?

pkg/grpccatalog/grpccatalog.go
```diff
@@ -93,12 +94,19 @@ func (r *GrpcCatalog) EnsureResources(key types.NamespacedName, displayName, pub
...
-		WithSpec(olm.SourceTypeGrpc, address, displayName, publisher)
-
+func newCatalogSource(labels map[string]string, key types.NamespacedName, displayName, publisher, namespace, address, owner string) *olm.CatalogSource {
+	var builder *builders.CatalogSourceBuilder
+	if owner == v1.OperatorSourceKind {
+		builder = new(builders.CatalogSourceBuilder).
+			WithOpsrcOwnerLabel(key.Name, namespace).
+			WithMeta(key.Name, key.Namespace).
+			WithSpec(olm.SourceTypeGrpc, address, displayName, publisher)
+	} else {
```
kevinrizza: Same comment as above.

pkg/builders/builders.go
aravindhp: You could use a switch statement for this.

pkg/builders/builders.go
```diff
@@ -19,3 +24,20 @@ const OpsrcOwnerNameLabel string = "opsrc-owner-name"
 // that are owned by the OperatorSource. When this label is set, the reconciler
 // should handle these resources when the OperatorSource is deleted.
 const OpsrcOwnerNamespaceLabel string = "opsrc-owner-namespace"
+
+// GetLabelMap returns a map with either the csc or opsrc owner name and
```
aravindhp: Please say `CatalogSourceConfig` and `OperatorSource` in comments.

pkg/builders/builders.go
```diff
@@ -1,14 +1,19 @@
 package builders
 
-// OwnerNameLabel is the label used to mark ownership over a given resources.
+import (
+	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v2"
+)
+
+// CscOwnerNameLabel is the label used to mark ownership over a given resources.
```
aravindhp: Given the two sets of constants, the description should be more explicit. `CscOwnerNameLabel is the label used to indicate the name of the CatalogSourceConfig that owns this resources` and so on. Please fix all the docs.

pkg/grpccatalog/grpccatalog.go
```diff
@@ -93,12 +94,19 @@ func (r *GrpcCatalog) EnsureResources(key types.NamespacedName, displayName, pub
...
+	if owner == v1.OperatorSourceKind {
+		builder = new(builders.CatalogSourceBuilder).
+			WithOpsrcOwnerLabel(key.Name, namespace).
+			WithMeta(key.Name, key.Namespace).
+			WithSpec(olm.SourceTypeGrpc, address, displayName, publisher)
+	} else if owner == v2.CatalogSourceConfigKind {
+		builder = new(builders.CatalogSourceBuilder).
+			WithCscOwnerLabel(key.Name, namespace).
+			WithMeta(key.Name, key.Namespace).
+			WithSpec(olm.SourceTypeGrpc, address, displayName, publisher)
```
aravindhp: You are defeating the purpose of a builder. Create the builder with the common elements and the `if else` should only set `WithOpsrcOwnerLabel` or  `WithCscOwnerLabel`

pkg/registry/registry.go
```diff
@@ -292,11 +295,17 @@ func (r *registry) getSubjects() []rbac.Subject {
...
-		WithOwnerLabel(r.key.Name, r.key.Namespace).
-		WithSpec(1, r.getLabel(), r.newPodTemplateSpec(registryCommand, needServiceAccount)).
-		Deployment()
+		WithSpec(1, r.getLabel(), r.newPodTemplateSpec(registryCommand, needServiceAccount))
+	if r.owner == v1.OperatorSourceKind {
+		return builder.WithOpsrcOwnerLabel(r.key.Name, r.key.Namespace).
+			Deployment()
+	} else if r.owner == v2.CatalogSourceConfigKind {
+		return builder.WithCscOwnerLabel(r.key.Name, r.key.Namespace).
+			Deployment()
```
aravindhp: You are again defeating the purpose of a builder. Just add the labels in the `if` statements.

This applies to every instance you are doing this.

pkg/operatorsource/configuring_test.go
```diff
@@ -155,7 +155,10 @@ func TestReconcile_NotConfigured_NewCatalogConfigSourceObjectCreated(t *testing.
 		builders.OpsrcOwnerNameLabel:      "foo",
 		builders.OpsrcOwnerNamespaceLabel: "marketplace",
 	}
-	opsrcIn.SetLabels(labelsWant)
+	labelsNeed := map[string]string{
+		"opsrc-group": "Community",
```
aravindhp: Why are you introducing this label?

galletti94: This is a label from a bit higher up https://github.com/operator-framework/operator-marketplace/pull/208/files/3d69194d66021d90a39618fd391a35ccedd95565..58cd8130be68800cfbddd24630df27330eeb5c2f#diff-0a815e1d2edb82030bf32b730e9aba4fR154

galletti94:
@aravindhp @kevinrizza @awgreene I addressed your comments. PTAL when you get a chance

pkg/builders/builders.go
```diff
@@ -1,21 +1,45 @@
...
+			OpsrcOwnerNameLabel:      name,
+			OpsrcOwnerNamespaceLabel: namespace,
+		}
+	case v2.CatalogSourceConfigKind:
+		return map[string]string{
+			CscOwnerNameLabel:      name,
+			CscOwnerNamespaceLabel: namespace,
+		}
+	default:
+		return nil
```
aravindhp: You should return an empty map here

pkg/operatorsource/configuring_test.go
```diff
@@ -225,13 +228,6 @@ func TestReconcile_CatalogSourceConfigAlreadyExists_Updated(t *testing.T) {
 	ctx := context.TODO()
 	opsrcIn := helperNewOperatorSourceWithPhase(namespace, name, phase.Configuring)
 
-	labelsWant := map[string]string{
-		"opsrc-group":                     "Community",
-		builders.OpsrcOwnerNameLabel:      "foo",
-		builders.OpsrcOwnerNamespaceLabel: "marketplace",
-	}
-	opsrcIn.SetLabels(labelsWant)
```
aravindhp: I see why you shouldn't be setting the object with the expected labels but I don't see where you are checking if the expected labels are present.

galletti94: Yeah it doesn't seem like I am checking the labels anywhere - good catch!

pkg/operatorsource/configuring_test.go
kevinrizza: @galletti94 Could you update this link? It's out of date because of a rebase I assume

kevinrizza:
Looks good aside from Aravindh's requests.

pkg/operatorsource/configuring_test.go
galletti94: After looking at this in more detail, I don't think this is possible for unit tests. Since these are just mock calls, child resources aren't actually getting created so how can we check that labels are correctly set?

aravindhp: This is not for the child resources but the object itself. I think you check the labels between `opsrcIn` and `opsrcGot`. Maybe that is why the labels were being set where you are removing it. So maybe you should not get rid of this code. The `assert.Equal()` might already be taking care of this check but please confirm.

galletti94: the `assert.Equal` here does check that the labels set before calling the reconciler are still there after the reconciler is called. This is orthogonal to the owner labels however. I will add the first label back but it seems misleading to add the second and third labels (lines 230 and 231).

In order to have unit tests for the owner labels I would need to add unit tests in the grpccatalog pkg and the registry pkg where the child resources are created. I would argue that this is redundant because the e2e tests now check for this label propagation. But I am happy to do this if the team thinks this is crucial.

pkg/builders/builders.go
```diff
@@ -1,21 +1,45 @@
...
-// that are owned by the OperatorSource. When this label is set, the reconciler
-// should handle these resources when the OperatorSource is deleted.
+// OpsrcOwnerNamespaceLabel is the label used to indicate the namespace of the OperatorSource
+// that owns this resources. When this label is set, the reconciler should handle these
+// resources when the OperatorSource is deleted.
 const OpsrcOwnerNamespaceLabel string = "opsrc-owner-namespace"
+
+// GetLabelMap returns a map with either the CatalogSourceConfig or the OperatorSource owner
+// name and namespace labels depending what kind is the owner
+func GetLabelMap(name, namespace, owner string) map[string]string {
```
aravindhp: Please rename this to `GetOwnerLabel()`. We shouldn't be using naming functions / variables after their types.

pkg/grpccatalog/grpccatalog.go
```diff
@@ -114,12 +118,9 @@ func newCatalogSource(labels map[string]string, key types.NamespacedName, displa
...
 
 // DeleteResources deletes a CatalogSource and all resources that make up a registry.
-func (r *GrpcCatalog) DeleteResources(ctx context.Context, name, namespace, targetNamespace string) (err error) {
+func (r *GrpcCatalog) DeleteResources(ctx context.Context, name, namespace, targetNamespace, owner string) (err error) {
 	allErrors := []error{}
-	labelMap := map[string]string{
-		builders.OwnerNameLabel:      name,
-		builders.OwnerNamespaceLabel: namespace,
-	}
+	labelMap := builders.GetLabelMap(name, namespace, owner)
```
aravindhp: Please rename to `ownerLabel` for the same reason

PR #211, opened 2019-06-28
Bug 1717509: [status] Populate RelatedObjects

aravindhp:
Problem:
The must-gather tool is not able to gather enough information about the marketplace operator.

Solution:
The must-gather tool requires a field, RelatedObjects, in the ClusterOperator CR to be populated with ObjectReferences of the resources associated with the operator. This PR populates this field with the operator's namespace and the OperatorSource/CatalogSourceConfig/CatalogSource resources in the same namespace.

Reason for inclusion in 4.1.z:
Without having the RelatedObjects in the ClusterOperator CR, the CEE team will not be able to use the must-gather tool to collect information about the marketplace operator from customer. This makes it hard for them to provide engineering with enough information about issues found on production clusters. Having this fix in 4.1 will resolve this issue 

(based on #186, #205)

### DIFF
--- a/test/testsuites/clusteroperatorstatustests.go
+++ b/test/testsuites/clusteroperatorstatustests.go
@@ -1,0 +1,56 @@
+package testsuites
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/operator-framework/operator-sdk/pkg/test"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+// ClusterOperatorStatusOnStartup is a test suite that ensures the ClusterOperator resource which
+// defines the status of the marketplace operator has the correct status upon initialization
+func ClusterOperatorStatusOnStartup(t *testing.T) {
+	ctx := test.NewTestCtx(t)
+	defer ctx.Cleanup()
+
+	// Get global framework variables.
+	client := test.Global.Client
+
+	// Get namespace
+	namespace, err := test.NewTestCtx(t).GetNamespace()
+	require.NoError(t, err, "Could not get namespace")
+
+	// Check that the ClusterOperator resource has the correct status
+	clusterOperatorName := "marketplace"
+	expectedTypeStatus := map[configv1.ClusterStatusConditionType]configv1.ConditionStatus{
+		configv1.OperatorProgressing: configv1.ConditionFalse,
+		configv1.OperatorAvailable:   configv1.ConditionTrue,
+		configv1.OperatorDegraded:    configv1.ConditionFalse}
+
+	// Poll to ensure ClusterOperator is present and has the correct status
+	// i.e. ConditionType has a ConditionStatus matching expectedTypeStatus
+	namespacedName := types.NamespacedName{Name: clusterOperatorName, Namespace: namespace}
+	result := &configv1.ClusterOperator{}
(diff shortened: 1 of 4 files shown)


aravindhp:
This PR should only after https://github.com/openshift/must-gather/pull/107/ merges

test/testsuites/clusteroperatorstatustests.go
```diff
@@ -0,0 +1,83 @@
...
+			if expectedTypeStatus[condition.Type] != condition.Status {
+				return false, fmt.Errorf("Expecting condition type %v of status %v but got %v", condition.Type, expectedTypeStatus[condition.Type], condition.Status)
+			}
+		}
+		return true, nil
+	})
+	assert.NoError(t, err, "ClusterOperator never reached expected status")
+
+	// Check if the expected default ObjectReferences are present in RelatedObjects
+	expectedRelatedObjects := []configv1.ObjectReference{
```
tkashem: if this is the same as https://github.com/operator-framework/operator-marketplace/pull/211/files#diff-a5666326aefdf43a8e89fd4871449fa0R269, then is there a way we can reuse this piece of initialization?

aravindhp: I don't want to do that on purpose as this is in the test case and I would rather not shared code between the main code and the tests. Similar argument as to why we don't use builders in the test cases.

tkashem: oops, didn't realize this is in the test code. it's good.

test/testsuites/clusteroperatorstatustests.go
```diff
@@ -0,0 +1,83 @@
...
+		for _, condition := range result.Status.Conditions {
+			if expectedTypeStatus[condition.Type] != condition.Status {
+				return false, fmt.Errorf("Expecting condition type %v of status %v but got %v", condition.Type, expectedTypeStatus[condition.Type], condition.Status)
+			}
+		}
+		return true, nil
+	})
+	assert.NoError(t, err, "ClusterOperator never reached expected status")
+
+	// Check if the expected default ObjectReferences are present in RelatedObjects
```
awgreene: This test seems to:
A. Check that the ClusterOperator is set to available.
B. Check that related ojbects are set.

Should this be separated into two tests?

aravindhp: This comment should have been made on #205. To make the change here would involve first making the change to master and then here. Given the benefits I would rather not.

aravindhp: Moreover even if it was split into two tests, it would be a lot of repeated code.

awgreene: Understood - no complains leaving as is.

derekwaynecarr:
i assume the hold should be removed.

the parent [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1717439) is VERIFIED.

/hold cancel

derekwaynecarr:
also lack ability to apply required label, will follow up on this tomorrow.

aravindhp:
> i assume the hold should be removed.
> 
> the parent [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1717439) is VERIFIED.

@derekwaynecarr, I had placed a hold waiting for https://github.com/openshift/must-gather/pull/107 to merge as this PR depends on that must-gather fix.

jwforres:
Yeah let's make sure openshift/must-gather#107 actually merges before we put this one in

/hold

just to make sure it doesn't go in ahead of the other one, but I expect we should be able to cancel the hold and cherry-pick this first thing tomorrow

aravindhp:
/hold cancel

@derekwaynecarr @jwforres openshift/must-gather#107 has merged. Please approve.

PR #212, opened 2019-07-01
WIP Bug 1721537: Fix degraded status on upgrade

awgreene:
Problem: When the Marketplace Operator was going through an upgrade,
the operator would report degraded as true when it was terminated
prior to the new operator coming up.

Cause: We had purposely programmed the operator to report that
had exited.

Solution: Rewrite the status reporting logic to provide meaningful
    data to the ClusterOperator. This rewrite includes:
    * Removing the tie between ClusterOperator Availability and the
    ratio of successful syncs to failed syncs. An invalid
    OperatorSource/CatalogSourceConfig and networks connectivity
    issues would push Marketplace towards the degraded state.
    Telemetry currently reports a number of Marketplace operators
    as degraded as a result.
    * Reporting that the ClusterOperator is available when 5
    CatalogSourceConfig or OperatorSource syncs have occurred.
    * Ensuring that the operator does not report degraded as true when the
    operator shuts down.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1721537


awgreene:
Please review @kevinrizza @aravindhp @galletti94

pkg/controller/catalogsourceconfig/catalogsourceconfig_controller.go
```diff
@@ -81,19 +81,11 @@ type ReconcileCatalogSourceConfig struct {
...
-		}
-	}()
+	operatorstatus.SendEventMessage(nil)
+
 	// Fetch the CatalogSourceConfig instance
 	instance := &v2.CatalogSourceConfig{}
-	err = r.client.Get(context.TODO(), request.NamespacedName, instance)
+	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
 	if err != nil {
 		if errors.IsNotFound(err) {
```
kevinrizza: What about if the get fails, and it's not from an `IsNotFound` error? Don't we want to increment the value in cluster status then?

kevinrizza: Same comment about both controllers btw

pkg/operatorstatus/monitor.go
```diff
@@ -0,0 +1,215 @@
...
+	// prior to reporting that it is available to the ClusterOperator.
+	minimumSyncs uint = 2
+
+	// degradedRatio is the ratio of failed transition to syncs that must be reached
+	// prior to reporting that the marketplace operator is  degraded.
+	degradedRatio = 0.7
+)
+
+var (
+	instance *statusMonitor
```
kevinrizza: Since this is a singleton, can we have a better name? `instance` doesn't make sense out of context. What about `statusMonitorSingleton` or `statusMon`?

pkg/operatorstatus/monitor.go
```diff
@@ -0,0 +1,215 @@
...
+			eventCh:      make(chan error, 32),
+		}
+	})
+
+	return nil
+}
+
+// StartReporting initializes the instance singleton and starts reporting
+// ClusterOperator status.
+func StartReporting() error {
```
kevinrizza: If this method returns an error shouldn't we try to report degraded? Maybe that's just counterintuitive.

pkg/operatorstatus/monitor.go
```diff
@@ -0,0 +1,215 @@
...
+		return fmt.Errorf("Error creating ClusterOpreator writer: %v", err)
+	}
+
+	// Initialize the instance singleton once.
+	once.Do(func() {
+		// init the instance singleton.
+		instance = &statusMonitor{
+			coWriter:     coWriter,
+			namespace:    namespace,
+			eventTracker: eventTracker{0, 0, 1000, 10, sync.Mutex{}},
```
kevinrizza: Initting this inline makes sense, but can we use the longform so it's clear what these values are? And can we comment on each of them so they aren't just magic numbers?

pkg/operatorstatus/monitor.go
awgreene: I like `statusMon`, we can put a comment about it being a singleton above it's declaration.

pkg/operatorstatus/monitor.go
awgreene: `StartReporting` will only generate an error if it encounters an error that prevents it from writing to the `ClusterOperator`, so we wouldn't be able to report degraded.

pkg/controller/catalogsourceconfig/catalogsourceconfig_controller.go
awgreene: I believe this is a good point. I will update the code to increment the number of failed transitions when an error other than `IsNotFound` is encountered during a `get` request in all places.

pkg/operatorstatus/builder.go
```diff
@@ -0,0 +1,190 @@
...
+		if b.status.Versions[i].Name == name {
+			return &b.status.Versions[i]
+		}
+	}
+
+	return nil
+}
+
+func (b *Builder) setRelatedObject(reference configv1.ObjectReference) {
+	for i := range b.status.RelatedObjects {
```
aravindhp: Is this really required? Without this check this function would be `setRelatedObject`

pkg/operatorstatus/builder.go
```diff
@@ -0,0 +1,190 @@
...
+	for i := range b.status.Versions {
+		if b.status.Versions[i].Name == name {
+			return &b.status.Versions[i]
+		}
+	}
+
+	return nil
+}
+
+func (b *Builder) setRelatedObject(reference configv1.ObjectReference) {
```
aravindhp: This is really not `setRelatedObject` but rather you are always ensuring that the related object is is present.

pkg/operatorstatus/builder.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// WithMarketplaceRelatedObjects populates RelatedObjects that are related to Marketplace.
+// RelatedObjects are consumed by https://github.com/openshift/must-gather
+func (b *Builder) WithMarketplaceRelatedObjects(namespace string) *Builder {
+	// Add the operator's namespace which will result in core resources being gathered
+	b.WithRelatedObject("", "namespaces", "", namespace)
+
+	// Add the non-core resources we care about
+	b.WithRelatedObject(v1.SchemeGroupVersion.Group, v1.OperatorSourceKind, namespace, "").
```
aravindhp: Please do a test with a private `OperatorSource` and ensure that other core resources created are being added. I remember testing this very early on but we should confirm this.

pkg/operatorstatus/builder.go
```diff
@@ -0,0 +1,190 @@
...
+		Namespace: namespace,
+		Name:      name,
+	}
+
+	b.setRelatedObject(reference)
+
+	return b
+}
+
+func (b *Builder) init() {
```
aravindhp: This is sort of an anti-pattern when you are writing a builder. I suggest rethinking this.

pkg/operatorstatus/builder.go
```diff
@@ -0,0 +1,190 @@
...
+	if b.status == nil {
+		b.status = &configv1.ClusterOperatorStatus{
+			Conditions:     []configv1.ClusterOperatorStatusCondition{},
+			Versions:       []configv1.OperandVersion{},
+			RelatedObjects: []configv1.ObjectReference{},
+		}
+	}
+}
+
+func (b *Builder) findCondition(conditionType configv1.ClusterStatusConditionType) *configv1.ClusterOperatorStatusCondition {
```
aravindhp: Why not use the [helpers](https://github.com/openshift/library-go/blob/master/pkg/config/clusteroperator/v1helpers/status.go)?

pkg/operatorstatus/builder.go
```diff
@@ -0,0 +1,190 @@
...
+	for i := range b.status.Conditions {
+		if b.status.Conditions[i].Type == conditionType {
+			return &b.status.Conditions[i]
+		}
+	}
+
+	return nil
+}
+
+func (b *Builder) setCondition(condition *configv1.ClusterOperatorStatusCondition) {
```
aravindhp: Why not use the [helpers](https://github.com/openshift/library-go/blob/master/pkg/config/clusteroperator/v1helpers/status.go)?

pkg/operatorstatus/eventtracker.go
```diff
@@ -0,0 +1,61 @@
...
+	e.preventOverflow()
+}
+
+// incrementFailedTransitions will increment the failedTranisitions field by one.
+func (e *eventTracker) incrementFailedTransitions() {
+	e.lock.Lock()
+	defer func() {
+		e.lock.Unlock()
+	}()
+	e.failedTransitions++
```
aravindhp: Shouldn't you prevent an overflow here too?

pkg/operatorstatus/eventtracker.go
```diff
@@ -0,0 +1,61 @@
...
+	e.lock.Lock()
+	defer func() {
+		e.lock.Unlock()
+	}()
+	return e.syncs, e.failedTransitions
+}
+
+// preventOverflow prevents the sync and failedTransition fields from overflowing
+// in long-running marketplace instances.
+func (e *eventTracker) preventOverflow() {
```
aravindhp: You should add a comment about requiring the lock to be held before calling.

cmd/manager/main.go
```diff
@@ -51,6 +51,12 @@ func main() {
 		log.Fatalf("failed to get watch namespace: %v", err)
 	}
 
+	// Report that the operator is progressing towards a new version.
```
aravindhp: Is this really what you are doing here? The function is doing way for than that.

pkg/operatorstatus/monitor.go
kevinrizza: Yeah, I figured that was the case. I just wanted to confirm.

awgreene: Marking as resolved.

cmd/manager/main.go
```diff
@@ -51,6 +51,12 @@ func main() {
...
+	err = operatorstatus.StartReporting()
```
aravindhp: This should only kick off if the API is present. There is no reason for these threads on vanilla kube. But you do need the check in `SendEventMessage()` either way.

pkg/operatorstatus/monitor.go
```diff
@@ -0,0 +1,215 @@
...
+		log.Warning("[status] Event channel is busy, not reporting event")
+	}
+}
+
+// monitorEvents will update the ClusterOpreator to reflect that Marketplace
+// is available or degraded based on the events reported.
+func (s *statusMonitor) monitorEvents() {
+	// reportInterval represents the frequency at which marketplace
+	// will attempt to update the ClusterOperator.
+	reportInterval := 1 * time.Minute
```
aravindhp: This should be a constant

pkg/operatorstatus/monitor.go
```diff
@@ -0,0 +1,215 @@
...
+			}
+
+			// Check if the ratio is appropriate.
+			if syncs > 0 && syncs >= minimumSyncs {
+				if float64(failures)/float64(syncs) < degradedRatio {
+					statusErr = s.reportAvailable()
+				} else {
+					statusErr = s.reportDegraded()
+				}
+				reportInterval = 5 * time.Minute
```
aravindhp: Constant again.

pkg/operatorstatus/writer.go
```diff
@@ -0,0 +1,101 @@
...
+
+	return &writer{
+		client:    client,
+		discovery: k8sInterface.Discovery(),
+	}, nil
+}
+
+// IsAPIAvailable return true if cluster operator API is present on the cluster.
+// Otherwise, exists is set to false.
+func (w *writer) isAPIAvailable() (exists bool, err error) {
```
aravindhp: This should be done outside of this context. The writer should exist only if the API is present.

cmd/manager/main.go
```diff
@@ -50,6 +50,20 @@ func main() {
...
+	}
+
+	// Report that the operator is progressing towards a new version.
+	err = operatorstatus.ReportProgressing(coWriter, namespace)
+	if err != nil {
+		log.Error(err)
+	}
+
+	// Start reporting status as a result of the number processed of syncs.
+	go operatorstatus.MonitorSyncs(coWriter, namespace)
```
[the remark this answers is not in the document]
awgreene: We could - but OpenShift is supposed to be self-recovering. If the CO API is not available when the operator starts, Marketplace should start reporting when it becomes available.

pkg/operatorstatus/monitor.go
awgreene: If this is a constant we won't be able to change the interval at which we update the `ClusterOperator`.

awgreene: Rather than update the interval, I could replace the code where the value is updated with a `sleep`.

pkg/operatorstatus/monitor.go
awgreene: Will update based on feedback from [this thread](https://github.com/operator-framework/operator-marketplace/pull/212#discussion_r301708497)

pkg/operatorstatus/writer.go
awgreene: This would again prevent us from recovering from a missing `CO`.

pkg/operatorstatus/eventtracker.go
awgreene: The number of syncs are always incremented and so is guaranteed to be equal to or greater than the count of failed transitions. Both are currently reduced when the limit is reached by syncs.

pkg/operatorstatus/builder.go
awgreene: Will do.

cmd/manager/main.go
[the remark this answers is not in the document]
aravindhp: Sure it can be self-recovering but there are ways to figure out if you are on an OpenShift vs a vanilla kube cluster like checking for the `ClusterVersion` API. Moreover I don't see how the marketplace operator comes up before CVO that owns this CRD.

pkg/operatorstatus/monitor.go
aravindhp: > If this is a constant we won't be able to change the interval at which we update the `ClusterOperator`.

`1 * time.Minute` is a constant

>Rather than update the interval, I could replace the code where the value is updated with a sleep

You should use a sleep to avoid busy polling but either way the interval should increase.

pkg/operatorstatus/eventtracker.go
aravindhp: This warrants a comment

cmd/manager/main.go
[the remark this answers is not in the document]
aravindhp: Or you make a command line option that you pass to the operator that indicates whether you report status or not.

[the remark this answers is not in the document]
aravindhp: And this option will only be present in the OpenShift deployment file and not in the upstream one.

PR #214, opened 2019-07-09
Bug 1721537: Fix degraded status on upgrade

awgreene:
Bug 1721537: Fix Degraded status on upgrade

Problem: There are three bugs that should be addressed within this bugzilla:
1. During an upgrade, multiple marketplace operators are running and updating the ClusterOperatorStatus at the same time, making it difficult to identify the actual state of the marketplace operator.
2. The marketplace operator is reporting degraded while it is upgrading. The marketplace operator reports degraded when the number of failed syncs surpases some threshold of total syncs. The marketplace operator currently reports a failed sync whenever an error is encountered while reconciling an operand (OperatorSources or CatalogSourceConfigs). This allows network issues or invalid operands to drive the marketplace operator to a degraded state.
3. It is difficult to identify why a ClusterOperatorStatus condition is in a given state via telemetry. Marketplace currently does not set a reason when setting conditions and telemetry only includes the state and reason.

Solution: Three changes need to be made to address the bugzilla:
1. Implement leader election using the functions provided by the Operator-SDK to prevent multiple marketplace operators from updating the ClusterOperatorStatus at the same time.
2. Rather than reporting a failed sync whenever an error is encountered while reconciling an operand, report a failed sync when the marketplace operator is unable to get or update an existing operand.
3. Update marketplace to include a reason when setting a condition that indicates why the condition is being set.

### DIFF
--- a/pkg/catalogsourceconfig/handler.go
+++ b/pkg/catalogsourceconfig/handler.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/operator-framework/operator-marketplace/pkg/datastore"
 	"github.com/operator-framework/operator-marketplace/pkg/phase"
+	operatorstatus "github.com/operator-framework/operator-marketplace/pkg/status"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
@@ -72,6 +73,9 @@ func (h *catalogsourceconfighandler) Handle(ctx context.Context, in *v2.CatalogS
 	// case, we need to update the modified CatalogSourceConfig object.
 	if updateErr := h.client.Update(ctx, out); updateErr != nil {
 		log.Errorf("Failed to update object - %v", updateErr)
+
+		// Error updating the object - report a failed sync.
+		operatorstatus.SendSyncMessage(updateErr)
 
 		if err == nil {
 			// No reconciliation err, but update of object has failed!
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -331,9 +331,8 @@ func (s *status) monitorClusterStatus() {
 	for {
 		select {
 		case <-s.stopCh:
-			// If the stopCh is closed, the operator will exit and CO should
-			// be set to degraded.
-			msg := "The operator has exited and is no longer reporting status."
+			// If the stopCh is closed, set all ClusterOperatorStatus conditions to false.
+			msg := "The operator has exited"
 			conditionListBuilder := clusterStatusListBuilder()
 			conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionFalse, msg)
 			conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionFalse, msg)
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -332,11 +332,12 @@ func (s *status) monitorClusterStatus() {
 		select {
 		case <-s.stopCh:
 			// If the stopCh is closed, set all ClusterOperatorStatus conditions to false.
+			reason := "OperatorExited"
 			msg := "The operator has exited"
 			conditionListBuilder := clusterStatusListBuilder()
-			conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionFalse, msg)
-			conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionFalse, msg)
-			statusConditions := conditionListBuilder(configv1.OperatorDegraded, configv1.ConditionFalse, msg)
+			conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionFalse, msg, reason)
+			conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionFalse, msg, reason)
+			statusConditions := conditionListBuilder(configv1.OperatorDegraded, configv1.ConditionFalse, msg, reason)
 			statusErr := s.setStatus(statusConditions)
 			if statusErr != nil {
 				log.Error("[status] " + statusErr.Error())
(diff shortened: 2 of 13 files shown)


pkg/status/status.go
```diff
@@ -324,10 +319,6 @@ func (s *status) syncChannelReceiver() {
 // monitorClusterStatus updates the ClusterOperator's status based on
 // the number of successful syncs / total syncs
 func (s *status) monitorClusterStatus() {
-	// Signal to the main channel that we have stopped reporting status.
-	defer func() {
-		close(s.monitorDoneCh)
-	}()
 	for {
 		select {
 		case <-s.stopCh:
```
aravindhp: Is the comment below accurate?

kevinrizza:
> * Please mention **how** you are not reporting degraded on exit. I for one don't fully understand how you are fixing this issue.
> * What happens if `degraded` is true because it was unable to reach Quay and the operator exits due to an upgrade? In that case, we still will have the scenario where we are reporting `degraded` is true during the upgrade.

Are we saying that we just never want to report degraded? I don't see how we can both have a more granular method of updating status like that *and* keep the PR small so it can make it into a Z stream.

awgreene:
@aravindhp

Thanks for the review, I have addressed the feedback.

> Please mention how you are not reporting degraded on exit. I for one don't fully understand how you are fixing this issue.

I made a mistake and did not push the intended commit - sorry about that. I have added the proper changes and updated the commit message.

>What happens if degraded is true because it was unable to reach Quay and the operator exits due to an upgrade? In that case, we still will have the scenario where we are reporting degraded is true during the upgrade.

With the latest update to the PR this should not happen.

@kevinrizza 
> Are we saying that we just never want to report degraded? I don't see how we can both have a more granular method of updating status like that and keep the PR small so it can make it into a Z stream.

With the latest update we should be still be able to report degraded. When the new marketplace-operator is elected leader, the status will be updated to:
```
progressing = true
degraded = false
available = false
```

pkg/status/status.go
awgreene: It is not, this is addressed in the latest commit.

awgreene:
Fixed a spelling error in the commit message with the latest push.

aravindhp:
>Are we saying that we just never want to report degraded? I don't see how we can both have a more granular method of updating status like that and keep the PR small so it can make it into a Z stream.

We will still report degraded if Quay is unreachable. Which is why I don't understand what exactly this PR is fixing. But say we are focusing on just fixing the state not being degraded during an upgrade, I think we need to account for the scenario where the operator is degraded i.e. shouldn't some reset happen at some point?

awgreene:
> We will still report degraded if Quay is unreachable. Which is why I don't understand what exactly this PR is fixing. But say we are focusing on just fixing the state not being degraded during an upgrade, I think we need to account for the scenario where the operator is degraded i.e. shouldn't some reset happen at some point?

The bugzilla that this PR was created to address highlights that:
1. Two marketplace operators are updating the status at one time.
2. That the status is updated to reflect that the operator is exited.

I agree the the two concerns you've highlighted should be addressed, but I don't believe that they should be addressed in this bug.

awgreene:
@kevinrizza @aravindhp I've split the PR into 3 pieces based on our in person discussion.

pkg/catalogsourceconfig/handler.go
```diff
@@ -73,6 +74,9 @@ func (h *catalogsourceconfighandler) Handle(ctx context.Context, in *v2.CatalogS
 	if updateErr := h.client.Update(ctx, out); updateErr != nil {
 		log.Errorf("Failed to update object - %v", updateErr)
 
+		// Error updating the object - report a failed sync.
+		operatorstatus.SendSyncMessage(updateErr)
```
aravindhp: Thinking a bit more about this... should we just send the sync message on every error here? i.e. move the send to call sites of `Handle()`.

pkg/status/status.go
```diff
@@ -331,9 +331,8 @@ func (s *status) monitorClusterStatus() {
...
-			// If the stopCh is closed, the operator will exit and CO should
-			// be set to degraded.
-			msg := "The operator has exited and is no longer reporting status."
+			// If the stopCh is closed, set all ClusterOperatorStatus conditions to false.
+			msg := "The running operator has exited, CVO will create a new operator."
```
aravindhp: "The operator has exited" is sufficient here.

pkg/catalogsourceconfig/handler.go
awgreene: I don't believe we would want to do that.

An [error from the reconcile](https://github.com/operator-framework/operator-marketplace/pull/214/files/0a5750740949bcd69a20cef63aed2bcaa96eca0f..af6316a7c9c8fb71dd36d6b33dfa534b6108617d#diff-f46f13415712030379e27befaa786593R57) call would increment the number of failed syncs - which is more or less what we do today.

awgreene:
@aravindhp @kevinrizza I have addressed existing comments.

awgreene:
I want to review the e2e upgrade test.

/hold

awgreene:
@kevinrizza @aravindhp I have added reasons the conditions that we report as requested offline, please review.

pkg/status/status.go
```diff
@@ -332,11 +332,12 @@ func (s *status) monitorClusterStatus() {
 		select {
 		case <-s.stopCh:
 			// If the stopCh is closed, set all ClusterOperatorStatus conditions to false.
+			reason := "operatorExited"
```
aravindhp: Camel case is `OperatorExited`. Please fix everywhere.

pkg/status/status.go
```diff
@@ -356,11 +357,12 @@ func (s *status) monitorClusterStatus() {
 			// Create the ClusterOperator in the progressing state if it does not exist
 			// or if it is the first report.
 			if s.clusterOperator == nil {
+				reason := "operatorProgressing"
```
aravindhp: Can't you specify the version you are progressing towards here?

pkg/status/status.go
```diff
@@ -374,9 +376,10 @@ func (s *status) monitorClusterStatus() {
 
 			// Report that marketplace is available after meeting minimal syncs.
 			if cohelpers.IsStatusConditionFalse(s.clusterOperator.Status.Conditions, configv1.OperatorAvailable) {
+				reason := "operatorAvailable"
```
aravindhp: Same comment here about the version.

pkg/status/status.go
awgreene: Will do.

pkg/status/status.go
awgreene: We can add the version if we believe it adds value.

I believed that it did not add value after looking at what information is available within telemetry. Telemetry provides the current version and the target version that the cluster is progressing to - so I thought repeating this information in the reason was a no-value-add. If you disagree, we can update this message to reflect the version, such as `OperatorProgressingTo4.2.0-0.ci-2019-07-10-180724`.

pkg/status/status.go
awgreene: Same comment as above regarding adding version to the reason.

proposed `Reason` if we add the version:
`OperatorProgressedTo4.2.0-0.ci-2019-07-10-180724`.

pkg/status/status.go
aravindhp: Given that telemetry provides the info, I am fine leaving it out. If **possible** please provide more context in each case.

pkg/status/status.go
awgreene: Interesting side note, the common examples [provided via wikipedia ](https://en.wikipedia.org/wiki/Camel_case) include `iPhone` and `eBay`, and the format that starts with a capital letter is typically called `UpperCamelCase` or `PascalCase`.

Existing reasons in `ClusterOperatorStatus Conditions` other Reason are using `PascalCase` so I suppose we should too.

aravindhp: Interesting indeed. Lets be consistent with existing reasons.

awgreene:
Should be good to go @aravindhp @kevinrizza

PR #215, opened 2019-07-09
Bug 1727023: [status] Add upgradeable condition

awgreene:
Problem: The Marketplace ClusterOperator currently does not report
the Upgradeable ClusterOperatorStatus condition.

Solution: When the Marketplace Operator is Progressing, report
Upgradeable as false. In all other cases, report Upgradeable as true.

### DIFF
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -47,6 +47,10 @@ const (
 
 	// coStatusReportInterval is the interval at which the ClusterOperator status is updated
 	coStatusReportInterval = 20 * time.Second
+
+	// Marketplace is always upgradeable and should include this message in the Upgradeable
+	// ClusterOperatorStatus condition.
+	upgradeableMessage = "Marketplace is upgradeable"
 )
 
 // status will be a singleton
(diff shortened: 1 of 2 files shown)


awgreene:
This should be merged after #214

pkg/status/status.go
```diff
@@ -359,6 +359,7 @@ func (s *status) monitorClusterStatus() {
 			if s.clusterOperator == nil {
 				conditionListBuilder := clusterStatusListBuilder()
 				conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionTrue, fmt.Sprintf("Progressing towards release version: %s", s.version))
+				conditionListBuilder(configv1.OperatorUpgradeable, configv1.ConditionFalse, "Marketplace is not upgradeable")
```
aravindhp: Why is the operator not upgradeable in this scenario? I can't think of a scenario where we are not upgradeable. Moreover if this condition is false, does CVO wait until it becomes true to attempt an upgrade?

awgreene: If the Operator is progressing to a new version while performing a migration we would not want to report that we are available for an upgrade.

aravindhp: That should be reflected in #207 and not here.

awgreene: That's fair, I will update the PR.

awgreene:
@aravindhp @kevinrizza this PR is ready for review.

awgreene:
I had forgotten to update the e2e-test, addressed in the updated commit.

PR #216, opened 2019-07-12
Bug 1696726: Don't replace pod spec on updates

aravindhp:
**Problem**
Customers would like to add nodeSelectors to the registry pods. However we replace the pod template spec in the registry deployment on updates causing such additions to be lost.

**Solution**
Instead of replacing the pod template spec only update the required fields like command service accounts.

### DIFF
--- a/test/testsuites/opsrccreationtests.go
+++ b/test/testsuites/opsrccreationtests.go
@@ -1,6 +1,7 @@
 package testsuites
 
 import (
+	"context"
 	"fmt"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


test/testsuites/opsrccreationtests.go
```diff
@@ -48,3 +53,67 @@ func testOperatorSourceGeneratesExpectedObjects(t *testing.T) {
...
+
+	// Get the test OperatorSource
+	testOpSrc := &v1.OperatorSource{}
+	namespacedName := types.NamespacedName{Name: helpers.TestOperatorSourceName, Namespace: namespace}
+	err = client.Get(context.TODO(), namespacedName, testOpSrc)
+	require.NoError(t, err)
+
+	// Force an update
+	testOpSrc.Status = v1.OperatorSourceStatus{}
+	err = client.Update(context.TODO(), deployment)
```
kevinrizza: Is this the best way to force an update? Won't this start to fail if we add validation on the status?

Can we just scale the marketplace deployment down to zero and then back up to one instead? Also, that seems like a good utility function to add in general so other tests can reuse it.

aravindhp: >Is this the best way to force an update? Won't this start to fail if we add validation on the status?

I agree that it not and we will need to update it with whatever new method we provide to force updates on `OperatorSources`

>Can we just scale the marketplace deployment down to zero and then back up to one instead? Also, that seems like a good utility function to add in general so other tests can reuse it.

I want to target the update to affect just one `OperatorSource` and not restart the operator causing all the objects to be reconciled again.

kevinrizza: Fair. Let's address it if/when we have that problem. We will probably need to define a method to update just one opsrc at that point anyway.

dariodsa:
I find this merge strange. If someone wants to specify ``node-selector``, that should be done from either ``CatalogSource`` or ``OperatorSource`` object. Now when node is drained, all those spec properties are lost.

PR #217, opened 2019-07-12
Bug 1700100: Stop scaling down registry deployment

aravindhp:
**Problem** 
Currently the operator scales down and then scales up the registry deployment, to cause it to forcefuly relaunch the pod so that updates from Quay are always picked up even when packages are not added or removed. This causes issues when we pick up a bad update that results in a crash looping pod resulting in an invalid registry for the CatalogSource.

**Solution**
Rather scaling down and up the deployment, update an annotation in the pod template spec which will also force the update.

**Note:** This PR alone won't fix BZ1700100. A fix to the e2e test that doesn't fail it on crash looping pods in the openshift-marketplace namespace is required.

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -2,6 +2,7 @@ package registry
 
 import (
 	"context"
+	"fmt"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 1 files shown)


awgreene:
Hello @aravindhp, the code implements the solution you described, but it is not obvious to me why scaling the deployment down and up can introduce a crash looping bug while adding an annotation does not. Could you please elaborate?

aravindhp:
> Hello @aravindhp, the code implements the solution you described, but it is not obvious to me why scaling the deployment down and up can introduce a crash looping bug while adding an annotation does not. Could you please elaborate?

Scaling down the deployment means we are going against the default rolling update [strategy](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy). Adding / updating the annotation will help maintaining that.

PR #220, opened 2019-07-15
Status

awgreene:
[status] CO status improvements

This PR introduces a number of small changes to the ClusterOperator Status reporting:

1. Currently the marketplace operator reports the status every twenty seconds. Once marketplace becomes available, there is no longer a need to update the status that frequently. This PR introduces a change so the marketplace operator only reports status once every 5 minutes once starts processing syncs.

2. If the marketplace ClusterOperator is deleted after the marketplace operator has deployed, the available and progressing conditions may be missing. This commit introduces a change that updates all status conditions once the marketplace starts processing syncs.

3. The status package currently has a defer statement that is called within a loop that runs for the life of the operator which can introduce memory issues. This PR introduces a change to remove the defer from the status package.


awgreene:
Closed due to code freeze on Friday.

PR #221, opened 2019-07-16
[proxy] Update marketplace to use a proxy

awgreene:
This commit introduces a change to enable marketplace to use a globally
configured proxy for its HTTP and HTTPS requests. To enable this
feature, the following changes were made:
* When the marketplace operator starts, a check will be made to
  identify if the Proxy API is available. If the API is available,
  marketplace will watch for changes to the "cluster" proxy. The
  cluster proxy is used to provide global proxy information.
* The CatalogSourceConfig and OperatorSource controllers have had a
  watch added that will alert them when a proxy is created, deleted,
  or destroyed. When the event comes from the cluster proxy,
  the controller will check if any child Resource deployments are not
  in sync with the environment variables defined by the cluster proxy.
* The marketplace operator will be configured to reach out to quay via
  the proxy.


pkg/proxy/envvar.go
```diff
@@ -0,0 +1,88 @@
...
+// Remove removes the given index from the list
+func Remove(list []corev1.EnvVar, index int) []corev1.EnvVar {
+	return append(list[:index], list[index+1:]...)
+}
+
+// Manage will ensure that the list of EnvVars are in sync with the proxy variables.
+// If the proxy variables are not set but are present in the list, they will be removed from the list.
+// If the proxy variables are set but missing from the list, they will be added to the list.
+// If the proxy variables are set and present in the list, their values will be updated in the list.
+func Manage(list []corev1.EnvVar) []corev1.EnvVar {
```
aravindhp: I understand this was introduced so that you can ensure that you don't remove any user added environment variables plus you will only add the proxy variables only when required.. I am not sure if we want the user to be able to add their own environment variables given the `operator-registry` application is what is run by the deployment. Also given the list management you are doing here it might be preferable to always have the proxy environment variables in the spec even when they are empty.

In other words this could just be `SetEnvVars()`

pkg/proxy/helpers.go
```diff
@@ -0,0 +1,40 @@
...
+
+var (
+	// ClusterProxyKey is the key for the cluster proxy.
+	ClusterProxyKey = types.NamespacedName{Name: ClusterProxyName}
+)
+
+// NeedsUpdate return true if the deployment is out of sync with the proxy environment variables.
+func NeedsUpdate(registryDeployment apps.Deployment) bool {
+	// Get the list of environment variables defined in the registry deployment
+	actualEnvVars := registryDeployment.Spec.Template.Spec.Containers[0].Env
```
aravindhp: Please check if the list is empty and return true if it is empty.

pkg/registry/registry.go
```diff
@@ -137,6 +138,9 @@ func (r *registry) ensureDeployment(appRegistries []string, needServiceAccount b
 			if needServiceAccount {
 				deployment.Spec.Template.Spec.ServiceAccountName = r.key.Name
 			}
+			// Add, Update, or Remove the proxy env vars based on their values in the marketplace pod.
+			deployment.Spec.Template.Spec.Containers[0].Env = proxy.Manage(deployment.Spec.Template.Spec.Containers[0].Env)
```
aravindhp: This would become `SetEnvVars()`

pkg/watches/helpers.go
```diff
@@ -171,3 +175,75 @@ func getCscOwnerKey(labels map[string]string) *client.ObjectKey {
...
+	pred := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			if e.Meta.GetName() == proxy.ClusterProxyName {
+				return true
+			}
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// If the cluster operator is ever changed we should kick off an event.
+			if e.MetaOld.GetName() == proxy.ClusterProxyName || e.MetaNew.GetName() == proxy.ClusterProxyName {
```
aravindhp: This check implies that the name changed which is not possible for kube resources. Checking either should be sufficient.

pkg/watches/helpers.go
```diff
@@ -171,3 +175,75 @@ func getCscOwnerKey(labels map[string]string) *client.ObjectKey {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// CheckProxyResource returns true if the deployment with the given namespacedname
+// needs to have its environment variables update due to a change in the cluster proxy.
+func CheckProxyResource(client cl.Client, name, namespace string) (bool, error) {
```
aravindhp: This is a proxy helper and should live in the proxy package.

pkg/watches/proxy_csc_mapper.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+type proxyToCatalogSourceConfigs struct {
+	client client.Client
+}
+
+// Map will update the proxy environment variables and create a reconcile request for
+// all CatalogSourceConfig.
+func (m *proxyToCatalogSourceConfigs) Map(obj handler.MapObject) []reconcile.Request {
+	// Ensure that proxy environment variables are in sync.
```
aravindhp: Will help if you mention this is for the operator and you are using that to track if changes are required for the individual requests.

test/e2e/main_test.go
```diff
@@ -17,6 +17,7 @@ import (
 )
 
 var coAPIPresent = true
+var ProxyAPIPresent = true
```
aravindhp: Why do you need this variable?

test/e2e/main_test.go
```diff
@@ -29,6 +30,8 @@ func TestMarketplace(t *testing.T) {
 	// Run Test Groups
 	if coAPIPresent {
 		t.Run("cluster-operator-status-test-group", testgroups.ClusterOperatorTestGroup)
+		// If CO API is available, assume proxy api is as well.
```
aravindhp: Then do you really need this variable. You might as well treat `coAPIPresent` as `onOpenShiftCluster`

test/testsuites/proxytests.go
```diff
@@ -0,0 +1,67 @@
...
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// ProxyTests is a test suite that ensure that the watches for proxy resources
+// are firing correctly and the deployment resources are updated as a result.
+func ProxyTests(t *testing.T) {
+	t.Run("no-proxy", testSetNoProxyVariable)
+}
+
+// testSetNoProxyVariable creates an OperatorSource and then set the cluster proxy.Spec.NoProxy
```
aravindhp: Where are you creating the `OperatorSource` in this test case?

test/testsuites/proxytests.go
```diff
@@ -0,0 +1,67 @@
...
+		} else {
+			assert.Equal(t, v.Value, "")
+		}
+	}
+
+	// Remove the NoProxy value in the proxy spec.
+	clusterProxy.Spec.NoProxy = ""
+	err = client.Update(context.TODO(), clusterProxy)
+	require.NoError(t, err, "Could not update the clusterProxy")
+}
```
aravindhp: For sake of completeness you might want to do the same test for `CatalogSourceConfigs` as you are adding watches there too. Lot of this code can be shared between the two test cases similar to what I did with watches for child resources.

deploy/upstream/05_role.yaml
```diff
@@ -3,6 +3,14 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: marketplace-operator
 rules:
+- apiGroups:
+  - config.openshift.io
```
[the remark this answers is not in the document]
awgreene: This is a good catch.

cmd/manager/main.go
```diff
@@ -84,6 +98,11 @@ func main() {
 		exit(err)
 	}
 
+	// Add proxy resource to scheme
+	if err := apiconfigv1.AddToScheme(mgr.GetScheme()); err != nil {
+		exit(err)
```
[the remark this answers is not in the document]
awgreene: This does not fail on Kube, I have tested it.

cmd/manager/main.go
```diff
@@ -59,6 +62,17 @@ func main() {
...
 
+	// Set proxy API availability
+	k8sInterface, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	err = proxy.SetProxyAvailability(k8sInterface.Discovery())
+	if err != nil {
+		log.Fatal(err)
```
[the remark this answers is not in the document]
awgreene: This method will only attempt to set proxy availability if the Proxy API exists on the cluster. Since CVO is managing the cluster proxy and marketplace, the operator will eventually resolve.

pkg/appregistry/appregistry.go
```diff
@@ -40,7 +42,17 @@ func (f *factory) New(options Options) (Client, error) {
 		return nil, err
 	}
 
-	transport := httptransport.New(u.Host, u.Path, []string{u.Scheme})
+	// A custom HTTPClient is used here since the default HTTPClients ProxyFromEnvironment
+	// uses a cache which won't let us update the proxy env vars.
+	cli := &http.Client{
+		Transport: &http.Transport{
+			Proxy: func(req *http.Request) (*url.URL, error) {
+				return httpproxy.FromEnvironment().ProxyFunc()(req.URL)
```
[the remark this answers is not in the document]
awgreene: This will work if the proxy variables are not set in the environment.

cmd/manager/main.go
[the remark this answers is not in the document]
shawn-hurley: if you can't get a config it makes sense to me to error hard, we wouldn't be able to contact k8s and I bet that `manager.New` would also fail and you would log.Fatal on line 82. 

The question for setting the proxy availability, I think is more nuanced, what are the error scenarios for this and does it make sense, that if something were to go wrong you would want the operator to continue.

pkg/appregistry/appregistry.go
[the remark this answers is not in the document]
awgreene: This does not return an error, we are setting the Proxy func: https://github.com/golang/go/blob/master/src/net/http/transport.go#L112-L121

pkg/catalogsourceconfig/succeeded.go
```diff
@@ -56,9 +61,23 @@ func (r *succeededReconciler) Reconcile(ctx context.Context, in *v2.CatalogSourc
...
 	}
 
-	r.log.Info(msg)
+	if needsProxyUpdate {
+		// The Environment Variables in the deployment created by the CatalogSourceConfig
+		// are out of sync with those in the Cluster Proxy.
+		// Drop the existing Status field so that reconciliation can start anew.
+		out.Status = v2.CatalogSourceConfigStatus{}
+		nextPhase = phase.GetNext(phase.Configuring)
+		msg = "Proxy environment variables not in sync, scheduling for configuring"
```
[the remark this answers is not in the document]
awgreene: That is a great idea and will be implemented in the next update to the PR. Perhaps we could add it to the `watches` package so it can be shared between the `CSC` and `opsrc` packages? Otherwise I will make a function in this file for each.

pkg/appregistry/appregistry.go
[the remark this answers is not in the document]
shawn-hurley: https://godoc.org/net/http#ProxyFromEnvironment

You could use that and make it more clear.

Also, this is how the default client works, so this might not even be needed

pkg/proxy/available.go
```diff
@@ -0,0 +1,47 @@
...
+
+// SetProxyAvailability will set isAPIAvailable to the correct value if
+// no unexpected errors are encountered.
+func SetProxyAvailability(discovery apidiscovery.DiscoveryInterface) error {
+	if discovery == nil {
+		return errors.New("discovery interface can not be <nil>")
+	}
+
+	opStatusGV := schema.GroupVersion{
+		Group:   "config.openshift.io",
```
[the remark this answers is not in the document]
awgreene: Good idea, I will use the `GroupName` and `GroupVersion.Version` defined here: https://github.com/openshift/api/blob/master/config/v1/register.go#L10-L11

pkg/proxy/envvar.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"os"
+
+	apiconfigv1 "github.com/openshift/api/config/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	// HTTP_PROXY is the URL of the proxy for HTTP requests.
+	envHTTPProxyName = "HTTP_PROXY"
```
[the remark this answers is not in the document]
awgreene: Understood, I will update the variables accordingly.

pkg/watches/helpers.go
awgreene: This is good to know, thank you Aravindh.

test/testsuites/proxytests.go
awgreene: This test is ran as part of the `OperatorSource` TestGroup, which creates an `OperatorSource` named `test-operators` prior to running this test.

test/testsuites/proxytests.go
awgreene: Sure thing!

test/e2e/main_test.go
awgreene: Good point - I had meant to remove this. I wanted to run the proxy tests part of the `OperatorGroup` tests but ran into a circular dependency issue. I instead exported a function in the `testgroups` package that lets you set a variable to run the proxy tests.

test/e2e/main_test.go
awgreene: This will still be needed, please read above.

awgreene: Perhaps naming the function `EnableProxyTests` would better portray the point of this function.

cmd/manager/main.go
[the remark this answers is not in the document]
aravindhp: I am a little puzzled by this. Moreover the logic should be that if the proxy API is not present we should not even be attempting to add it to the schem.

pkg/catalogsourceconfig/succeeded.go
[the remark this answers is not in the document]
aravindhp: This specific to `CatalogSourceConfigs` and `OperatorSources`. I would suggest adding it to `*_types.go` similar to `EnsureGVK()` etc

test/e2e/main_test.go
aravindhp: I am asking if you need the `proxyAPIAvailable` variable.

test/testsuites/proxytests.go
aravindhp: I know. You should remove the statement that you are creating the `OperatorSource` in this function.

awgreene: Ahh

cmd/manager/main.go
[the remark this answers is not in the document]
awgreene: I can move this within an if statement that checks for Proxy API.

test/e2e/main_test.go
awgreene: We do, because the `coAPIPresent` in the `e2e` package cannot be referenced from the `testgroups` package.

pkg/appregistry/appregistry.go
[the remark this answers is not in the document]
awgreene: Thanks @shawn-hurley, I tried using `ProxyFromEnvironment` but the result seems to be cached, setting the env variables did cause marketplace to use the proxy.

pkg/apis/operators/v2/catalogsourceconfig_types.go
```diff
@@ -79,6 +79,12 @@ func (csc *CatalogSourceConfig) EnsureGVK() {
 	csc.SetGroupVersionKind(gvk)
 }
 
+// ForceReconcile forces the CatalogSourceConfig object to be reconciled.
+func (csc *CatalogSourceConfig) ForceReconcile() {
```
aravindhp: Thinking a bit for about this, I suggest `ForceUpdate()`.

pkg/proxy/helpers.go
```diff
@@ -12,26 +15,50 @@ const (
 
 var (
 	// ClusterProxyKey is the key for the cluster proxy.
-	ClusterProxyKey = types.NamespacedName{Name: ClusterProxyName}
+	clusterProxyKey = types.NamespacedName{Name: ClusterProxyName}
 )
 
+// CheckProxyResource returns true if the deployment with the given namespacedname
```
aravindhp: Nit: Did you mean to input the `namespaced` name here? If you not you might want to fix your comment.

test/testgroups/opsrccsctests.go
```diff
@@ -50,12 +50,12 @@ func OpSrcCscTestGroup(t *testing.T) {
...
-	if proxyAPIAvailable {
+	if proxyTestsEnabled {
 		t.Run("proxy-test-suites", testsuites.ProxyTests)
 	}
-	t.Run("opsrc-creation-test-suite", testsuites.OpSrcCreation)
-	t.Run("csc-target-namespace-test-suite", testsuites.CscTargetNamespace)
-	t.Run("packages-test-suite", testsuites.PackageTests)
-	t.Run("csc-invalid-tests", testsuites.CscInvalid)
-	t.Run("watch-tests", testsuites.WatchTests)
+	// t.Run("opsrc-creation-test-suite", testsuites.OpSrcCreation)
```
aravindhp: I am assuming you missed reverting this before pushing :-)

pkg/proxy/envvar.go
```diff
@@ -1,88 +1,60 @@
...
 
 const (
-	// HTTP_PROXY is the URL of the proxy for HTTP requests.
-	envHTTPProxyName = "HTTP_PROXY"
+	// HTTPProxy is the name of the environment variable that sets the proxy for HTTP requests.
+	HTTPProxy = "HTTP_PROXY"
 
-	// HTTPS_PROXY is the URL of the proxy for HTTPS requests.
-	envHTTPSProxyName = "HTTPS_PROXY"
+	// HTTPSProxy is the name of the environment variable that sets the proxy for  HTTPS requests.
```
aravindhp: Nit: extra space after `for`.

pkg/proxy/envvar_test.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"os"
+	"testing"
+
+	"github.com/operator-framework/operator-marketplace/pkg/proxy"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func TestGetOperatorEnvVars(t *testing.T) {
```
aravindhp: Would this test pass even if `TestSetEnvVars()` ran first?

pkg/proxy/helpers.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	desiredEnvVars := GetOperatorEnvVars()
+
+	if len(desiredEnvVars) != len(actualEnvVars) {
+		return true
+	}
+
+	// Marketplace takes ownership of env vars, and expects them
+	// to be in the order returned by GetOperatorEnvVars.
```
aravindhp: This is risky. There is no guarantee even then if the ordering will be the same. It is safer to sort and then compare. Take a look at [cache.IsEntryStale()](https://github.com/operator-framework/operator-marketplace/blob/master/pkg/catalogsourceconfig/cache.go#L77).

pkg/registry/registry.go
```diff
@@ -137,6 +138,9 @@ func (r *registry) ensureDeployment(appRegistries []string, needServiceAccount b
 			if needServiceAccount {
 				deployment.Spec.Template.Spec.ServiceAccountName = r.key.Name
 			}
+			// Add, Update, or Remove the proxy env vars based on their values in the marketplace pod.
```
aravindhp: I think you are just updating now.

pkg/registry/registry.go
```diff
@@ -354,6 +358,10 @@ func (r *registry) newPodTemplateSpec(registryCommand []string, needServiceAccou
 	if needServiceAccount {
 		podTemplateSpec.Spec.ServiceAccountName = r.key.Name
 	}
+
+	// Add, Update, or Remove the proxy env vars based on their values in the marketplace pod.
```
aravindhp: I think you are just updating here.

pkg/watches/helpers.go
```diff
@@ -207,3 +209,54 @@ func getOpSrcOwnerKey(labels map[string]string) *client.ObjectKey {
...
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// If the cluster operator is ever changed we should kick off an event.
+			// Only check if the old name was "cluster", a kube object cannot have it's
+			// name changed in an update.
+			if e.MetaOld.GetName() == proxy.ClusterProxyName {
+				return true
+			}
+			return false
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
```
aravindhp: You have to take `e.DeleteStateUnknown` into account. Please see the watches for child resources for an example.

pkg/apis/operators/v2/catalogsourceconfig_types.go
awgreene: Sure.

test/testgroups/opsrccsctests.go
awgreene: That is correct.

pkg/proxy/envvar_test.go
awgreene: Good point. I will set the env vars above.

pkg/proxy/helpers.go
awgreene: Understood, I will make sure to sort and then compare the two lists.

pkg/proxy/envvar_test.go
```diff
@@ -0,0 +1,90 @@
...
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	NO_Proxy    = "NO_PROXY"
+	HTTP_Proxy  = "HTTP_PROXY"
+	HTTPS_Proxy = "HTTPS_PROXY"
+)
+
+func TestGetOperatorEnvVars(t *testing.T) {
```
aravindhp: Nit: I am not sure what feature this test is exercising that is not exercised in `TestSetEnvVars()`

awgreene: That's a good point, should I remove the test?

awgreene: I'll remove the test.

cmd/manager/main.go
```diff
@@ -59,6 +62,17 @@ func main() {
 		log.Fatal(err)
 	}
 
+	// Set proxy API availability
+	k8sInterface, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	err = proxy.SetProxyAvailability(k8sInterface.Discovery())
```
ecordell: are we relying on the pod getting restarted if discovery is temporarily unavailable?

awgreene:
Ready for review @ecordell

cmd/manager/main.go
awgreene: Yes - if the discover api fails marketplace would log a fatal error.

test/testgroups/opsrccsctests.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+var proxyTestsEnalbed = false
```
ecordell: super-nit: `Enalbed` -> `Enabled`

PR #222, opened 2019-07-16
Bug 1732164: - [metrics] Report metrics for default operator-sources

varshaprasad96:
Problem:
The OpenShift Marketplace does not have a method to communicate to the user,
when a default operator source fails.
Solution:
This PR adds the custom metrics collection logic for three major operator sources -
redhat, community and certified. When an operator source fails to reconcile,
the respective operator-source name, error message and count of failures
are reported by the prometheus. Alerts are created on these metrics, so as to 
integrate them with telemetry server.


Gopkg.toml
```diff
@@ -10,6 +10,10 @@ required = [
   "sigs.k8s.io/controller-tools/pkg/crd/generator",
```
awgreene: Please follow the 50/72 rule in your commit.

cmd/manager/main.go
```diff
@@ -26,10 +32,13 @@ import (
 const (
 	// TODO: resyncInterval is hardcoded to 1 hour now, it would have to be
 	// configurable on a per OperatorSource level.
-	resyncInterval = time.Duration(60) * time.Minute
-
+	resyncInterval             = time.Duration(60) * time.Minute
 	initialWait                = time.Duration(1) * time.Minute
 	updateNotificationSendWait = time.Duration(10) * time.Minute
+
+	// Endpoint and path where the metrics would be exposed.
```
awgreene: This should be `// Endpoint and path where the metrics are be exposed.`

cmd/manager/main.go
```diff
@@ -57,10 +69,15 @@ func main() {
 		log.Fatal(err)
 	}
 
-	// Create a new Cmd to provide shared dependencies and start components
-	mgr, err := manager.New(cfg, manager.Options{Namespace: namespace})
+	// Create a new Cmd to provide shared dependencies and start components at the specified
```
awgreene: `... in the specified namespace.`

cmd/manager/main.go
```diff
@@ -83,6 +100,20 @@ func main() {
...
 
+	ctx := context.TODO()
+
+	// Create Service object to expose the metrics port.
+	service, err := metrics.ExposeMetricsPort(ctx, metricsPort)
+	if err != nil {
+		log.Errorf("Unable to expose metrics port: %v", err)
+	}
+
+	// Create a serviceMonitor for the namespace, and register the service created above to it.
```
awgreene: Please remove the `,`

cmd/manager/main.go
```diff
@@ -83,6 +100,20 @@ func main() {
 		exit(err)
 	}
 
+	ctx := context.TODO()
+
+	// Create Service object to expose the metrics port.
```
awgreene: `Create a Service...`

pkg/customMetrics/customMetrics.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+// Registry-namespaces of operator sources.
+const (
```
awgreene: This will need to be updated based on our offline conversation with @aravindhp

pkg/customMetrics/customMetrics.go
```diff
@@ -0,0 +1,105 @@
...
+	REDHAT_OPERATOR_NS    = "redhat-operators"
```
awgreene: Please add comments on all these constants.

pkg/customMetrics/customMetrics.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// List of custom metrics which are to be collected.
+	metricsList = []prometheus.Collector{
+		redhatErrorCount,
+		communityErrorCount,
+		certifiedErrorCount,
+		customOperatorErrorCount,
+	}
+
+	// custom metrics port and path where the metrics would be
```
awgreene: Same as above... Why is this described in two places?

pkg/customMetrics/customMetrics.go
```diff
@@ -0,0 +1,105 @@
...
+	// exposed.
+	metricsPath = "/metrics"
+	metricsPort = ":8383"
```
awgreene: Same as above... Why is this described in two places? You could probably make the `metricsPath, metricsPort, and metricsHost` global constants within the metrics package.

pkg/customMetrics/customMetrics.go
```diff
@@ -0,0 +1,105 @@
+package customMetrics
```
awgreene: I vote to rename this package metrics.

pkg/operatorsource/handler.go
```diff
@@ -4,15 +4,23 @@ import (
...
 	"github.com/operator-framework/operator-marketplace/pkg/appregistry"
+	"github.com/operator-framework/operator-marketplace/pkg/customMetrics"
 	"github.com/operator-framework/operator-marketplace/pkg/datastore"
 	"github.com/operator-framework/operator-marketplace/pkg/phase"
 	log "github.com/sirupsen/logrus"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
+// Names of the operator-sources shipped with marketplace
```
awgreene: There is no need to redefine these constants.

pkg/operatorsource/handler.go
```diff
@@ -123,3 +137,18 @@ func (h *operatorsourcehandler) transition(ctx context.Context, logger *log.Entr
 
 	return reconciliationErr
 }
+
+// updateMetrics takes the name of the operator source which fails to reconcile, and
+// increments the error count of the respective operator metric.
+func updateMetrics(opsrc string) {
```
awgreene: I would move this to the `metrics` package.

pkg/customMetrics/customMetrics.go
varshaprasad96: The default package provided by SDK for metrics collection is also named as `metrics`.  Just for the differentiation, this is named customMetrics. If that's fine,  shall I revert back to metrics?

awgreene: That is fine with me.

cmd/manager/main.go
```diff
@@ -57,10 +69,15 @@ func main() {
...
-	mgr, err := manager.New(cfg, manager.Options{Namespace: namespace})
+	// Create a new Cmd to provide shared dependencies and start components at the specified
+	// namespace.
+	mgr, err := manager.New(cfg, manager.Options{
+		Namespace:      namespace,
+		MapperProvider: restmapper.NewDynamicRESTMapper,
+	})
 	if err != nil {
 		log.Fatal(err)
+		os.Exit(1)
```
galletti94: why are we using `os.Exit(1)` here?

pkg/operatorsource/configuring.go
```diff
@@ -5,7 +5,7 @@ import (
 	"errors"
 
 	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/shared"
-	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	v1 "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
```
galletti94: Please check that you are using `gofmt` as your  formatting tool in your editor

pkg/operatorsource/handler.go
```diff
@@ -4,15 +4,23 @@ import (
 	"context"
 
 	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/shared"
-	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	v1 "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
```
galletti94: same comment as above

pkg/operatorsource/handler.go
```diff
@@ -123,3 +137,18 @@ func (h *operatorsourcehandler) transition(ctx context.Context, logger *log.Entr
 
 	return reconciliationErr
 }
+
+// updateMetrics takes the name of the operator source which fails to reconcile, and
+// increments the error count of the respective operator metric.
+func updateMetrics(opsrc string) {
+	switch opsrcName := opsrc; opsrcName {
+	case REDHAT_OPERATOR:
+		customMetrics.CounterForRedHatOpsrcError(opsrcName).Inc()
```
galletti94: Is the implementation of `Inc()` handled by prometheus? I am just curious because usually the first example of distributed programming errors is one of incrementing counters lol

cmd/manager/main.go
varshaprasad96: Sorry, this wouldn't be needed. I will rebase and push the changes.

pkg/operatorsource/handler.go
varshaprasad96: True!. Fortunately, Prometheus is handling that here.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,123 @@
...
+import (
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+const (
+	// Registry-namespaces of operator sources.
+	REDHAT_OPERATOR_NS    = "redhat-operators"
```
awgreene: GO Variables should not be in caps. Applies everywhere.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,123 @@
...
+	COMMUNITY_OPERATOR_NS = "community-operators"
+	CERTIFIED_OPERATOR_NS = "certified-operators"
+
+	// Default values which are to be used as labels
+	// for prometheus metrics.
+	REGISTRY_NAMESPACE = "registry_ns"
```
awgreene: What is `REGISTRY_NAMESPACE` used for?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,123 @@
...
+var (
+	redhatErrorCount = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "redhat_operator_reconcile_failure_count",
+			Help: "Monotonic count of times redhat-opsrc was unable to reconcile",
+		},
+		[]string{REGISTRY_NAMESPACE, OPSRC_NAME},
+	)
+	communityErrorCount = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
```
awgreene: Are we able to expose a string as an error? Knowing how many times an operator failed is not helpful, knowing why it failed is.

cmd/manager/main.go
```diff
@@ -85,6 +98,27 @@ func main() {
 		exit(err)
 	}
 
+	ctx := context.TODO()
+
+	// convert the value of MetricsPort from string to int32.
+	port, err := strconv.ParseInt(customMetrics.MetricsPort, 10, 32)
```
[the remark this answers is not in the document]
varshaprasad96: Its because the metricsPort in `http.ListenAndServe` requires a String input. So, conversion is need at anyone of the place.

pkg/metrics/metrics.go
varshaprasad96: It is a label used, for Prometheus metrics.

varshaprasad96:
The PR has been updated to handle two cases for reporting errors of default operators:
1. Marketplace cannot reach Quay, as it is unavailable ("Status (404) error")
2. The provided host does not exists (No such host exists)
Currently, for other cases, a generic error message is returned.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,90 @@
...
+// RecordMetrics starts the server, records the custom metrics and exposes
+// the metrics at the specified endpoint.
+func RecordMetrics() {
+	// Register metrics for the operator with the prometheus.
+	registerMetrics()
+
+	// Start the server and expose the registered metrics.
+	http.Handle(METRICSPATH, promhttp.Handler())
+	port := ":" + METRICSPORT
+	go http.ListenAndServe(port, nil)
```
ecordell: This is registering and serving prom from the global muxer - I assume there's a health check also on it somewhere?

That would mean the health check is exposed on two ports, which probably won't cause an error but is kind of weird.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,90 @@
...
+	ERROR      = "opsrc_error"
+	OPSRC_NAME = "opsrc_name"
+
+	// Path, Port and Host where custom metrics would be exposed.
+	METRICSPATH = "/metrics"
+	METRICSPORT = "8383"
+	METRICSHOST = "0.0.0.0"
+
+	// Error message to be returned when quay.io is down.
+	UNREACHABLE_ERROR = "Quay.io is unavailabe"
```
ecordell: it's a lot easier to query prometheus when label values don't have spaces. What about something like `quay_unavailable`, `no_host`, etc?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,90 @@
...
+
+// counterForOpsrcError returns the number of times error was encountered while
+// contacting quay for operstor-sources.
+func counterForOpsrcError(opsrc, opsrcError string) prometheus.Counter {
+	return operatorSourceError.WithLabelValues(opsrc, opsrcError)
+}
+
+// updateMetrics takes the name of the operator source which fails to reconcile, and
+// increments the error count of the respective operator metric.
+func UpdateMetrics(opsrc, opsrcError string) {
```
ecordell: nit: to make it clear that that you should not pass any old value here in the future, it might make sense to pass `opsrcError` as an enum value instead of an arbitrary string.

pkg/metrics/metrics.go
varshaprasad96: Yes, these are the same port and path values which SDK uses to expose metrics and health check occurs here too. 
Shall I change, or is it fine to use it from a global mux?

pkg/metrics/metrics.go
varshaprasad96: Actually, I am taking `opsrcError` based on the error message we get from `ListPackages` during Configuring phase. So it can be either of the three values (no_host, quay_down, or a custom error message) as of now. Did you mean that we pass the label values as constants ?

pkg/metrics/metrics.go
ecordell: Ah - if they're both configured on the same port that's fine. I should've checked 😄 I assumed they were configured on different ports.

pkg/metrics/metrics.go
ecordell: I meant something along the lines of 

```go
type OpSrcError string

const (
        NO_HOST_ERROR OpSrcError = "no_host"
        UNREACHABLE_ERROR OpSrcError = "quay_unavailable"
)

func UpdateMetrics(opsrc, opSrcErr OpSrcError) {
  // ...
}
```

That way it's super clear to anyone calling `UpdateMetrics` that you shouldn't make up your own string and pass it in.

varshaprasad96: Thanks @ecordell. Got it..

manifests/07_operator.yaml
```diff
@@ -41,7 +41,6 @@ spec:
           - marketplace-operator
           args:
           - -registryServerImage=quay.io/openshift/origin-operator-registry
-          - -defaultsDir=/defaults
```
awgreene: Did you mean to remove this line?

manifests/07_operator.yaml
```diff
@@ -56,12 +55,12 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            - name: OPERATOR_NAME
```
awgreene: Why are you swapping the location of this env var?

cmd/manager/main.go
```diff
@@ -118,6 +129,25 @@ func main() {
 		exit(err)
 	}
 
+	// convert the value of MetricsPort from string to int32.
+	port, err := strconv.ParseInt(customMetrics.METRICSPORT, 10, 32)
```
awgreene: I don't believe that this will compile.

cmd/manager/main.go
```diff
@@ -51,7 +58,7 @@ func printVersion() {
 func main() {
 	printVersion()
 
-	// Parse the command line arguments
+	// Parse the command line arguments for the registry server image
```
awgreene: Please undo this change.

cmd/manager/main.go
```diff
@@ -118,6 +129,9 @@ func main() {
 		exit(err)
 	}
 
+	// generate metrics and publish it at an endpoint.
```
awgreene: Please capitalize the first letter in you comments, comments are sentences.

cmd/manager/main.go
```diff
@@ -210,3 +224,28 @@ func ensureDefaults(cfg *rest.Config, scheme *kruntime.Scheme) error {
 
 	return err
 }
+
+// GenerateMetrics creates a service, service monitor, starts a metrics server
+// and records the metrics to expose at the specified endpoint.
+
```
awgreene: Please remove this newline.

cmd/manager/main.go
```diff
@@ -210,3 +224,28 @@ func ensureDefaults(cfg *rest.Config, scheme *kruntime.Scheme) error {
...
+func GenerateMetrics(cfg *rest.Config, namespace string) {
+
+	// Create a Service object to expose the metrics port.
+	service, err := metrics.ExposeMetricsPort(context.TODO(), metricsPort)
+	if err != nil {
+		log.Errorf("Unable to expose metrics port: %v", err)
+	}
+
+	// Ensure that an empty service object is not passed to create a service monitor.
```
awgreene: `// Ensure that a nil pointer is not added to the slice`

cmd/manager/main.go
```diff
@@ -3,10 +3,10 @@ package main
 import (
```
awgreene: Given that this work is now tracked via a bug, we will need to rewrite the commit message and change the PR header.

cmd/manager/main.go
```diff
@@ -93,9 +90,8 @@ func main() {
 	// watch for CatalogSources in targetNamespaces being deleted and recreate
 	// them.
 	mgr, err := manager.New(cfg, manager.Options{
-		Namespace:          namespace,
-		MapperProvider:     restmapper.NewDynamicRESTMapper,
-		MetricsBindAddress: fmt.Sprintf("%s:%d", metricsHost, metricsPort),
```
awgreene: Can you explain why this is removed? I assume it is to prevent us from reporting sdk metrics, but I want to confirm.

cmd/manager/main.go
```diff
@@ -230,6 +226,13 @@ func ensureDefaults(cfg *rest.Config, scheme *kruntime.Scheme) error {
 
 func GenerateMetrics(cfg *rest.Config, namespace string) {
 
+	// convert the value of MetricsPort from string to int32.
```
awgreene: Please capitalize `// Convert`.

cmd/manager/main.go
```diff
@@ -230,6 +226,13 @@ func ensureDefaults(cfg *rest.Config, scheme *kruntime.Scheme) error {
...
+	port, err := strconv.ParseInt(customMetrics.MetricsPort, 10, 32)
+	if err != nil {
+		log.Errorf("Unable to parse metrics port: %v", err)
```
awgreene: If this fails, won't there be an issue when you call `metricsPort := int32(port)`?

cmd/manager/main.go
```diff
@@ -230,6 +226,13 @@ func ensureDefaults(cfg *rest.Config, scheme *kruntime.Scheme) error {
 
 func GenerateMetrics(cfg *rest.Config, namespace string) {
 
+	// convert the value of MetricsPort from string to int32.
+	port, err := strconv.ParseInt(customMetrics.MetricsPort, 10, 32)
```
awgreene: Why not just defined `customMetrics.MetricsPort` as an `int32`?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+
+// RecordMetrics starts the server, records the custom metrics and exposes
+// the metrics at the specified endpoint.
+func RecordMetrics() {
+	// Register metrics for the operator with the prometheus.
+	registerMetrics()
+
+	// Start the server and expose the registered metrics.
+	http.Handle(MetricsPath, promhttp.Handler())
+	port := ":" + MetricsPort
```
awgreene: If MetricsPort was a string, you could set port to `fmt.Sprintf(":%d", MetricsPort)`.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+
+const (
+	// Default values which are to be used as labels
+	// for prometheus metrics.
+	Error     = "opsrc_error"
+	OpsrcName = "opsrc_name"
+
+	// Path, Port and Host where custom metrics would be exposed.
+	MetricsPath = "/metrics"
+	MetricsPort = "8383"
```
awgreene: This should just be an `integer`

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+	MetricsHost = "0.0.0.0"
+
+	// Error message to be returned when quay.io is down.
+	UnreachableError OpsrcError = "quay_unavailabe"
```
awgreene: Documentation for Global variables should start with the name of the global variable. This applies in all places. Please see marketplace code for examples.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+	// Error message to be returned when host address is wrong.
+	NoHostExsistsError OpsrcError = "no_host_exsists"
+)
+
+// Metric variables that report error based on the operator-source
+// which fails to reconcile.
+var (
+	operatorSourceError = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "operator_source_failure_count",
```
awgreene: This could be `opsrc_failures`, unless promethes has some convention I am unaware of. Thoughts @kevinrizza ?

awgreene: `opsrc_errors` works here as well.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+
+// updateMetrics takes the name of the operator source which fails to reconcile, and
+// increments the error count of the respective operator metric.
+func UpdateMetrics(opsrc, opsrcError string) {
+	counterForOpsrcError(opsrc, opsrcError).Inc()
+}
+
+// ErrorMessage takes the error which appears while Listing the packages,
+// parses it, and returns the error message which is to be published.
+func ErrorMessage(err string) OpsrcError {
```
awgreene: This is a good start, we will need to review how we wish to ID quay errors.

pkg/operatorsource/handler.go
awgreene: This still needs to be addressed in all places.

cmd/manager/main.go
varshaprasad96: Yes, `MetricsBindAddress` would run the command to register default metrics with at the port/path provided by SDK. Since we are adding our own custom metrics to it and registering them, this needs to be removed. Or else, we would be registering metrics to the same port twice, which will result in an error.

cmd/manager/main.go
varshaprasad96: Actually no. It would return a zero when anything other than a valid integer is passed (like a string).

pkg/metrics/metrics.go
varshaprasad96: MetricsPort as an integer cannot to passed to `http.ListenAndServe`. Casting is required in either one of the place (before starting the server, or before passing it to `ExposeMetrics` function)

cmd/manager/main.go
varshaprasad96: Because `http.ListenAndServe` requires a string.

pkg/metrics/metrics.go
varshaprasad96: Prometheus requirements are just to have "_" between words. Since this is of the type `counter` in Prometheus metrics, I named it as `operator_source_failure_count`. This could be anything though..

awgreene: That's good to know, let's see what name @kevinrizza thinks is good, perhaps `opsrc_failure_count`?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+
+// counterForOpsrcError returns the number of times error was encountered while
+// contacting quay for operstor-sources.
+func counterForOpsrcError(opsrc, opsrcError string) prometheus.Counter {
+	return operatorSourceError.WithLabelValues(opsrc, opsrcError)
+}
+
+// updateMetrics takes the name of the operator source which fails to reconcile, and
+// increments the error count of the respective operator metric.
+func UpdateMetrics(opsrc, opsrcError string) {
```
ecordell: I think this should be `func UpdateMetrics(opsrc, opsrcError OpsrcError)`?

varshaprasad96: The labels of the prometheus metrics only take in String type. So I created a function `OpsrcErrorToString`, which would just cast it into a string only while updating the labels.

varshaprasad96: I had used that in `configuring.go`, while incrementing metrics. But that doesn't serve the purpose. So I have updated it here! :+1:

cmd/manager/main.go
```diff
@@ -210,3 +219,30 @@ func ensureDefaults(cfg *rest.Config, scheme *kruntime.Scheme) error {
...
+
+func generateMetrics(cfg *rest.Config, namespace string) {
+
+	// Create a Service object to expose the metrics port.
+	service, err := metrics.ExposeMetricsPort(context.TODO(), customMetrics.MetricsPort)
+	if err != nil {
+		log.Errorf("Unable to expose metrics port: %v", err)
+	}
+
+	// Ensure that a nil pointer is not added to the slice.
```
kevinrizza: Generally I think adding a comment about the mechanics of the code is bad practice unless it is particularly unorthodox. I think in this case the purpose of checking for nil before generating the slice makes sense. If this is generally the only purpose, I think removing this comment makes sense here.


That being said if `service == nil` isn't it possible that we want to do something? Return an error perhaps? What state are we in if that happens? I think the same comment could be made just further above if `metrics.ExposeMetricsPort()` returns an error.

cmd/manager/main.go
```diff
@@ -210,3 +219,30 @@ func ensureDefaults(cfg *rest.Config, scheme *kruntime.Scheme) error {
...
+	services := []*v1.Service{}
+	if service != nil {
+		services = append(services, service)
+	}
+
+	// Create a serviceMonitor for the namespace and register the service created above to it.
+	_, err = metrics.CreateServiceMonitors(cfg, namespace, services)
+	if err != nil {
+		log.Errorf("Error Creating Service Monitor: %v", err)
```
kevinrizza: Same comment as above with respect to error handling. Is just logging appropriate?

deploy/role.yaml
```diff
@@ -120,3 +121,12 @@ rules:
   - create
   - delete
   - update
+  - list
+  - watch
```
kevinrizza: This seems like the wrong thing to do. Why do we need to list and watch secrets for this? Shouldn't we just add another rbac grouping for pods if that is what's needed?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+import (
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+// Error to be  returned when marketplace is unable to reach quay.io.
```
kevinrizza: We should avoid specifically referencing quay.io whenever possible. Keep in mind that the fact that quay hosts app registry doesn't mean that it's the only way to get an app registry instance.

pkg/metrics/metrics.go
kevinrizza: I think keeping our variable names short, especially when they are going to be displayed in a UI is good. `opsrc_failure_count` seems fine.

cmd/manager/main.go
```diff
@@ -118,6 +124,9 @@ func main() {
 		exit(err)
 	}
 
+	// Generate metrics and publish it at an endpoint.
```
kevinrizza: Should this be async? Though I suppose we care that we start recording before we start handling events so we don't miss any? Have we considered what all this initialization will do (in terms of time cost) if we don't at least serve the health check first?

pkg/metrics/metrics.go
kevinrizza: Agreed. It seems like string comparison will be brittle (if these error messages change, this code will silently break).

Perhaps we can derive meaning from control flow instead? Or at least http error codes if we can get them from the client?

varshaprasad96: Yeah, true. We do get a http 404 error code when quay is not available. But when the endpoint is wrong, it remains in `configuring` phase, trying to connect. There is no specific error code returned in the error message. I am not sure on how to proceed with that case though..

cmd/manager/main.go
```diff
@@ -36,8 +40,7 @@ import (
 const (
 	// TODO: resyncInterval is hardcoded to 1 hour now, it would have to be
 	// configurable on a per OperatorSource level.
-	resyncInterval = time.Duration(60) * time.Minute
-
+	resyncInterval             = time.Duration(60) * time.Minute
```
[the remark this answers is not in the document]
varshaprasad96: Its due to formatting issues. Ill remove the spaces.

pkg/metrics/metrics.go
kevinrizza: We put ourselves back in the configuring phase because we know there was some error and we want to retry. We should be able to dig into where we decide to put ourselves back into that phase and probably learn more. And, if there isn't something more specific being returned, is it possible for us to adjust how the appr client is giving a result? Are we really restricted to just strings? Could we update the error that is returned to have an `ErrorCode` field or something similar?

cmd/manager/main.go
varshaprasad96: The `RecordMetrics()` function uses the endpoint defined by SDK. Hence the health check does happen here before the metrics are served. The scrapping of data by Prometheus is synchronous. But yes, as you said we can decouple recording metrics logic from this function and start collecting metrics initially. Creation of service and service monitor could happen later.

cmd/manager/main.go
varshaprasad96: If the service is nil, then we return an error stating `Unable to expose metrics port`. That part  is just to make sure that we do not pass a nil to the next function. Because if `nil` value is passed to `CreateServiceMonitor`, then a segfault occurs. So, when a service is not created, then servicemonitor is not created either and we do not expose metrics.

awgreene: I believe @kevinrizza is trying to point out that we may want to fail the operator rather than just logging the error and carrying on.

deploy/role.yaml
varshaprasad96: `watch` is not needed. I just reconfirmed with the SDK team too. `list` would be required and is sufficient!

varshaprasad96:
@awgreene  @ecordell - I have updated the code based on the suggestions provided. Please have a look at it.

awgreene:
Closed in favor of #246

PR #223, opened 2019-07-16
[migration] Update existing Subscriptions and remove stale resources

galletti94:
With #198 introducing a change in a way that a datastore CatalogSource is directly created by an OperatorSource instead of a CatalogSourceConfig, this PR performs some upgrade logic to update/cleanup existing/stale
resources in existing clusters. Specifically:
* Subscriptions created by the UI with labels "csc-owner-name" and "csc-owner-namespace" are updated to reference datastore CatalogSources, instead of installed CatalogSoures.
* ClusterRole was added to list Subscriptions in all namespaces.
* Stale datastore CatalogSourceConfigs created before #198 are deleted.
* Stale installed CatalogSourceConfigs created during operator installation in old clusters are deleted.


pkg/status/status.go
```diff
@@ -408,3 +408,15 @@ func (s *status) monitorClusterStatus() {
 		}
 	}
 }
+
+// ReportMigration sets the clusterOperator status to signal that migration logic is in progress,
+// while upgrading from openshift 4.1.x to openshift 4.2.0
+func ReportMigration(cfg *rest.Config, mgr manager.Manager, namespace string, version string, stopCh <-chan struct{}) error {
+	instance = new(cfg, mgr, namespace, version, stopCh)
+	conditionListBuilder := clusterStatusListBuilder()
+	conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionTrue, fmt.Sprintf("Performing migration logic to progress towards release version: %s", instance.version), "Upgrading from 4.1 to 4.2")
```
aravindhp: You cannot hard code the version in the reason. Moreover the reason has to be one CamelCase word, so you could just go with `Upgrading`

pkg/status/status.go
```diff
@@ -408,3 +408,15 @@ func (s *status) monitorClusterStatus() {
...
+	msg := fmt.Sprintf("Determining status")
+	conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionFalse, msg, "Upgrading from 4.1 to 4.2")
+	statusConditions := conditionListBuilder(configv1.OperatorDegraded, configv1.ConditionFalse, msg, "Upgrading from 4.1 to 4.2")
```
aravindhp: You also need to set the `Upgradable` condition to false here.

test/helpers/helpers.go
```diff
@@ -334,3 +348,104 @@ func CheckCscSuccessfulDeletion(client test.FrameworkClient, cscName, namespace,
...
+	}
+	err = WaitForSuccessfulDeployment(client, *marketplace)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// CreateDatastoreCsc returns a newly built CatalogSourceConfig
+func CreateDatastoreCsc(name, namespace string) *v2.CatalogSourceConfig {
```
aravindhp: Please be consistent and use `CreateDatastoreCscDefinition` like we are doing with `CreateOperatorSourceDefinition`. Applies to all places.

test/helpers/helpers.go
```diff
@@ -40,6 +41,19 @@ const (
 	// TestOperatorSourceLabelValue is a label value added to the opeator source returned
 	// by the CreateOperatorSource function.
 	TestOperatorSourceLabelValue string = "Community"
+
+	// TestDatastoreCscName is the name of a CatalogSourceConfig that is returned by
+	// the CreateDatastoreCsc function.
+	TestDatastoreCscName string = "test-operators"
```
aravindhp: Won't there be a conflict with the child resources of the test `OperatorSource` given they both have the same name?

test/helpers/helpers.go
```diff
@@ -334,3 +348,104 @@ func CheckCscSuccessfulDeletion(client test.FrameworkClient, cscName, namespace,
...
+			Namespace: namespace,
+		},
+		Spec: v2.CatalogSourceConfigSpec{
+			TargetNamespace: namespace,
+			Packages:        "",
+		},
+	}
+}
+
+// CreateSubscription returns a newly built Subscription with the lables
```
aravindhp: `labels`

test/helpers/helpers.go
```diff
@@ -334,3 +348,104 @@ func CheckCscSuccessfulDeletion(client test.FrameworkClient, cscName, namespace,
...
+			TargetNamespace: namespace,
+			Packages:        "",
+		},
+	}
+}
+
+// CreateInstalledCsc returns a newly built CatalogSourceConfig
+func CreateInstalledCsc(namespace string) *v2.CatalogSourceConfig {
+	name := TestInstalledCscPublisherName + namespace
+	return &v2.CatalogSourceConfig{
```
aravindhp: Doesn't this require some labels too?

test/helpers/helpers.go
```diff
@@ -334,3 +348,104 @@ func CheckCscSuccessfulDeletion(client test.FrameworkClient, cscName, namespace,
...
+	err = WaitForSuccessfulDeployment(client, *marketplace)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// CreateDatastoreCsc returns a newly built CatalogSourceConfig
+func CreateDatastoreCsc(name, namespace string) *v2.CatalogSourceConfig {
+	lables := make(map[string]string)
```
aravindhp: `labels`

test/testsuites/migrationtests.go
```diff
@@ -0,0 +1,60 @@
...
+	t.Run("subscriptions-are-updated", testSubscriptionsUpdated)
+}
+
+// testCatalogSourceConfigsCleanedUp ensures that after a cluster is migrated
+// from openshift 4.1.x to openshift 4.2.0, the following stale objects are cleaned up:
+// Stale CatalogSourceConfigs created during operator installation in a openshift 4.1.x cluster.
+// Datastore CatalogSourceConfigs created by OperatorSources in a openshift 4.1.x cluster.
+func testCatalogSourceConfigsCleanedUp(t *testing.T) {
+	// Get test namespace
+	namespace, err := test.NewTestCtx(t).GetNamespace()
```
aravindhp: Don't you need to `defer ctx.Cleanup()` to be on the safe side?

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,224 @@
...
+	return installedCscs
+}
+
+// deleteInstalledCscs deletes the CSCs installed during operator installation.
+// The child resources of the CSCs are delete by the finalizer.
+func (m *migrator) deleteInstalledCscs(cscs []types.NamespacedName, operatorNamespace string) {
+	for _, cscInfo := range cscs {
+		// If the CatalogSourceConfig namespace information is missing, try and find the
+		// CatalogSourceConfig in the marketplace-operator's namespace
+		if cscInfo.Namespace == "" {
```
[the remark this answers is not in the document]
galletti94: Good point, these csc's should always be in the `operatorNamespace`.

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,224 @@
...
+// For example, for a CatalogSourceConfig named
+// `installed-community-openshift-operators`, extractCsName
+// extracts `community` off the name, appends `operators`
+// to it, and returns `community-operators` as output
+func ExtractCsName(cscName string) string {
+	possibleCsName := strings.Split(cscName, "-")[1]
+	return fmt.Sprintf("%s-%s", possibleCsName, "operators")
+}
+
+// newCatalogSourceConfig returns a newly built CatalogSourceConfig
```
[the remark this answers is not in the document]
galletti94: There is no builder for `CatalogSourceConfig` anymore

pkg/status/status.go
galletti94: I wasn't aware of that thanks for pointing this out!

test/helpers/helpers.go
galletti94: What labels were you thinking of? I don't think we need the owner labels here because they are added to the child resources by the operator.

test/helpers/helpers.go
galletti94: I think you are correct - good catch!

test/testsuites/migrationtests.go
galletti94: Isn't it done from the parent testgroup here https://github.com/operator-framework/operator-marketplace/pull/223/files/5c068a0e5aef87360da32e2b7e7288ac797109ab#diff-33a154e1b96c5e7d9d61ed336d00d7efR20? Maybe I misunderstood

pkg/migrator/migrator.go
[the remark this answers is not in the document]
aravindhp: Ah you are right. Sorry for the noise.

test/helpers/helpers.go
aravindhp: I was thinking of any labels that the UI would add and that you are looking for in the upgrade code.

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,224 @@
...
+
+	subscriptions := &olm.SubscriptionList{}
+	// Get the list of existing Subscriptions that have the label "csc-owner-name"
+	err := m.client.List(context.TODO(), options, subscriptions)
+	if err != nil {
+		m.logger.Errorf(fmt.Sprintf("Client error: %s", err.Error()))
+		return []types.NamespacedName{}
+	}
+	for _, instance := range subscriptions.Items {
+		installedCscs = append(installedCscs, types.NamespacedName{Name: instance.GetLabels()[builders.CscOwnerNameLabel], Namespace: instance.GetLabels()[builders.CscOwnerNamespaceLabel]})
```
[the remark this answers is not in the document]
aravindhp: This seems to imply that you are identifying installed `CatalogSourceConfigs` using the owner labels. Is that correct? Won't user created ones also be picked up here? Or do we want to fix up user created ones too?

test/testsuites/migrationtests.go
aravindhp: You are creating a new context here.

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,224 @@
...
+		if err != nil {
+			m.logger.Errorf("[migration] Failed to delete CatalogSourceConfig %s. Error: %s", csc.GetName(), err.Error())
+		} else {
+			m.logger.Infof("[migration] Datastore CSC %s scheduled for deletion.", csc.GetName())
+		}
+	}
+}
+
+// findCatalogSource infers the datastore CatalogSource created by the OperatorSource
+// in an Openshift 4.2.0 cluster. The inferred datastore CatalogSource will then be
```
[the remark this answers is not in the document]
galletti94: How about `4.1.z` as following the convention `x.y.z`?

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,224 @@
...
+			// infer the CatalogSource from the OperatorSource that has the package
+			datastoreCs, err = findCsFromOpsrc(&instance, m.client)
+			if err != nil {
+				m.logger.Errorf("[migration] Could not infer datastore CatalogSource for Subscription %s.", instance.GetName())
+				continue
+			}
+		}
+		// update the Subscription to reference the datastore CatalogSource
+		instance.Spec.CatalogSource = datastoreCs.GetName()
+		instance.Spec.CatalogSourceNamespace = datastoreCs.GetNamespace()
```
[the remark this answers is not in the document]
aravindhp: If it is the owner labels that are being used, then they should not be removed here.

test/helpers/helpers.go
galletti94: I don't see any use of labels in the UI code. We just use the naming convention `installed-...`

pkg/migrator/migrator.go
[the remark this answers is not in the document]
aravindhp: Re-reading this again, do we even need to mention the version? If there needs to be mention of a version, it should be about the `Subscriptions` created in a `4.1.z` cluster that references an installed `CatalogSourceConfig`.

test/testsuites/migrationtests.go
galletti94: I see - thanks!

pkg/migrator/migrator.go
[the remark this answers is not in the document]
galletti94: https://github.com/openshift/console/blob/release-4.1/frontend/public/components/operator-hub/operator-hub-subscribe.tsx#L113

test/helpers/helpers.go
galletti94: We only need these labels set on the subscription object so that the csc can be found. This is being done correctly below in the `CreateSubscription` function

test/helpers/helpers.go
galletti94: I don't think this is an issue though since the goal of this csc is to be correctly referenced by the subscription after upgrade. It isn't deleted or updated or anything that would cause conflicts with the `operatorsource`. Also this is probably intended to mimic the 4.1 scene of having csc's and opsrc's with the same name before upgrade.

pkg/migrator/migrator.go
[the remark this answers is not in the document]
galletti94: We are indeed setting these owner labels in the 4.1 UI ^

pkg/migrator/migrator.go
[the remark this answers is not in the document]
galletti94: Good point

pkg/migrator/migrator.go
```diff
@@ -39,115 +39,141 @@ func NewMigrator(client client.Client) migrator {
...
+// Migrate is a wrapper for doMigrate so that we can retry the migration in case
+// of failures.
 func (m *migrator) Migrate(operatorNamespace string) {
-	installedCscs := m.updateSubscriptions()
-	m.deleteInstalledCscs(installedCscs, operatorNamespace)
-	m.deleteDatastoreCscs(operatorNamespace)
+	errorList := m.doMigrate(operatorNamespace)
+	if errorList != nil {
+		for _, err := range errorList {
+			m.logger.Errorf(err)
```
aravindhp: Rather than logging the errors here we should log them in place.

pkg/migrator/migrator.go
```diff
@@ -39,115 +39,141 @@ func NewMigrator(client client.Client) migrator {
...
+		}
+		m.logger.Infof("[migration] Failed to migrate")
+		m.retry = m.retry - 1
+		if m.retry > 0 {
```
aravindhp: Won't a `for` loop be more cleaner? I don't see a reason to do a recursion.

cmd/manager/main.go
```diff
@@ -103,6 +105,18 @@ func main() {
 	log.Print("Starting the Cmd.")
 	stopCh := signals.SetupSignalHandler()
 
+	// set ClusterOperator status to report Migration
+	status.ReportMigration(cfg, mgr, namespace, os.Getenv("RELEASE_VERSION"), stopCh)
+
+	client, err := client.New(cfg, client.Options{})
+	if err != nil {
+		exit(err)
```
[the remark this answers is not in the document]
aravindhp: What are you going to do in this scenario? I am not sure if we should be exiting the operator.

pkg/migrator/migrator.go
```diff
@@ -39,115 +39,141 @@ func NewMigrator(client client.Client) migrator {
...
 
-// Migrate updates existing Subscriptions, deletes the CSCs installed during
+// doMigrate updates existing Subscriptions, deletes the CSCs installed during
 // operator installation as well the stale datastore CSCs when the cluster is
-// migrating from Openshift 4.1.x to Openshift 4.2.x
+// migrating from Openshift 4.1.z to Openshift 4.2.z
+func (m *migrator) doMigrate(operatorNamespace string) []string {
+	installedCscs, errorList := m.updateSubscriptions()
+	if errorList != nil {
+		return errorList
```
aravindhp: I am wondering if we should collect the list of errors here too and return it at the end. For that we would need to ensure that the errors in one step will not affect the next.

pkg/migrator/migrator.go
```diff
@@ -39,115 +39,141 @@ func NewMigrator(client client.Client) migrator {
...
+	retry  int
 }
 
-// Migrate updates existing Subscriptions, deletes the CSCs installed during
+// doMigrate updates existing Subscriptions, deletes the CSCs installed during
 // operator installation as well the stale datastore CSCs when the cluster is
-// migrating from Openshift 4.1.x to Openshift 4.2.x
+// migrating from Openshift 4.1.z to Openshift 4.2.z
+func (m *migrator) doMigrate(operatorNamespace string) []string {
+	installedCscs, errorList := m.updateSubscriptions()
```
aravindhp: Please use `errors`

pkg/migrator/migrator.go
```diff
@@ -39,115 +39,141 @@ func NewMigrator(client client.Client) migrator {
...
+	}
 }
 
 // updateSubscriptions updates the existing Subscriptions'
 // spec.source and spec.sourcenamespace fields. Existing
 // Subscriptions referenced installed CatalogSources, which
 // are updated to reference datastore CatalogSources
 // instead.
-func (m *migrator) updateSubscriptions() []types.NamespacedName {
+func (m *migrator) updateSubscriptions() ([]types.NamespacedName, []string) {
```
aravindhp: This should return error aggregate. Please see https://github.com/operator-framework/operator-marketplace/blob/master/pkg/catalogsourceconfig/triggerer.go#L62

Comment applies to all functions.

pkg/migrator/migrator.go
```diff
@@ -39,115 +39,141 @@ func NewMigrator(client client.Client) migrator {
...
-		return []types.NamespacedName{}
+		errorList = append(errorList, fmt.Sprintf("[migration] Client error: %s", err.Error()))
+		return []types.NamespacedName{}, errorList
 	}
-	for _, instance := range subscriptions.Items {
-		installedCscs = append(installedCscs, types.NamespacedName{Name: instance.GetLabels()[builders.CscOwnerNameLabel], Namespace: instance.GetLabels()[builders.CscOwnerNamespaceLabel]})
-		// try to infer the datastore CatalogSource from the Subscription
-		datastoreCs, err := findCatalogSource(&instance, m.client)
+	for _, subscription := range subscriptions.Items {
+		installedCscs = append(installedCscs, types.NamespacedName{Name: subscription.GetLabels()[builders.CscOwnerNameLabel], Namespace: subscription.GetLabels()[builders.CscOwnerNamespaceLabel]})
```
aravindhp: - I commented about this earlier too. You cannot assume that these labels exist
- Please make this multi-line

pkg/migrator/migrator.go
```diff
@@ -39,115 +39,141 @@ func NewMigrator(client client.Client) migrator {
...
 				continue
 			}
 		}
-		// update the Subscription to reference the datastore CatalogSource
-		instance.Spec.CatalogSource = datastoreCs.GetName()
-		instance.Spec.CatalogSourceNamespace = datastoreCs.GetNamespace()
-		err = m.client.Update(context.TODO(), &instance)
+		// Update the Subscription to reference the datastore CatalogSource
+		subscription.Spec.CatalogSource = datastoreCs.GetName()
+		subscription.Spec.CatalogSourceNamespace = datastoreCs.GetNamespace()
```
aravindhp: This is where you would the remove the labels

pkg/operatorsource/configuring_test.go
```diff
@@ -231,7 +231,7 @@ func TestReconcile_CatalogSourceConfigAlreadyExists_Updated(t *testing.T) {
 	}
 	opsrcIn.SetLabels(labelsNeed)
 
-	// We expect the remote registry to return a non-empty list of manifest(s).
+g	// We expect the remote registry to return a non-empty list of manifest(s).
```
aravindhp: How is this code compiling?

pkg/migrator/migrator.go
galletti94: I see and just report a generic error to be passed to the parent so that it knows whether to retry?

galletti94: or do we report the errors in place and also collect them so that they can be passed to the parent function?

aravindhp: Please see https://github.com/operator-framework/operator-marketplace/pull/223#discussion_r304573404

pkg/migrator/migrator.go
galletti94: Yeah there was no specific reason for me to do this recursively - just habit

pkg/migrator/migrator.go
galletti94: Ah must have slipped through the cracks thanks for catching that

galletti94:
@aravindhp @awgreene I addressed the comments - PTAL

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,251 @@
...
+				types.NamespacedName{
+					Name:      subscription.GetLabels()[builders.CscOwnerNameLabel],
+					Namespace: subscription.GetLabels()[builders.CscOwnerNamespaceLabel]})
+			// Try to infer the datastore CatalogSource from the Subscription
+			datastoreCs, err := findCatalogSource(&subscription, m.client)
+			if k8s_errors.IsNotFound(err) {
+				// Infer the CatalogSource from the OperatorSource that has the package
+				datastoreCs, err = findCsFromOpsrc(&subscription, m.client)
+				if err != nil {
+					msg := fmt.Sprintf("[migration] Could not infer datastore CatalogSource for Subscription %s", subscription.GetName())
```
aravindhp: Why are you throwing away the error that you got and replacing it with a custom error? It is better to propagate that error up instead of replacing it with your own error. Adding a message to the existing error is fine like you are doing other places. This comment applies to all places that you are doing this. This applies to all places that you are doing this.

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,251 @@
...
+// are updated to reference datastore CatalogSources
+// instead.
+func (m *migrator) updateSubscriptions() ([]types.NamespacedName, []error) {
+	var installedCscs []types.NamespacedName
+	var errors []error
+	options := &client.ListOptions{}
+	options.SetLabelSelector(fmt.Sprintf(builders.CscOwnerNameLabel))
+
+	subscriptions := &olm.SubscriptionList{}
+	// Get the list of existing Subscriptions that have the label "csc-owner-name"
```
aravindhp: Please add a comment that the 4.1 UI added these labels.

test/helpers/helpers.go
```diff
@@ -373,3 +387,104 @@ func CheckCscSuccessfulDeletion(client test.FrameworkClient, cscName, namespace,
...
+		return err
+	}
+	replicas := marketplace.Spec.Replicas
+	zero := int32(0)
+	marketplace.Spec.Replicas = &zero
+	err = client.Update(context.TODO(), marketplace)
+	if err != nil {
+		return err
+	}
+	err = client.Get(context.TODO(), types.NamespacedName{Name: "marketplace-operator", Namespace: namespace}, marketplace)
```
aravindhp: I remember running into a problem where back to back updates to a deployment doesn't get picked up if the second update is just reverting to the original state. So you could cause a test flake here.

test/helpers/helpers.go
```diff
@@ -373,3 +387,104 @@ func CheckCscSuccessfulDeletion(client test.FrameworkClient, cscName, namespace,
...
+		},
+		Spec: v2.CatalogSourceConfigSpec{
+			TargetNamespace: namespace,
+			Packages:        "",
+		},
+	}
+}
+
+// CreateInstalledCsc returns a newly built CatalogSourceConfig
+func CreateInstalledCsc(namespace string) *v2.CatalogSourceConfig {
```
aravindhp: Nit: missing `Definition` prefix. Please be consistent.

test/helpers/helpers.go
```diff
@@ -373,3 +387,104 @@ func CheckCscSuccessfulDeletion(client test.FrameworkClient, cscName, namespace,
...
+		Spec: v2.CatalogSourceConfigSpec{
+			TargetNamespace: namespace,
+			Packages:        "",
+		},
+	}
+}
+
+// CreateSubscription returns a newly built Subscription with the labels
+// `csc-owner-name` and `csc-owner-namespace`
+func CreateSubscription(name, namespace string) *olm.Subscription {
```
aravindhp: Nit: missing `Definition` prefix. Please be consistent.

deploy/role.yaml
```diff
@@ -7,15 +7,12 @@ rules:
   - operators.coreos.com
   resources:
   - catalogsources
-  - catalogsourceconfigs
```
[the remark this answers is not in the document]
galletti94: I must have screwed something up during rebase - thanks for catching that!

pkg/migrator/migrator.go
galletti94: I am not sure I understand. We are raising these errors in the log as they appear. I am only surfacing these errors so that I know whether the migration failed or not and displaying that it failed. I get that we are not using these surfaced errors for anything right now and so collecting them might be overkill for the current usage. Is there a particular way you imagined these errors being used?

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,251 @@
...
+	err := m.client.List(context.TODO(), options, subscriptions)
+	if err != nil {
+		msg := fmt.Sprintf("[migration] Client error: %s", err.Error())
+		errors = append(errors, fmt.Errorf(msg))
+		m.logger.Errorf(msg)
+		return []types.NamespacedName{}, errors
+	}
+	for _, subscription := range subscriptions.Items {
+		if builders.HasOwnerLabels(subscription.GetLabels(), v2.CatalogSourceConfigKind) {
+			installedCscs = append(
```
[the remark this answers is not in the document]
galletti94: Good catch!

pkg/migrator/migrator.go
aravindhp: Where is `err` not being reflected in the message you are constructing? What is the point in getting a log from a customer with the error message that you have shown? You are anyhow collecting the error messages. So why throw away the real error when you are collecting them?

galletti94: Oh ok I didn't notice `err` wasn't being reported back in that error! Yes I totally agree - that's my bad. Sorry for the misunderstanding

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,251 @@
...
+
+// Migrate updates existing Subscriptions, deletes the CSCs installed during
+// operator installation as well the stale datastore CSCs when the cluster is
+// migrating from Openshift 4.1.z to Openshift 4.2.z
+func (m *migrator) Migrate(operatorNamespace string) error {
+	installedCscs, errors := m.updateSubscriptions()
+	errors = append(errors, m.deleteInstalledCscs(installedCscs, operatorNamespace)...)
+	errors = append(errors, m.deleteDatastoreCscs(operatorNamespace)...)
+	if errors != nil {
+		return fmt.Errorf("[migration] Failed to complete migration")
```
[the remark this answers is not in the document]
galletti94: Wasn't sure what to use it for since the errors are being logged in place anyway. Do you have a use case in mind?

pkg/builders/builders.go
```diff
@@ -43,3 +43,39 @@ func GetOwnerLabel(name, namespace, owner string) map[string]string {
 		return map[string]string{}
 	}
 }
+
+// RemoveOwnerLabels removes the owner labels from a set of labels
+func RemoveOwnerLabels(labels map[string]string, owner string) map[string]string {
```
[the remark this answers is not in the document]
galletti94: Sounds good to me

pkg/migrator/migrator.go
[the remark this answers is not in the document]
aravindhp: Better to return the errors. The call site is place to make the decision as to what to do with it.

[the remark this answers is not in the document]
galletti94: makes sense

test/helpers/helpers.go
```diff
@@ -373,3 +387,104 @@ func CheckCscSuccessfulDeletion(client test.FrameworkClient, cscName, namespace,
 
 	return nil
 }
+
+// RestartMarketplace scales the marketplace deployment down to zero and then scales
+// it back up to it's original number of replicas, and waits for a successful deployment.
+func RestartMarketplace(client test.FrameworkClient, namespace string) error {
+	marketplace := &apps.Deployment{}
+	err := client.Get(context.TODO(), types.NamespacedName{Name: "marketplace-operator", Namespace: namespace}, marketplace)
```
aravindhp: To address the problem of this test always failing when you run with `operator-sdk up local`, you could make this test contingent on the presence of the `marketplace-operator` deployment.

cmd/manager/main.go
```diff
@@ -107,6 +109,21 @@ func main() {
...
+	}
+
+	// Perform migration logic to upgrade cluster from 4.1.z to 4.2.z
+	migrator := migrator.NewMigrator(client)
+	err = migrator.Migrate(namespace)
+	for err != nil {
```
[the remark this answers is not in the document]
ecordell: I was thinking we would attempt, repeatedly, in the foreground, until we either give up or succeed.

This is what we're doing for a similar migration in olm: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/cmd/olm/cleanup.go#L23

The difference here is that marketplace can technically keep working if this migration doesn't succeed. But I would think we should still block here because we know that if this migration doesn't run, there will be a problem for a future update (unless we intend to keep CSCs around forever?) Correct me if I'm wrong.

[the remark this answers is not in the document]
galletti94: I agree with this approach. I just want to clarify that we do not delete all CSCs as part of this migration - only the CSCs created by the operator or by the UI. In the case of user-created CSCs, the migration code will just ignore their existence (note the operator will still be able to act on these properly). Although we are able to delete them if we want to, the argument was made that it was a bad user experience to have resources created in 4.1 disappear in 4.2. Anyway if the migration code doesn't run I believe we can still manage to delete all CSCs as part of a 4.2 to 4.3 migration if we decide to depricate CSCc in 4.3+.

galletti94:
@aravindhp I addressed your comments. PTAL when you get a chance

pkg/builders/builders.go
```diff
@@ -43,3 +43,39 @@ func GetOwnerLabel(name, namespace, owner string) map[string]string {
 		return map[string]string{}
 	}
 }
+
+// CloneAndRemoveOwnerLabels removes the owner labels from a set of labels
+func CloneAndRemoveOwnerLabels(labels map[string]string, owner string) map[string]string {
```
aravindhp: Nit: I mentioned this before in https://github.com/operator-framework/operator-marketplace/pull/223#discussion_r305064649. I don't understand why you need to write your own helper when there is  [labels.CloneAndRemoveLabel](https://godoc.org/k8s.io/kubernetes/pkg/util/labels#CloneAndRemoveLabel).

test/helpers/helpers.go
aravindhp: I still think there is a risk of a test flake here if you are not waiting for scale down.

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,264 @@
...
+				}
+			}
+			// Update the Subscription to reference the datastore CatalogSource
+			subscription.Spec.CatalogSource = datastoreCs.GetName()
+			subscription.Spec.CatalogSourceNamespace = datastoreCs.GetNamespace()
+			// Remove the owner labels from the subscription
+			subscription.Labels = labelsutil.CloneAndRemoveLabel(
+				labelsutil.CloneAndRemoveLabel(subscription.GetLabels(), builders.CscOwnerNameLabel),
+				builders.CscOwnerNamespaceLabel)
+			err = m.client.Update(context.TODO(), &subscription)
```
ecordell: Any reason not to patch? OLM also reads and writes Subscriptions, so if we can reduce the chance of conflicts that would be ideal.

galletti94:
@aravindhp I updated based on your feedback

Gopkg.toml
```diff
@@ -97,3 +97,7 @@ required = [
 [[constraint]]
   branch = "master"
   name = "github.com/openshift/client-go"
+
+[[constraint]]
+  name = "k8s.io/kubernetes"
+  version = "1.15.1"
```
aravindhp: This will be a nack from @shawn-hurley. 

We cannot move to 1.15 before rest of OpenShift. We actually should be pinning to  kube version to 1.14. Do you need 1.15.1 to use `labels.CloneAndRemoveLabel()`? I see it in the [vendor folders](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/vendor/k8s.io/kubernetes/pkg/util/labels/labels.go#L41) in OLM and they are pinned to 1.14. 

@ecordell please correct if I am wrong?

test/testsuites/migrationtests.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+// MigrationTests is a test suite that ensures the following:
+// * All stale datastore CatalogSourceConfigs created by marketplace after the creation of
+// OperatorSources in a 4.1.x cluster are deleted when the cluster is migrated to openshift 4.2.0.
+// * Installed CatalogSourceConfigs created during operator installation in a 4.1 cluster are deleted.
+// * Existing Subscriptions are updated to reference the datastore CatalogSources instead
+// of installed CatalogSources created during operator installation in a openshift 4.1.x cluster.
+func MigrationTests(t *testing.T) {
+	t.Run("catalogsourceconfigs-are-cleaned-up", testCatalogSourceConfigsCleanedUp)
```
ecordell: We should test the negative as well - make sure that a Subscription that doesn't point to a CSC's CatalogSource is not touched in any way.

pkg/builders/builders.go
aravindhp: If you end up not using this, then it should be removed.

Gopkg.toml
galletti94: Ah that's my bad I just did `dep ensure --add ...` - what is the correct way of adding this dependency such that it doesn't bump to 1.15?

pkg/migrator/migrator.go
galletti94: I see - I didn't think of that, thanks for pointing it out!

aravindhp: controller-runtime and by extension the SDK client does not provide `patch`. Please see https://github.com/operator-framework/operator-sdk/issues/1077. We would have to move using the raw REST client to do that. Might too much of a change at this point.

Gopkg.toml
aravindhp:
```
dep ensure --add --help
Usage: dep ensure [-update | -add] [-no-vendor | -vendor-only] [-dry-run] [-v] [<spec>...]

Project spec:

  <import path>[:alt source URL][@<constraint>]


Ensure gets a project into a complete, reproducible, and likely compilable state:

  * All imports are fulfilled
  * All rules in Gopkg.toml are respected
  * Gopkg.lock records immutable versions for all dependencies
  * vendor/ is populated according to Gopkg.lock

Ensure has fast techniques to determine that some of these steps may be
unnecessary. If that determination is made, ensure may skip some steps. Flags
may be passed to bypass these checks; -vendor-only will allow an out-of-date
Gopkg.lock to populate vendor/, and -no-vendor will update Gopkg.lock (if
needed), but never touch vendor/.

The effect of passing project spec arguments varies slightly depending on the
combination of flags that are passed.


Examples:

  dep ensure                                 Populate vendor from existing Gopkg.toml and Gopkg.lock
  dep ensure -add github.com/pkg/foo         Introduce a named dependency at its newest version
  dep ensure -add github.com/pkg/foo@^1.0.1  Introduce a named dependency with a particular constraint
```
:point_up:

aravindhp: Please make it a separate commit if possible.

galletti94:
@aravindhp @ecordell I addressed your comments. PTAL when you get a chance. I added a separate commit for the dependency addition

pkg/migrator/migrator.go
ecordell: Made a note to bring this up next sync - I think patch is pretty important for reducing collisions.

test/helpers/helpers.go
```diff
@@ -373,3 +391,168 @@ func CheckCscSuccessfulDeletion(client test.FrameworkClient, cscName, namespace,
...
+// CheckSubscriptionNotUpdated checks that a user created subscription
+// was not updated during migration.
+func CheckSubscriptionNotUpdated(client test.FrameworkClient, name, namespace string) error {
+	subscription := &olm.Subscription{}
+	specSource := TestInstalledCscPublisherName + namespace
+	err := client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: namespace}, subscription)
+	if err != nil {
+		return err
+	}
+	if subscription.Spec.CatalogSource != specSource {
```
ecordell: Is it reasonable to check resourceversion here? Or will something else potentially (validly) update the subscription?

PR #224, opened 2019-07-17
[opsrc] Add watches for deletion of child resources

aravindhp:
Watch for deletion of CatalogSources, Deployments, Services, ServiceAccounts, Roles and RoleBindings and recreate them if they belong to a OperatorSource.
- Add a watch that uses EnqueueRequestsFromMapFunc with a predicate that filters out non-deletion events
- Deletion predicate function only returns true if the event belongs to a child resource
- Add a mapper function that figures out the reconciliation request given the resource being deleted
- Make newReconciler return *ReconcileOperatorSource to allow the mapper to have access to the client
- Update the succeeded reconciler to check if the child resources have been deleted and set the phase to configuring if it has

### DIFF
--- a/pkg/controller/operatorsource/operatorsource_controller.go
+++ b/pkg/controller/operatorsource/operatorsource_controller.go
@@ -6,6 +6,7 @@ import (
 	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
 	operatorsourcehandler "github.com/operator-framework/operator-marketplace/pkg/operatorsource"
 	"github.com/operator-framework/operator-marketplace/pkg/status"
+	"github.com/operator-framework/operator-marketplace/pkg/watches"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -22,8 +23,8 @@ func Add(mgr manager.Manager) error {
 	return add(mgr, newReconciler(mgr))
 }
 
-// newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager) reconcile.Reconciler {
+// newReconciler returns a new ReconcileOperatorSource
+func newReconciler(mgr manager.Manager) *ReconcileOperatorSource {
 	client := mgr.GetClient()
 	handler := operatorsourcehandler.NewHandler(mgr, client)
 	return &ReconcileOperatorSource{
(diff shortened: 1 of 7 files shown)


pkg/watches/helpers.go
```diff
@@ -171,3 +190,20 @@ func getCscOwnerKey(labels map[string]string) *client.ObjectKey {
 
 	return &client.ObjectKey{Namespace: ownerNamespace, Name: ownerName}
 }
+
+// getOpSrcOwnerKey checks for the OperatorSource owner labels within the labels
+// of the child resource  and computes the key if present. It returns nil if
```
awgreene: nit - there is an extra space between `resource` and `and`.

pkg/controller/operatorsource/operatorsource_controller.go
```diff
@@ -23,7 +24,7 @@ func Add(mgr manager.Manager) error {
 }
 
 // newReconciler returns a new reconcile.Reconciler
```
awgreene: This comment is out of sync with the function.

pkg/controller/operatorsource/operatorsource_controller.go
```diff
@@ -33,7 +34,7 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
```
awgreene: This comment is out of sync with the function.

pkg/watches/helpers.go
```diff
@@ -155,6 +159,21 @@ func cscDeleteFunc(e event.DeleteEvent) bool {
 	return true
 }
 
+// opsrcDeleteFunc is the predicate function for checking delete events. It returns
+// true only if the object's owner is a OperatorSource.
```
awgreene: nit - should be `an OperatorSource.`

pkg/watches/helpers.go
aravindhp: This is also present in the `CatalogSourceConfig` code. I will fix it there too.

pkg/controller/operatorsource/operatorsource_controller.go
aravindhp: This is also present in the `CatalogSourceConfig` code. I will fix it there too.

pkg/controller/operatorsource/operatorsource_controller.go
aravindhp: This is also present in the `CatalogSourceConfig` code. I will fix it there too.

aravindhp:
@awgreene I have addressed your comments. PTAL.

PR #225, opened 2019-07-18
Adding build support for archs other than amd64

Pensu:
This modifies the build.sh script to add the build support for archs other than amd64.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -15,4 +15,4 @@ PROJECT_NAME="marketplace-operator"
 REPO_PATH="github.com/operator-framework/operator-marketplace/"
 BUILD_PATH="${REPO_PATH}/cmd/manager"
 echo "building "${PROJECT_NAME}"..."
-GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o ${BIN_DIR}/${PROJECT_NAME} $BUILD_PATH
+GOOS=linux GOARCH=$(go env GOARCH) CGO_ENABLED=0 go build -o ${BIN_DIR}/${PROJECT_NAME} $BUILD_PATH


aravindhp:
Is this required for OpenShift 4.2?

Pensu:
@aravindhp Yes, we would love to get this in 4.2.

aravindhp:
Love != required :-) My question: Is not having multi-arch support a blocker for 4.2? Just trying to ensure that we are not introducing an issue a day before code freeze.  It is not like we have CI that tests multiple architecture.
/cc @shawn-hurley @ecordell

Pensu:
> Love != required :-) My question: Is not having multi-arch support a blocker for 4.2? Just trying to ensure that we are not introducing an issue a day before code freeze. It is not like we have CI that tests multiple architecture.
> /cc @shawn-hurley @ecordell

@aravindhp Ah, ok. Not sure about the blocker part, but as far as CI is concerned, this is not going to impact the flow/tests in any way. I am just replacing hardcoded amd64 with the arch of the platform where we are building the binaries. This will help create binaries for other platforms like most of the OpenShift projects already have (the ones that don't have are listed in https://github.com/openshift/origin/issues/23385). Plus the testing can be done by CI systems one they have multi arch support. So, to answer a part of your question, this is not going to introduce any issues for the CI/test flow at all. :)

ecordell:
This seems low risk, _if_ we know for sure that GOARCH will be set properly as part of the ART build. Do we know that?

Pensu:
> This seems low risk, _if_ we know for sure that GOARCH will be set properly as part of the ART build. Do we know that?

@ecordell If you running in a go env, GOARCH will give you the arch value (amd64/ppc64le/s390x etc) regardless of the platform it's running on. Lot of openshift projects are using this variable to run builds, case in point, https://github.com/openshift/cluster-network-operator/blob/master/hack/build-go.sh#L26. For ART builds, is there a way to verify what GOARCH will output in those environments?

ecordell:
Let's wait for feature freeze, file a BZ, then get a build going through ART. That way this can still make 4.2 but won't inadvertently block the initial build.

Pensu:
Sure @ecordell. Thanks. Any idea when the feature freeze is?

yselkowitz:
The point is that GOOS/GOARCH should *not* be manually set so that go will compile natively for the platform on which it is running.  This is _required_ for 4.2 to allow multi-arch support.

ecordell:
Closed in favor of #236

PR #226, opened 2019-07-19
[defaults] Move ownership of default OperatorSources

aravindhp:
- [x] Add command line parameter to input a defaults folder
- [x] Create the defaults if present
- [x] Don't allow deletion of defaults
- [x] Prevent updates to the spec of default OperatorSources
- [x] Move out default OperatorSources from manifests folder
- [x] Update Dockerfile to create a new "defaults" directory in the image and pass that as a command line parameter
- [x] Write e2e test


pkg/defaults/defaults.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/api/errors"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+var (
+	// Dir is the directory where the default OperatorSource definitions are
+	// placed on disk. It will be empty if defaults are not required.
+	Dir string
```
ecordell: non-blocking: instead of a global `dir` for the package, `defaults` could have a `dir` member. In the future if we needed multiple types of defaults to apply we can just make a new default tracker.

pkg/defaults/defaults.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+	return utilerrors.NewAggregate(allErrors)
+}
+
+// PopulateTracker populates the defaultsTracker on initialization
+func PopulateTracker() error {
+	// Default directory has not been specified
+	if Dir == "" {
+		return nil
```
ecordell: should this return an error (or even a panic)? it seems like if we're ever calling this without a default directory, that's a bug.

pkg/defaults/defaults.go
aravindhp: Should be easy enough to do. I anyhow need to rebase now that #228 has merged. Let me see if I can quickly do it or I will leave it as it is.

pkg/defaults/defaults.go
aravindhp: The [default value](https://github.com/operator-framework/operator-marketplace/pull/226/commits/cac4f5c349a4d3628f028857e09f4e8307fa6134#diff-9e63f7b0791edcb4a53c944552b88970R57) of `Dir` is blank so it works. I tested this out without passing a default and it worked. I just could not figure out an easy way to add an e2e.

pkg/defaults/defaults.go
aravindhp: Shoot. I forgot to put a hold on the PR and it merged. Commit history now looks a bit weird with two commits fixing the `isEqual()`. Sorry about that. I have already made the change you requested and will add it to the next PR.

PR #227, opened 2019-07-19
Handle OperatorHub cluster config resource

aravindhp:
- [config] Handle OperatorHub config resources
- [defaults] Refactor
- Generalize config API discovery 

TODO:
- [x] Add new config API: operatorhub.configs.openshift.io
- [x] Generate boilerplate code
- [x] Flesh out the handler
- [x] e2e tests
- [x] Update commit messages
- [x] Ensure upstream compatibility
- [x] Move type to OpenShift API repo https://github.com/openshift/api/pull/388
- [x] Add option to disable all default sources https://github.com/openshift/api/pull/409
- [x] client-gen https://github.com/openshift/client-go/pull/114, https://github.com/openshift/client-go/pull/115
- [x] Add CRD and empty config resource to CCO https://github.com/openshift/cluster-config-operator/pull/76/


aravindhp:
WIP should be removed only after https://github.com/openshift/api/pull/388 and https://github.com/openshift/cluster-config-operator/pull/76 has merged.

cmd/manager/main.go
```diff
@@ -69,13 +70,8 @@ func main() {
...
 
-	// Set proxy API availability
-	k8sInterface, err := kubernetes.NewForConfig(cfg)
-	if err != nil {
-		log.Fatal(err)
-	}
-
-	err = proxy.SetProxyAvailability(k8sInterface.Discovery())
+	// Set OpenShift config API availability
+	err = configv1.SetConfigAPIAvailability(cfg)
```
kevinrizza: Thinking out loud, this is the second OpenShift specific api we query for. Do we possibly want to combine the checks for this and ClusterOperator into a more thoughtful mechanism? This is outside the scope of this PR though.

pkg/controller/operatorhub/operatorhub_controller.go
```diff
@@ -0,0 +1,129 @@
...
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	// Needed because SDK does not get the GVK
+	instance.EnsureGVK()
```
kevinrizza: Based on your API pr, it seems like this might no longer be needed?

cmd/manager/main.go
awgreene: This is actually the third with the addition of the proxy api.

awgreene: I'd approve standardizing how we check for APIs.

aravindhp: I agree with standardizing it and I am made the first inroad here while constraining it to the scope of the functionality being added. I did not want to try and rewrite all of it here.

pkg/controller/operatorhub/operatorhub_controller.go
aravindhp: Correct, I will remove it.

pkg/operatorhub/operatorhub.go
```diff
@@ -0,0 +1,65 @@
...
+
+// OperatorHub is the interface to interact with the OperatorHub configuration in
+//  memory.
+type OperatorHub interface {
+	Get() map[string]bool
+	Set(hub configv1.OperatorHubSpec)
+}
+
+// New returns the singleton instance of HubConfig
+func New() OperatorHub {
```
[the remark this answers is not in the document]
aravindhp: Sure. I will go with `GetSingleton()` give `OperatorHub` is redundant as per go conventions. BTW, in `Defaults` it is actually instantiating a new instance, so that can stay as `New()`

aravindhp:
@kevinrizza I have made the changes you requested. PTAL.

ecordell:
/lgtm

Verified all on a cluster as well. Thanks @aravindhp!

One question (no changes required 😄) - will there be a convenient way to tell if a cluster is installed "disconnected"? (I imagine there must be, since phone-home services like telemeter won't work in that state).

If so, a nice followup would be to detect that state and automatically disable the defaults.

aravindhp:
>Verified all on a cluster as well. Thanks @aravindhp!

Thanks for testing :+1: 

>will there be a convenient way to tell if a cluster is installed "disconnected"?

Not from the cluster config object but the operator should be able to glean that if it finds default sources not being able to connect to Quay.

>I imagine there must be, since phone-home services like telemeter won't work in that state

I have seen a disconnected cluster calling home. So apparently customers could choose to open some access.

 >If so, a nice followup would be to detect that state and automatically disable the defaults.

Is this desirable? Won't it be better for an external entity or admin to do this? Do we want the operator that handles changes to the cluster config to also make changes to it?

PR #229, opened 2019-07-23
Bug 1711070 - Set container resource requests

kevinrizza:
* Set container resource requests for operator deployment
* Set container resource requests for registry pod deployments

https://bugzilla.redhat.com/show_bug.cgi?id=1711070

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -53,6 +53,10 @@ spec:
             httpGet:
               path: /healthz
               port: 8080
+          resources:
+            requests:
+              cpu: 10m
+              memory: 50Mi
           env:
             - name: WATCH_NAMESPACE
               valueFrom:
--- a/deploy/upstream/08_operator.yaml
+++ b/deploy/upstream/08_operator.yaml
@@ -35,6 +35,10 @@ spec:
             httpGet:
               path: /healthz
               port: 8080
+          resources:
+            requests:
+              cpu: 10m
+              memory: 50Mi
           env:
             - name: WATCH_NAMESPACE
               valueFrom:
--- a/manifests/08_operator.yaml
+++ b/manifests/08_operator.yaml
@@ -53,6 +53,10 @@ spec:
             httpGet:
               path: /healthz
               port: 8080
+          resources:
+            requests:
+              cpu: 10m
+              memory: 50Mi
           env:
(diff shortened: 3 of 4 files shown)


kevinrizza:
Putting a hold on this pr while I wait for a response regarding the magic numbers for these resource limits:

https://bugzilla.redhat.com/show_bug.cgi?id=1711070#c7

ecordell:
@kevinrizza What do you think of this plan:

Pick requests for `marketplace-operator`:
 - Something around what current cpu usage is
 - Breathing room for memory request, maybe 120% of what current memory usage is with the default operatorsources

Pick a default request for the registry pods that marketplace spins up:
 - Something around what current cpu usage is
 - Breathing room for memory request, maybe 120% of whatever the max is for one of the default registry pods.

Then, in the future, we can:
- remove the appregistry caching from marketplace-operator, which will make its memory relatively static (and lower than tody)
- remove actual content storage from operator-registry, so that size doesn't fluctuate as much and we can have some reasonable defaults

Having any values for resource requests are better than what we have today, and we can continue to refine over time. Right now the scheduler doesn't know anything about these pods.

kevinrizza:
> @kevinrizza What do you think of this plan:
> 
> Pick requests for `marketplace-operator`:
> 
> * Something around what current cpu usage is
> * Breathing room for memory request, maybe 120% of what current memory usage is with the default operatorsources
> 
> Pick a default request for the registry pods that marketplace spins up:
> 
> * Something around what current cpu usage is
> * Breathing room for memory request, maybe 120% of whatever the max is for one of the default registry pods.
> 
> Then, in the future, we can:
> 
> * remove the appregistry caching from marketplace-operator, which will make its memory relatively static (and lower than tody)
> * remove actual content storage from operator-registry, so that size doesn't fluctuate as much and we can have some reasonable defaults
> 
> Having any values for resource requests are better than what we have today, and we can continue to refine over time. Right now the scheduler doesn't know anything about these pods.

This makes sense to me for now. I did some cursory memory testing with a cluster in AWS and set these defaults:

50mi for the operator deployment
100mi for all registry deployments

I updated the pr accordingly. @ecordell @tkashem @awgreene Please review.

ecordell:
@kevinrizza could you do a quick rebase on this?

PR #230, opened 2019-07-24
Bug 1732760: Ensure cscs not deleted on operator restart

galletti94:
CatalogSourceConfigs that were not created by an OperatorSource or by
the UI as part of the 4.1 install flow are deleted on operator restart.
This is because a labelSelector was malformed and hence ignored. This
fixes that issue and adds a test case for this scenario.

### DIFF
--- a/pkg/migrator/migrator.go
+++ b/pkg/migrator/migrator.go
@@ -7,11 +7,11 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/operator-framework/operator-marketplace/pkg/builders"
-
 	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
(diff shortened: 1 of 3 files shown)


pkg/migrator/migrator.go
```diff
@@ -156,7 +156,7 @@ func (m *migrator) deleteInstalledCscs(cscs []types.NamespacedName, operatorName
 func (m *migrator) deleteDatastoreCscs(operatorNamespace string) []error {
 	var errors []error
 	options := &client.ListOptions{}
-	options.SetLabelSelector(fmt.Sprintf("opsrc-datastore: \"true\""))
+	options.SetLabelSelector(fmt.Sprintf("opsrc-datastore"))
```
kevinrizza: Does this need to be hardcoded at all? Can we make this a const? I know at one time before we removed all this stuff it was part of the API package.

galletti94: oh duh

galletti94: good catch

galletti94: addressed

PR #231, opened 2019-07-24
Bug 1722096:  fix panic when scale down of registry deployment fails

ecordell:
This should prevent the panic - `deployment` was being shadowed, and on timeout will be set to `nil`. This removes the shadowing so that the name comes from the non-nil `deployment` defined above.

I would typically add an e2e but I would need to rewrite a chunk of this to do that (otherwise the e2e test will always take 1+ minutes), and this section seems to have changed significantly between 4.1 and 4.2. But if anyone sees a convenient way to test this for 4.1, I'm all ears :)

### DIFF
--- a/pkg/catalogsourceconfig/registry.go
+++ b/pkg/catalogsourceconfig/registry.go
@@ -126,7 +126,7 @@ func (r *registry) ensureDeployment(appRegistries []string, needServiceAccount b
 
 		// Wait for the deployment to scale down. We need to get the latest version of the object after
 		// the update, so we use the object returned here for scaling up.
-		if deployment, err = r.waitForDeploymentScaleDown(2*time.Second, 1*time.Minute); err != nil {
+		if _, err = r.waitForDeploymentScaleDown(2*time.Second, 1*time.Minute); err != nil {
 			r.log.Errorf("Failed to scale down Deployment %s : %v", deployment.GetName(), err)
 			return err
 		}


pkg/catalogsourceconfig/registry.go
```diff
@@ -126,7 +126,7 @@ func (r *registry) ensureDeployment(appRegistries []string, needServiceAccount b
 
 		// Wait for the deployment to scale down. We need to get the latest version of the object after
 		// the update, so we use the object returned here for scaling up.
-		if deployment, err = r.waitForDeploymentScaleDown(2*time.Second, 1*time.Minute); err != nil {
+		if _, err = r.waitForDeploymentScaleDown(2*time.Second, 1*time.Minute); err != nil {
```
aravindhp: Won't you have the same issue with other resources like  `roles`, `rolebindings` etc in the scenario where the update fails?

ecordell: I don't see anything in those functions that could cause a similar issue - did I miss something?

aravindhp:
```
		role.Rules = getRules()
		err = r.client.Update(context.TODO(), role)
		if err != nil {
			r.log.Errorf("Failed to update Role %s : %v", role.GetName(), err)
			return err
		}
		r.log.Infof("Updated Role %s", role.GetName())
	}
```
Would an error in the update call above result in `role` becoming `nil`? If that won't happen, then the fix you have made will suffice.

ecordell: No, the issue with `waitForDeployment` is that it explicitly returns `nil, err` if there's an error (which happens when there's a timeout). The client calls won't nil out a non-nil reference that's passed in.

jwforres:
eye-balling this it looks pretty low-risk and fixes a panic, given it sounds like the code has already diverged significantly in this area in 4.2, we don't get much benefit from soak testing. In lieu of the soak testing, @ecordell can you confirm you've manually run with this code change in place on a 4.1 cluster and triggered the two code paths here (happy path vs deployment timeout)

ecordell:
My attempts to replicate this in a cluster:

- Patching catalogsourceconfig to trigger a rollout, followed by continual patching of the `replicas` to 1
- Pausing the deployment, then patching catalogsourceconfig
- Patching catalogsourceconfig, and waiting various amounts of time between 1-3 seconds to patch the deployment `replicas` to 1
- Patching catalogsourcesconfig and changing the serviceaccount permissions to not be allowed to `get` deployments

I've been unsuccessful in triggering this codepath in a cluster with any of these, I would have to be very lucky, and my attempts to hit the probability jackpot haven't worked. Open to any additional ideas for verifying.

ecordell:
Validated in a 4.1 cluster:

```
INFO[0125] Waiting for scale down of Deployment certified-operators (1/0)  name=certified-operators targetNamespace=openshift-marketplace type=CatalogSourceConfig
INFO[0125] Waiting for scale down of Deployment certified-operators (1/0)  name=certified-operators targetNamespace=openshift-marketplace type=CatalogSourceConfig
ERRO[0125] Failed to scale down Deployment certified-operators : timed out waiting for the condition  name=certified-operators targetNamespace=openshift-marketplace type=CatalogSourceConfig
INFO[0126] Reconciling CatalogSourceConfig openshift-marketplace/community-operators
```

Since my attempts to black-box test (described above) failed, I removed the code that changes the deployment replicas before the `wait.Poll` and ran that against the 4.1 cluster. This is where the BZ shows the operator panic, but execution continues after the failed `Poll`.

PR #232, opened 2019-07-25
Bug 1732577: Handle delete events while operator is restarting

kevinrizza:
Problem:
An edge case bug was added when the default opsrc feature was added.
If a default opsrc was marked for deletion but the finalizer was
not removed by the operator before the operator restarted, then the
`defaults.EnsureAll()` method would attempt to create it. In that case,
the client would return an error and the operator would exit.

Solution:
When EnsureAll is called on startup, also ensure that there are no
finalizers still included on the operator source. If they are still
there, ignore it and allow the normal reconciliation cycle to take care
of deleting and recreating it.

Additionally, once the finalizer is removed at the end of the
reconciliation loop, wait for the opsrc to be cleaned up by the garbage
collector before recreating it.

https://bugzilla.redhat.com/show_bug.cgi?id=1732577

### DIFF
--- a/pkg/apis/operators/shared/shared.go
+++ b/pkg/apis/operators/shared/shared.go
@@ -37,6 +37,19 @@ func RemoveFinalizer(objectMeta *metav1.ObjectMeta, deletingFinalizer string) {
 	return
 }
 
+// HasFinalizer checks to see if the finalizer exists in the object's ObjectMeta.
+func HasFinalizer(objectMeta *metav1.ObjectMeta, expectedFinalizer string) bool {
+	finalizerExists := false
+	for _, finalizer := range objectMeta.Finalizers {
+		if finalizer == expectedFinalizer {
+			finalizerExists = true
+			break
+		}
+	}
+
+	return finalizerExists
+}
+
 // IsObjectInOtherNamespace returns true if the namespace is not the watched
 // namespace of the operator. An false, error will be returned if there was an
 // error getting the watched namespace.
--- a/pkg/controller/operatorsource/operatorsource_controller.go
+++ b/pkg/controller/operatorsource/operatorsource_controller.go
@@ -2,8 +2,11 @@ package operatorsource
 
 import (
 	"context"
+	"time"
 
 	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	"github.com/operator-framework/operator-marketplace/pkg/defaults"
+	"github.com/operator-framework/operator-marketplace/pkg/operatorhub"
 	operatorsourcehandler "github.com/operator-framework/operator-marketplace/pkg/operatorsource"
 	"github.com/operator-framework/operator-marketplace/pkg/status"
 	"github.com/operator-framework/operator-marketplace/pkg/watches"
@@ -79,6 +82,15 @@ func (r *ReconcileOperatorSource) Reconcile(request reconcile.Request) (reconcil
 	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
 	if err != nil {
 		if errors.IsNotFound(err) {
+			// If the object is not found but is one of the default operator sources, let's ensure it gets recreated.
+			if defaults.IsDefaultSource(request.Name) {
+				err = defaults.New(defaults.GetGlobalDefinitions(), operatorhub.GetSingleton().Get()).Ensure(r.client, request.Name)
+				if err != nil {
+					// If we run into an error when attempting to ensure the default, let's requeue and try again
+					return reconcile.Result{}, err
+				}
+			}
+
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
@@ -92,9 +104,12 @@ func (r *ReconcileOperatorSource) Reconcile(request reconcile.Request) (reconcil
 	// Needed because sdk does not get the gvk
 	instance.EnsureGVK()
 
-	err = r.OperatorSourceHandler.Handle(context.TODO(), instance)
+	requeue, err := r.OperatorSourceHandler.Handle(context.TODO(), instance)
 	if err != nil {
 		return reconcile.Result{}, err
+	}
+	if requeue {
+		return reconcile.Result{RequeueAfter: time.Second * 5}, nil
 	}
 
 	return reconcile.Result{}, nil
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -240,7 +240,7 @@ func ensureAbsent(client wrapper.Client, def v1.OperatorSource, cluster *v1.Oper
 // ensurePresent ensure that that the default OperatorSource is present on the cluster
 func ensurePresent(client wrapper.Client, def v1.OperatorSource, cluster *v1.OperatorSource) error {
 	// Create if not present or is deleted
-	if cluster.Name == "" || !cluster.ObjectMeta.DeletionTimestamp.IsZero() {
+	if cluster.Name == "" || (!cluster.ObjectMeta.DeletionTimestamp.IsZero() && len(cluster.Finalizers) == 0) {
 		err := client.Create(context.TODO(), &def)
 		if err != nil {
 			return err
--- a/pkg/operatorsource/deleted.go
+++ b/pkg/operatorsource/deleted.go
@@ -3,13 +3,13 @@ package operatorsource
 import (
 	"context"
 
+	"github.com/operator-framework/operator-marketplace/pkg/grpccatalog"
+
 	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/shared"
 	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
 	wrapper "github.com/operator-framework/operator-marketplace/pkg/client"
 	"github.com/operator-framework/operator-marketplace/pkg/datastore"
 	"github.com/operator-framework/operator-marketplace/pkg/defaults"
-	"github.com/operator-framework/operator-marketplace/pkg/grpccatalog"
-	"github.com/operator-framework/operator-marketplace/pkg/operatorhub"
 	"github.com/operator-framework/operator-marketplace/pkg/phase"
 	log "github.com/sirupsen/logrus"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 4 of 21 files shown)


pkg/defaults/defaults.go
```diff
@@ -130,8 +130,11 @@ func processOpSrc(client wrapper.Client, defaultOpsrc v1.OperatorSource) error {
 		return err
 	}
 
-	// Create if not present or is deleted
-	if errors.IsNotFound(err) || !opsrcCluster.ObjectMeta.DeletionTimestamp.IsZero() {
+	// Create if not present OR the object exists AND it was cleaned up already
+	// (marked for deletion, finalizer removed) but the garbage collector hasn't gotten
```
aravindhp: Nit: I suggest making it two sentences for easier understanding.  `..., finalizer removed). However if it was marked for deletion and the garbage collected hasn't gotten to it yet, i.e. the finalizers have not been removed, do nothing here and allow the deletion reconciler to remove the finalizer and recreate the object.`

test/helpers/helpers.go
```diff
@@ -456,6 +475,50 @@ func RestartMarketplace(client test.FrameworkClient, namespace string) error {
 	return nil
 }
 
+// ScaleDownMarketplace scales the marketplace deployment down to zero
+func ScaleDownMarketplace(client test.FrameworkClient, namespace string) error {
```
aravindhp: Nit: I suggest using this in `RestartMarketplace()` otherwise it is never going to get updated.

test/testsuites/defaultopsrctests.go
```diff
@@ -103,6 +105,57 @@ func testUpdateDefaultOpSrc(t *testing.T) {
 	assert.ObjectsAreEqualValues(defaultOpSrc.Spec, clusterOpSrc.Spec)
 }
 
+func testDeleteDefaultOpsrcWhileStopped(t *testing.T) {
```
aravindhp: Nit: please add doc for the test.

test/helpers/helpers.go
```diff
@@ -456,6 +475,50 @@ func RestartMarketplace(client test.FrameworkClient, namespace string) error {
...
+	err = WaitForDeploymentScaled(client, "marketplace-operator", namespace, zero)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// ScaleUpMarketplace scales the marketplace deployment up to one
+func ScaleUpMarketplace(client test.FrameworkClient, namespace string) error {
```
galletti94: What do you think about merging this and the above functions into a single `ScaleMarketplace` function that takes in the scale factor?

aravindhp: A good abstraction is to have a lower level helper that takes scale as an input but have higher level ScaleUp and ScaleDown wrappers that call this function.

galletti94: I like that idea!

kevinrizza: Yeah, thinking about this more there should just be one function `ScaleMarketplace` that just takes the number of replicas. Then to scale up and down you call `ScaleMarketplace(0)` wait for it to return then call `ScaleMarketplace(1)`. Then the `RestartMarketplace` method can just call scale the second time with the previous replica size in order to preserve the behavior.

kevinrizza:
Rebased and updated the test helpers based on conversation above. Please take another look.

@aravindhp @galletti94 @tkashem

test/helpers/helpers.go
```diff
@@ -428,31 +447,46 @@ func RestartMarketplace(client test.FrameworkClient, namespace string) error {
...
+	initialReplicas := marketplace.Spec.Replicas
 	zero := int32(0)
 	marketplace.Spec.Replicas = &zero
 	err = client.Update(context.TODO(), marketplace)
 	if err != nil {
 		return err
 	}
 	// Wait for deployment to scale down
-	err = WaitForDeploymentScaled(client, "marketplace-operator", namespace, zero)
+	err = ScaleMarketplace(client, namespace, int32(0))
```
aravindhp: You need to wait for it scale down. If the scale down and up happens quickly, the net result will be a no-op as in no action will be taken.

kevinrizza: The `ScaleMarketplace()` method calls `WaitForDeploymentScaled()`, so it will wait.

pkg/defaults/defaults.go
```diff
@@ -144,8 +144,12 @@ func processOpSrc(client wrapper.Client, defaultOpsrc v1.OperatorSource) error {
 		return err
 	}
 
-	// Create if not present or is deleted
-	if errors.IsNotFound(err) || !opsrcCluster.ObjectMeta.DeletionTimestamp.IsZero() {
+	// Create if not present OR the object exists AND it was cleaned up already
+	// (marked for deletion, finalizer removed). However if it was marked for deletion
+	// and the garbage collector hasn't gotten around to it yet (i.e. the finalizers have
+	// not been removed) do nothing here and allow the deletion reconciler to remove the
+	// finalizer and recreate the object.
```
tkashem: We are waiting for the `finalizer` to be removed, then can we not wait until it is deleted by GC? Are there some edge cases with the deletion logic in marketplace or with the GC we are trying to work around here?

kevinrizza: It's not going to be removed because the object is marked for deletion but the operator isn't running, so the delete reconciler is never called and thus the finalizer won't be removed.

tkashem: okay, `len(opsrcCluster.Finalizers) == 0` is always going to be `true` since `processOpSrc` is always called after `out.RemoveFinalizer()`. https://github.com/operator-framework/operator-marketplace/blob/master/pkg/operatorsource/deleted.go#L77

Maybe that's why you need it? - https://github.com/operator-framework/operator-marketplace/blob/0d9a16a00b97d195487e4f4f9f244e635a19cfe3/pkg/operatorsource/succeeded.go#L83

Just a thought, maybe this is a better location to put the `defaults.Ensure()` call - https://github.com/operator-framework/operator-marketplace/blob/0d9a16a00b97d195487e4f4f9f244e635a19cfe3/pkg/operatorsource/outofsync.go#L63? This ensures that we are always in sync irrespective of the phase.

Please don't bother with my suggestion if your changes works :), I just wanted to understand the issue.

pkg/operatorsource/deleted.go
```diff
@@ -89,6 +94,22 @@ func (r *deletedReconciler) Reconcile(ctx context.Context, in *v1.OperatorSource
 
 	r.logger.Info("Finalizer removed, now garbage collector will clean it up.")
 
+	// Now let's make sure the garbage collector finished its work before we try to recreate it.
```
aravindhp: Shouldn't we do this only if it is a default `OperatorSource`?

kevinrizza: Sorry, this should be held. I'm using this PR as a way of testing against the CI environment.

pkg/defaults/defaults.go
```diff
@@ -169,3 +177,32 @@ func processOpSrc(client wrapper.Client, defaultOpsrc v1.OperatorSource) error {
 
 	return nil
 }
+
+func (d *defaults) WaitForDeleteIfDefault(client wrapper.Client, opsrc *v1.OperatorSource) error {
```
aravindhp: Please add a doc

pkg/defaults/defaults.go
```diff
@@ -169,3 +177,32 @@ func processOpSrc(client wrapper.Client, defaultOpsrc v1.OperatorSource) error {
...
+	logrus.Infof("[defaults] Checking if cleaned up OperatorSource %s is a default", opsrc.Name)
+	_, present := defaultsTracker[opsrc.Name]
```
aravindhp: I am adding a helper called `IsDefaultSource()` that can be used. I suggest waiting for #227 to go in and use that.

pkg/operatorsource/deleted.go
```diff
@@ -89,6 +89,11 @@ func (r *deletedReconciler) Reconcile(ctx context.Context, in *v1.OperatorSource
 
 	r.logger.Info("Finalizer removed, now garbage collector will clean it up.")
 
+	err = defaults.New().WaitForDeleteIfDefault(r.client, in)
```
aravindhp: Please add a comment saying why this is needed.

test/e2e/main_test.go
```diff
@@ -31,9 +31,9 @@ func TestMarketplace(t *testing.T) {
 		t.Run("cluster-operator-status-test-group", testgroups.ClusterOperatorTestGroup)
 		testgroups.EnableProxyTests()
 	}
+	t.Run("migration-test-group", testgroups.MigrationTestGroup)
```
aravindhp: Is there a reason to change the order?

kevinrizza: It's described in the commit message. There's an unrelated bug related to the migration code (that I'm in the process of replicating and then I'll create a bugzilla). This is just to avoid the race condition that causes it -- basically we are restarting the operator in the test I added in this pr and the migration test doesn't always handle itself well when the test operator isn't already serving content. We could also in theory put a wait on the status of the opsrc in the test, but I would rather address that holistically in a separate PR that also fixes the migration bug.

aravindhp: Sorry I am not able to tie in the commit message to the change here. I must be missing something. Anyhow, a comment here would help too as folks reading the code won't look at the commit message immediately.

kevinrizza: Sure, I'll quote below:

```
This also adds a test case to the e2e test suite that ensures that the
operator being stopped won't cause this bug to appear in the future.
An issue with an unrelated test case was found with the e2e tests
exposed by the addition of this test (which scales the operator's
deployment down and back up to simulate the restart). That issue
needs to be addressed holistically, so for now this pr simply reorders
the tests to avoid that race condition.
```

kevinrizza:
@ecordell @tkashem @awgreene I updated the garbage collector wait to be asynchronous, but that made this pr fairly large now. Please take another look.

pkg/controller/operatorsource/operatorsource_controller.go
```diff
@@ -79,6 +82,11 @@ func (r *ReconcileOperatorSource) Reconcile(request reconcile.Request) (reconcil
 	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
 	if err != nil {
 		if errors.IsNotFound(err) {
+			// If the object is not found but is one of the default operator sources, let's ensure it gets recreated.
+			if defaults.IsDefaultSource(request.Name) {
+				err = defaults.New(defaults.GetGlobalDefinitions(), operatorhub.GetSingleton().Get()).Ensure(r.client, request.Name)
```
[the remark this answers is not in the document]
kevinrizza: Good point, updated in latest.

pkg/apis/operators/shared/shared.go
```diff
@@ -37,6 +37,18 @@ func RemoveFinalizer(objectMeta *metav1.ObjectMeta, deletingFinalizer string) {
 	return
 }
 
+// HasFinalizer checks to see if the finalizer exists in the object's ObjectMeta.
+func HasFinalizer(objectMeta *metav1.ObjectMeta, expectedFinalizer string) bool {
+	finalizerExists := false
+	for _, finalizer := range objectMeta.Finalizers {
+		if finalizer == expectedFinalizer {
+			finalizerExists = true
```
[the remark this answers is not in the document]
kevinrizza: +1, updated in latest.

kevinrizza:
@tkashem @awgreene addressed comments. ptal.

PR #233, opened 2019-07-25
Bug 1733454: Specify kind in op registry labels

galletti94:
Every operator registry pod is created with a label
"marketplace.catalogsourceconfig" even though it may be created by an
OperatorSource. This fixes this issue by gating on the kind before
assigning these labels to the operator registry pod.

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -248,10 +248,17 @@ func (r *registry) ensureServiceAccount() error {
 	return nil
 }
 
-// getLabels returns the label that must match between the Deployment's
+// getLabel returns the label that must match between the Deployment's
 // LabelSelector and the Pod template's label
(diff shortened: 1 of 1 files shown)


pkg/registry/registry.go
```diff
@@ -250,8 +250,15 @@ func (r *registry) ensureServiceAccount() error {
 
 // getLabels returns the label that must match between the Deployment's
```
aravindhp: Nit: Given you are in the area please change to `getLabel`.

aravindhp:
/hold

Please open a BZ for this as this cannot be merged without a BZ reference.

galletti94:
@aravindhp I addressed your comment. PTAL when you get a chance

aravindhp:
/hold cancel

@galletti94 lgtm

PR #234, opened 2019-07-26
Bug 1732503: Restore default opsrc spec

kevinrizza:
Problem:
Currently, an issue exists where if a default opsrc is modified and that modification results in a reconciliation loop that puts the opsrc in a
failed state, it does not recover automatically.

Solution:
Rather than rely on the reconciliation loop to run erroneously on these opsrces and then returning them to their original state afterwards, add a filter to prevent updates to the spec that do not match the default. Do this by adding method to the defaults package `RestoreSpecIfDefault` that always overwrites the spec of any of the defaults to what is expected, and call that method at the beginning of the opsrc handler.

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -31,6 +31,7 @@ var (
 type Defaults interface {
 	EnsureAll(client wrapper.Client) error
 	Ensure(client wrapper.Client, opsrcName string) error
+	RestoreSpecIfDefault(in *v1.OperatorSource)
 }
 
 type defaults struct {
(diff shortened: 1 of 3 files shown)


aravindhp:
/hold

This needs a BZ for it to be merged.

kevinrizza:
> /hold
> 
> This needs a BZ for it to be merged.

Yep, didn't finish updating this pr before this comment was made :)

pkg/defaults/defaults.go
```diff
@@ -41,6 +42,17 @@ func New() Defaults {
 	return &defaults{}
 }
 
+func (d *defaults) FilterDefaultOpSrc(in *v1.OperatorSource) {
```
aravindhp: Missing doc. Moreover rather than filtering this seems to be `RestoreSpecIfDefault()`

kevinrizza: Good point, that's a better name than the one I thought of. Name updated and doc added in latest commit.

PR #235, opened 2019-07-31
Bug 1732160: Degrade when default opsrc fails

awgreene:
This PR introduces a change so the marketplace operator will report
that it has degraded when one of the the enabled default
OperatorSources are unable to reach the succeeded phase.

### DIFF
--- a/pkg/operatorhub/operatorhub.go
+++ b/pkg/operatorhub/operatorhub.go
@@ -43,7 +43,16 @@ func GetSingleton() OperatorHub {
 func (o *operatorhub) Get() map[string]bool {
 	o.lock.Lock()
 	defer o.lock.Unlock()
-	return o.current
+
+	// Create a map to return.
+	clone := make(map[string]bool)
+
+	// Populate the newly created map.
+	for key, value := range o.current {
+		clone[key] = value
+	}
+
+	return clone
 }
 
 // Set sets the current configuration based on the spec. If the spec is empty,
--- a/pkg/operatorsource/configuring.go
+++ b/pkg/operatorsource/configuring.go
@@ -11,6 +11,7 @@ import (
 	"github.com/operator-framework/operator-marketplace/pkg/datastore"
 	"github.com/operator-framework/operator-marketplace/pkg/grpccatalog"
 	"github.com/operator-framework/operator-marketplace/pkg/phase"
+	"github.com/operator-framework/operator-marketplace/pkg/status"
 	log "github.com/sirupsen/logrus"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -78,9 +79,9 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *v1.OperatorSo
 
 	r.logger.Infof("Downloading metadata from Namespace [%s] of [%s]", in.Spec.RegistryNamespace, in.Spec.Endpoint)
 
-	metadata, err := r.getManifestMetadata(&in.Spec, in.Namespace)
+	metadata, err, reason := r.getManifestMetadata(&in.Spec, in.Namespace)
 	if err != nil {
-		nextPhase = phase.GetNextWithMessage(phase.Configuring, err.Error())
+		nextPhase = phase.GetNextWithMessageAndReason(phase.Configuring, err.Error(), reason)
 		return
 	}
 
@@ -91,7 +92,7 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *v1.OperatorSo
 		// resolve this situation. As soon as the user pushes new operator
 		// manifest(s) registry sync will detect a new release and will trigger
 		// a new reconciliation.
-		nextPhase = phase.GetNextWithMessage(phase.Failed, err.Error())
+		nextPhase = phase.GetNextWithMessageAndReason(phase.Failed, err.Error(), status.AppRegistryMetadataEmptyError)
 		return
 	}
 
@@ -100,7 +101,7 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *v1.OperatorSo
 	isResyncNeeded, err := r.writeMetadataToDatastore(in, out, metadata)
 	if err != nil {
 		// No operator metadata was written, move to Failed phase.
-		nextPhase = phase.GetNextWithMessage(phase.Failed, err.Error())
+		nextPhase = phase.GetNextWithMessageAndReason(phase.Failed, err.Error(), status.DataStoreWriteError)
 		return
 	}
 
@@ -128,7 +129,7 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *v1.OperatorSo
 	}
 	err = grpcCatalog.EnsureResources(key, in.Spec.DisplayName, in.Spec.Publisher, in.Namespace, in.Name, packages, v1.OperatorSourceKind, labels)
 	if err != nil {
-		nextPhase = phase.GetNextWithMessage(phase.Configuring, err.Error())
+		nextPhase = phase.GetNextWithMessageAndReason(phase.Configuring, err.Error(), status.EnsureResourcesError)
 		return
 	}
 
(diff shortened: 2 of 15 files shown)


awgreene:
Relies on #227 
/hold

pkg/operatorsource/appregistryhelper.go
```diff
@@ -6,7 +6,6 @@ import (
 	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
 	"github.com/operator-framework/operator-marketplace/pkg/appregistry"
 	interface_client "github.com/operator-framework/operator-marketplace/pkg/client"
-
```
kevinrizza: this is noise.

pkg/status/appregistryerror.go
```diff
@@ -0,0 +1,40 @@
+package status
+
+// opsrcErrorConst implements errors related to default OperatorSources.
+type opsrcErrorString string
+
+const (
+	AppRegistryListPackagesError  opsrcErrorString = "AppRegistryListPackagesError"
```
kevinrizza: It seems like you defined all of these types, but they're not used in the control flow at all. If that's the case, why not just craft the strings inline? I thought the purpose of the custom error type here was to differentiate between actual errors from the app-registry API and construction errors. If we are defining them to mean the same thing (from the perspective of the status reporting) then what value do they bring other than having different error strings?

pkg/status/appregistryerror.go
```diff
@@ -0,0 +1,40 @@
+package status
```
kevinrizza: Is this defined in the status package to avoid an import cycle? If we aren't using the type for control flow, why not just not import this type at all and in the status reporting methods just use the generic `Error` type?

pkg/status/appregistryerror.go
awgreene: Crafting the strings in line is a good proposal. The ClusterOperator allows us to report a `Message` and a `Reason`, where:

- The `Message` can be a string
- The `Reason` should be CamelCase.

Given that we want to use both, I suggest keeping the data structure but removing the `consts` as you suggested.

awgreene: Also - `Reason` is available via telemetry where `Message` is not.

pkg/status/appregistryerror.go
awgreene: We want to be able to update the `ClusterOperator` with both a `Reason` and a `Message` as described [here](https://github.com/operator-framework/operator-marketplace/pull/235#discussion_r309719019)

awgreene: There is no reason this is in the status package other than the fact that it is used to report status.

pkg/status/status.go
```diff
@@ -88,6 +91,26 @@ func SendSyncMessage(err error) {
...
+}
+
+// ReportOperatorSourceSuccess is used to report that an enabled default OperatorSource has entered the succeeded state.
+func ReportOperatorSourceSuccess(name string) {
+	// Do not report the success if the coAPI is not available or it is not an enabled default OperatorSource.
+	if instance == nil || instance.coAPINotPresent || !defaults.IsDefaultSource(name) {
+		return
+	}
+
+	delete(instance.defaultOpsrcErrors, name)
```
galletti94: Why could we not end up in a situation where we are concurrently trying to update the same global variable? And one is trying to delete it why the other is trying to add an entry?

pkg/status/status.go
```diff
@@ -406,3 +427,15 @@ func ReportMigration(cfg *rest.Config, mgr manager.Manager, namespace string, ve
 	statusConditions := conditionListBuilder(configv1.OperatorDegraded, configv1.ConditionFalse, msg, "Upgrading")
 	return instance.setStatus(statusConditions)
 }
+
+// getFirstKeyValue returns the first key and value returned while iterating
+// over the given map[string] StatusError.
+// If the map is empty, "" and nil are returned.
+func getFirstKeyValue(m map[string]StatusError) (key string, err StatusError) {
```
galletti94: Nit: the name is a bit misleading because there isn't really a "first" key in a dictionary - or at least that can change depending on what you insert into it?

pkg/status/status.go
awgreene: Given the purpose of this code it is unlikely that we will encounter a race conditio. The controller handles OperatorSource events linearly - so I felt that adding a lock was over engineering. @ecordell can weigh in.

ecordell: Even if this function is never called concurrently, we still need to be concerned about concurrent reads from `monitorClusterStatus`, right?

pkg/status/status.go
```diff
@@ -365,27 +389,24 @@ func (s *status) monitorClusterStatus() {
...
-				conditionListBuilder(configv1.OperatorUpgradeable, configv1.ConditionTrue, upgradeableMessage, "OperatorAvailable")
-				if isSucceeding {
-					statusConditions = conditionListBuilder(configv1.OperatorDegraded, configv1.ConditionFalse, fmt.Sprintf("Current CR sync ratio (%g) meets the expected success ratio (%g)", *ratio, successRatio), "OperandTransitionsSucceeding")
+				reason := "OperatorAvailable"
+				conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionFalse, fmt.Sprintf("Successfully progressed to release version: %s", s.version), reason)
+				conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionTrue, fmt.Sprintf("Available release version: %s", s.version), reason)
+				conditionListBuilder(configv1.OperatorUpgradeable, configv1.ConditionTrue, upgradeableMessage, reason)
+				defaultOpsrcName, err := getFirstKeyValue(s.defaultOpsrcErrors)
+				if defaultOpsrcName != "" {
+					statusConditions = conditionListBuilder(configv1.OperatorDegraded, configv1.ConditionTrue, fmt.Sprintf("Default OperatorSource (%s) is failing: %v", defaultOpsrcName, err.Error()), err.Reason())
```
ecordell: Is there a reason we wouldn't want to see _all_ of the failing default operatorsources here?

pkg/operatorsource/succeeded.go
```diff
@@ -68,6 +69,9 @@ func (r *succeededReconciler) Reconcile(ctx context.Context, in *v1.OperatorSour
 
 	err = defaults.New(defaults.GetGlobalDefinitions(), operatorhub.GetSingleton().Get()).Ensure(r.client, in.Name)
 
+	// Report the OperatorSource success.
+	status.ReportOperatorSourceSuccess(out.Name)
```
ecordell: This reports that the OperatorSource is syncing properly, but not whether or not the catalogsource can successfully pull from an appregistry (i.e. we can be serving 0 packages and this won't report degraded).

Maybe this is fine? The BZ for this says:

> 1. Make marketplace unable to successfully reconcile one of the default OperatorSources.

can you tell me how to actually do that on a cluster? everything I think of involved disrupting the connection to quay or damaging the catalogsource pod in some way, which doesn't actually affect operatorsource reconciliation.

pkg/operatorsource/configuring.go
```diff
@@ -78,9 +79,13 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *v1.OperatorSo
 
 	r.logger.Infof("Downloading metadata from Namespace [%s] of [%s]", in.Spec.RegistryNamespace, in.Spec.Endpoint)
 
-	metadata, err := r.getManifestMetadata(&in.Spec, in.Namespace)
-	if err != nil {
+	metadata, statusErr := r.getManifestMetadata(&in.Spec, in.Namespace)
+	if statusErr != nil {
+		err = errors.New(statusErr.Error())
```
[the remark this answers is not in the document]
awgreene: Ah this is correct - good catch.

pkg/operatorsource/configuring.go
```diff
@@ -92,6 +97,10 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *v1.OperatorSo
 		// manifest(s) registry sync will detect a new release and will trigger
 		// a new reconciliation.
 		nextPhase = phase.GetNextWithMessage(phase.Failed, err.Error())
+
+		// Report the OperatorSource StatusError.
+		statusErr = status.NewStatusError("AppRegistryMetadataEmptyError", err)
```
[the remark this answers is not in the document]
awgreene: Sure thing!

pkg/status/status.go
awgreene: That's true @ecordell, I'll look into adding a lock.

pkg/status/status.go
awgreene: No - but we can only report a single condition that summaries why marketplace has degraded. Furthermore, the only information retrieved via telemetry includes the Name, Status, and Reason. Reason has to be a PascalCase string and I couldn't find a way to report multiple `opsrc` failures with such a requirement. Thoughts?

ecordell: Convinced in person

pkg/operatorhub/operatorhub.go
```diff
@@ -32,6 +32,7 @@ type operatorhub struct {
 type OperatorHub interface {
```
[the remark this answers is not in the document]
aravindhp: @tkashem @awgreene I finally remember the main reason why I stored the OperatorHub singleton in memory :-). This was to support defaults in upstream kube where the config object will not be present while not littering the code with checks to see if that API is available or not. This fact is not immediately obvious and I apologize for not adding a comment about this. I realize the kube way is to always go to etcd for cluster state, so if that is preferable I think you can just rewrite the `Get()` method to check for the API and then get the object. That way you can remove the in memory map, not have API checks all over the place and still support defaults in upstream kube. This is what I should have done in the first place. I am open to fixing this in the 4.3 time frame if no one on the team has cycles to do this.

PR #237, opened 2019-08-01
[proxy] Enable Proxy

awgreene:
This PR will update marketplace to work through the cluster proxy based on the latest developments.


manifests/07_operator.yaml
```diff
@@ -3,6 +3,8 @@ kind: Deployment
 metadata:
   name: marketplace-operator
   namespace: openshift-marketplace
+  annotations:
+    config.openshift.io/inject-proxy: "marketplace-operator"
```
ecordell: is the value arbitrary?

pkg/registry/registry.go
```diff
@@ -139,8 +139,7 @@ func (r *registry) ensureDeployment(appRegistries []string, needServiceAccount b
 				deployment.Spec.Template.Spec.ServiceAccountName = r.key.Name
 			}
 			// Update proxy env vars to match those in the operator.
-			deployment.Spec.Template.Spec.Containers[0].Env = proxy.GetInstance().GetEnvVars()
-
+			deployment.Spec.Template.Spec.Containers[0].Env = proxy.GetProxyEnvVars()
```
ecordell: Why not put the annotation on the registry deployments, too?

manifests/07_operator.yaml
awgreene: Based on [this line in the PR](https://github.com/openshift/cluster-version-operator/pull/224/files#diff-20dbece9affd4bfefb9de7caca6bc037R52) it looks like you need to provide the names of the pods.

pkg/registry/registry.go
awgreene: The CVO team does not want us adding this annotation to our operands.

awgreene:
This PR relies on OpenShift `proxy/cert` work to be done before it can go through.

/hold

test/helpers/helpers.go
```diff
@@ -542,3 +543,40 @@ func CheckSubscriptionNotUpdated(client test.FrameworkClient, name, namespace st
 	}
 	return nil
 }
+
+var isCoAPIPresent *bool
```
kevinrizza: Shouldn't this be set at the top of this file, maybe as part of a var () block?

awgreene: My mistake - I had meant to move this as you suggested.

awgreene:
Should be all set @kevinrizza

test/helpers/helpers.go
```diff
@@ -542,3 +548,38 @@ func CheckSubscriptionNotUpdated(client test.FrameworkClient, name, namespace st
...
 }
+
+// EnsureClusterOperatorIsAvailable will make a single attempt to add the ClusterOperator
+// to the FrameworkScheme. If an error is encountered in this first call, it will
+// be returned.
+// Subsequent calls will always return whether or not the ClusterOperator was added
+// to the FrameworkScheme and nil.
+// The boolean returned by this method can be used to identify if the tests are
+// being ran on an OpenShift cluster.
+func EnsureClusterOperatorIsAvailable() (bool, error) {
```
aravindhp: I suggest not doing things this way. This should be a function that is run once upon initialization and then you have another function that indicates availability. That was the pattern you were following before and I am not sure why you went away from it. Just my 2 bits.

awgreene: @kevinrizza and I had discussed this offline when he reviewed my code. IIRC, his argument was that we shouldn't impose ordering / dependencies on the order that we call functions. That by merging these two functions into a singular one, we guarantee that changes to the first function will not affect the behavior of the second function.

awgreene: I will bring this up with @kevinrizza when he returns from PTO so he may weigh in.

aravindhp: That is a fair argument but the way you are using the function is inconsistent and uses the author's knowledge about execution order. Otherwise you would always be checking for error in all places. By not doing so you are in fact introducing ordering indirectly.

awgreene: That is a fair point as well @aravindhp.

PR #238, opened 2019-08-06
Bug 1731338: OPSRC and CSC deployment name collision

dinhxuanvu:
Add a prefix to deployment's name to distinguish the deployments from
catalog source config and operator source.

Signed-off-by: Vu Dinh <vdinh@redhat.com>

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -38,7 +38,14 @@ var action = []string{"grpc_health_probe", "-addr=localhost:50051"}
 
 // DefaultServerImage is the registry image to be used in the absence of
 // the command line parameter.
-const DefaultServerImage = "quay.io/openshift/origin-operator-registry"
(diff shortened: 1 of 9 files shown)


pkg/registry/registry.go
```diff
@@ -6,8 +6,8 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
-	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v2"
+	v1 "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
```
kevinrizza: This is noise and unrelated to the rest of the pr.

kevinrizza:
Also the `e2e` test failure is valid because it checks for dependent resources. Since you renamed those, you need to ensure that the tests now use that prefix.

dinhxuanvu:
PR #254 replaces this PR.

PR #240, opened 2019-08-09
Bug 1739562: [migration] Don't kill due to invalid subscription

kevinrizza:
Currently, the migration code (to move objects from 4.1 -> 4.2 by
removing the catalogsourceconfig from the install path of OLM) iterates
over each subscription and attempts find a valid catalog source to
associate it with. Right now however, there is logic to abort starting
the operator if there are errors finding an association during this
process.

This seems problematic, given that it is entirely possible that any
given subscription can be generated by the user and may be invalid.
Instead we should do a best effort to attempt to create this association
and in the case that we cannot find a suitable one we simply log a
warning and continue as usual.

Additionally, this pr fixes an issue where, in the case where the
subscription points to a specific catalog source config (through labels)
and the package listed is invalid (or is just not set) that a null pointer
exception could occur and the marketplace would crash.

Associated bugzilla:
https://bugzilla.redhat.com/show_bug.cgi?id=1739562

### DIFF
--- a/test/testsuites/migrationtests.go
+++ b/test/testsuites/migrationtests.go
@@ -24,6 +24,7 @@ func MigrationTests(t *testing.T) {
 	t.Run("subscriptions-are-updated", testSubscriptionsUpdated)
 	t.Run("user-created-subscriptions-are-not-updated", testUserCreatedSubscriptions)
 	t.Run("catalogsourceconfigs-not-deleted", testCatalogSourceConfigsNotDeleted)
+	t.Run("bad-subscription-ignored", testCatalogSourceConfigsNotDeleted)
 }
 
 // testDatastoreAndInstalledCatalogSourceConfigsCleanedUp ensures that after a cluster is migrated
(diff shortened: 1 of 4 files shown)


test/testsuites/migrationtests.go
```diff
@@ -103,3 +104,17 @@ func testCatalogSourceConfigsNotDeleted(t *testing.T) {
...
+// any existing subscription that could not be migrated (because it points
+// to a non existent package, for example) is ignored
+func testInvalidSubscriptionIgnored(t *testing.T) {
+	ctx := test.NewTestCtx(t)
+	defer ctx.Cleanup()
+
+	namespace, err := ctx.GetNamespace()
+	require.NoError(t, err, "Could not get namespace")
+
+	err = test.Global.Client.Get(context.TODO(), types.NamespacedName{Name: helpers.TestInvalidSubscriptionName, Namespace: namespace}, &olm.Subscription{})
```
galletti94: Shouldn't you check that the catalogsource field is the same to make sure the subscription wasn't updated?

galletti94: https://github.com/operator-framework/operator-marketplace/pull/240/files/c92ce0484dbcf8bbbdd403118787908637ea134b#diff-25b86b22e24668a7c91ff67f35ffd5cdR87

PR #241, opened 2019-08-13
Bug 1740824: Fix degraded status on upgrade

awgreene:
Bug 1740824: Fix Degraded status on upgrade

Problem: There are three bugs that should be addressed within this bugzilla:

During an upgrade, multiple marketplace operators are running and updating the ClusterOperatorStatus at the same time, making it difficult to identify the actual state of the marketplace operator.
The marketplace operator is reporting degraded while it is upgrading. The marketplace operator reports degraded when the number of failed syncs surpasses some threshold of total syncs. The marketplace operator currently reports a failed sync whenever an error is encountered while reconciling an operand (OperatorSources or CatalogSourceConfigs). This allows network issues or invalid operands to drive the marketplace operator to a degraded state.
It is difficult to identify why a ClusterOperatorStatus condition is in a given state via telemetry. Marketplace currently does not set a reason when setting conditions and telemetry only includes the state and reason.
Solution: Three changes need to be made to address the bugzilla:

Implement leader election using the functions provided by the Operator-SDK to prevent multiple marketplace operators from updating the ClusterOperatorStatus at the same time.
Rather than reporting a failed sync whenever an error is encountered while reconciling an operand, report a failed sync when the marketplace operator is unable to get or update an existing operand.
Update marketplace to include a reason when setting a condition that indicates why the condition is being set.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"net/http"
 	"os"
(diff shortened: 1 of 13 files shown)


awgreene:
PTAL @shawn-hurley

PR #242, opened 2019-08-15
[CA] Use Certificate Authority bundle

awgreene:
This commit creates a controller that watches for events on the "marketplace-trusted-ca" ConfigMap. In the event that the ConfigMap is updated, the marketplace operator will restart, reconciling existing operands and recreate their deployments with the correct CA certificates.


awgreene:
Waiting on https://github.com/openshift/cluster-network-operator/pull/271/ before this can merge.

/hold

pkg/controller/configmap/configmap_controller.go
```diff
@@ -0,0 +1,85 @@
...
+type ReconcileConfigMap struct {
+}
+
+// Reconcile will restart the marketplace operator.
+func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	log.Infof("Certificate Authorization ConfigMap %s/%s has been updated, restarting marketplace.", request.Namespace, request.Name)
+
+	os.Exit(0)
+
+	// This code should never be reached.
```
galletti94: So why include it?

awgreene:
> Great work! Do we need to add an e2e for this?

@galletti94 Well if the proxy is configured to use a proxy the e2e tests should fail as all requests that don't go through the proxy should fail. If the proxy is not configured, the requests shouldn't go through a (non-existant) proxy.

pkg/controller/configmap/configmap_controller.go
awgreene: On second thought - this should be reached by vanilla kubernetes. Updating the comment.

pkg/controller/configmap/configmap_controller.go
```diff
@@ -0,0 +1,85 @@
...
+// True should only be returned when the ConfigMap is updated by the cert-injector-controller.
+func getPredicateFunctions() predicate.Funcs {
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// If the ConfigMap is ever changed we should kick off an event.
+			if e.MetaOld.GetName() == ca.TrustedCaConfigMapName {
+				log.Infof("Update")
```
awgreene: I'm going to remove this in an upcoming PR update after the ongoing e2e-tests finish.

pkg/controller/configmap/configmap_controller.go
```diff
@@ -0,0 +1,94 @@
...
+
+// Reconcile will restart the marketplace operator.
+func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	// Check if the ConfigMap is in the same namespace that marketplace is deployed in.
+	objectInOtherNamespace, err := shared.IsObjectInOtherNamespace(request.Namespace)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// If this is an IpenShift cluster we should restart Marketplace.
```
danehans: s/IpenShift/OpenShift/

manifests/08_operator.yaml
```diff
@@ -67,3 +67,14 @@ spec:
...
+            - name: marketplace-trusted-ca
+              mountPath: /etc/pki/ca-trust/extracted/pem/
+      volumes:
+        - name: marketplace-trusted-ca
+          configMap:
+            name: marketplace-trusted-ca
+            items:
+            # Require that this data is present in the ConfigMap before the operator is deployed.
+            - key: ca-bundle.crt
+              path: ca-bundle.crt
```
danehans: Same as [this](https://github.com/operator-framework/operator-marketplace/pull/242/files#r314846834) comment.

pkg/controller/configmap/configmap_controller.go
awgreene: Thanks for catching this!

manifests/08_operator.yaml
awgreene: Good to know - updated in the latest commit.

deploy/operator.yaml
```diff
@@ -67,3 +67,14 @@ spec:
...
+            - name: marketplace-trusted-ca
+              mountPath: /etc/pki/ca-trust/extracted/pem/
+      volumes:
+        - name: marketplace-trusted-ca
+          configMap:
+            name: marketplace-trusted-ca
+            items:
+            # Require that this data is present in the ConfigMap before the operator is deployed.
+            - key: ca-bundle.crt
+              path: ca-bundle.crt
```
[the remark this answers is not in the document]
awgreene: Thanks, addressed in the latest commit.

eparis:
This was reverted because of https://bugzilla.redhat.com/show_bug.cgi?id=1743699
we will need to redo this work, likely while making the configmap optional.

PR #244, opened 2019-08-21
Revert “[CA] Use Certificate Authority bundle"

awgreene:
Reverts operator-framework/operator-marketplace#242

The marketplace-operator relies on the cluster-network-operator to
inject a ConfigMap with the trusted Certificate Authorities bundle.
The marketplace-operator is upgraded at position 50 while the
cluster-network-operator is upgraded at position 70. Because of the
ordering, the marketplace-operator will not start due to the missing
CA bundle.

Due to the severity of the issue, marketplace is being reverted to https://github.com/operator-framework/operator-marketplace/commit/93e58ab4804fa74735ba85dcf69f3e53674ef96a


shawn-hurley:
/lgtm


Please add back with optional mount

eparis:
merged with all green CI to get the product back able to upgrade.

Thank you both so much for the extremely fast response!!!

wking:
Linking back to [rhbz#1743699][1].

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1743699

PR #245, opened 2019-08-21
[CA] Use Certificate Authority bundle

awgreene:
This commit creates a controller that watches for events on the
"marketplace-trusted-ca" ConfigMap. In the event that the ConfigMap is
updated, the marketplace operator will restart, reconciling existing
operands and recreate their deployments with the correct CA
certificates by mounting the ca-service.crt.

This commit is an updated version of commit
e71ce1e969e301507ca72a995e98fa81724cf1fd which specifies that the
marketplace-trusted-ca is optional. This will prevent upgrades from
failing.


awgreene:
The following command can be used to compare the current commit against the original implementation: `git diff e71ce1e969e301507ca72a995e98fa81724cf1fd`.

awgreene:
I confirmed that this PR never reports degraded despite the following `OperatorSources` failing to reach the succeeded state:
```
certified-operators   appregistry   https://quay.io/cnr   certified-operators   Certified Operators   Red Hat     Configuring   Get https://quay.io/cnr/api/v1/packages?namespace=certified-operators: proxyconnect tcp: x509: certificate sig
ned by unknown authority   173m
community-operators   appregistry   https://quay.io/cnr   community-operators   Community Operators   Red Hat     Configuring   Get https://quay.io/cnr/api/v1/packages?namespace=community-operators: proxyconnect tcp: x509: certificate sig
ned by unknown authority   173m
redhat-operators      appregistry   https://quay.io/cnr   redhat-operators	Red Hat Operators     Red Hat     Configuring   Get https://quay.io/cnr/api/v1/packages?namespace=redhat-operators: proxyconnect tcp: x509: certificate signed
 by unknown authority      173m
```

deploy/operator.yaml
```diff
@@ -67,3 +67,15 @@ spec:
...
+          volumeMounts:
+            - name: marketplace-trusted-ca
+              mountPath: /etc/pki/ca-trust/extracted/pem/
+      volumes:
+        - name: marketplace-trusted-ca
+          configMap:
+            optional: true
+            name: marketplace-trusted-ca
+            items:
+            # Require that this data is present in the ConfigMap before the operator is deployed.
```
awgreene: This comment is out of date and will be updated once @kevinrizza and @tkashem finish their review.

manifests/08_operator.yaml
```diff
@@ -67,3 +67,15 @@ spec:
...
+          volumeMounts:
+            - name: marketplace-trusted-ca
+              mountPath: /etc/pki/ca-trust/extracted/pem/
+      volumes:
+        - name: marketplace-trusted-ca
+          configMap:
+            optional: true
+            name: marketplace-trusted-ca
+            items:
+            # Require that this data is present in the ConfigMap before the operator is deployed.
```
awgreene: This comment is out of date and will be updated once @kevinrizza and @tkashem finish their review.

pkg/certificateauthority/caconfigmap.go
```diff
@@ -0,0 +1,64 @@
...
+			Name:      TrustedCaConfigMapName,
+			MountPath: trustedCaMountPath,
+		},
+	}
+}
+
+// getCaOnDisk returns the contents of the Certificate Authority bundle on disk as a byte
+// array or returns the error encountered when attempting to do so.
+func getCaOnDisk() ([]byte, error) {
+	b, err := ioutil.ReadFile(trustedCaMountPath + caBundlePath) // just pass the file name
```
awgreene: I will remove `// just pass the file name` once @kevinrizza and @tkashem finish their review.

pkg/certificateauthority/handler.go
```diff
@@ -0,0 +1,62 @@
...
+//
+// Handle handles a new event associated with OperatorSource type.
+//
+// ctx represents the parent context.
+// event encapsulates the event fired by operator sdk.
+type Handler interface {
+	Handle(ctx context.Context, operatorSource *corev1.ConfigMap) error
+}
+
+// configmapHandler implements the Handler interface
```
awgreene: I will change `configmapHandler` to `configmaphandler` once reviews are finished.

pkg/certificateauthority/handler.go
```diff
@@ -0,0 +1,62 @@
...
+type configmaphandler struct {
+	// This client, initialized using mgr.Client() above, is a split client
```
awgreene: This comment will be removed once the code is reviewed.

pkg/controller/configmap/configmap_controller.go
```diff
@@ -0,0 +1,130 @@
...
+	// that reads objects from the cache and writes to the apiserver
+	client  client.Client
+	handler ca.Handler
+}
+
+// Reconcile will restart the marketplace operator if the Certificate Authority ConfigMap is
+// not in sync with the Certificate Authority bundle on disk..
+func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	// Check if the CA ConfigMap is in the same namespace that Marketplace is deployed in.
+	objectInOtherNamespace, err := shared.IsObjectInOtherNamespace(request.Namespace)
```
kevinrizza: Can we rename this variable to something slightly different? I think using `object` is hard to read. Maybe reference the cert? `certConfigInOtherNamespace`?

awgreene: Sure - I had just followed the convention where this method was used elsewhere.

pkg/certificateauthority/handler.go
```diff
@@ -0,0 +1,62 @@
...
+	// Retrieve the Certificate Authority bundle from Disk.
+	// If an error is returned and is not related to a nonexistant file, return an error.
+	caOnDisk, err := getCaOnDisk()
+	if err != nil && !os.IsNotExist(err) {
+		log.Infof("[ca] Error reading from disk: %v", err)
+		return err
+	}
+
+	// Compare the Certificate Authority bundle on disk with the one in the ConfigMap.
+	if caBundle != string(caOnDisk) {
```
tkashem: nit: `bytes.Equal` should work too, in fact that's what go crypto lib does - https://golang.org/src/crypto/x509/x509.go?s=25381:25433#L741

pkg/controller/configmap/configmap_controller.go
```diff
@@ -0,0 +1,130 @@
...
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	err = r.handler.Handle(context.TODO(), caConfigMap)
```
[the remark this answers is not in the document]
awgreene: Can do!

pkg/certificateauthority/caconfigmap.go
```diff
@@ -56,9 +57,21 @@ func MountCaConfigMap(deployment *appsv1.Deployment) {
...
+	b, err := ioutil.ReadFile(path)
 	if err != nil {
 		return []byte{}, err
 	}
 	return b, nil
 }
+
+// isRunningOnPod returns true if marketplace is being ran on a pod.
+func isRunningOnPod() bool {
+	_, err := getFileOnDisk("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
```
tkashem: Nit: `getFileOnDisk` is responsible for reading the content of a file, using it to check the existence of a file seems to be a misuse. Maybe ` info, err := os.Stat(filename)` is more appropriate?

tkashem: Question: Is Checking the existence of `/var/run/secrets/kubernetes.io/serviceaccount/namespace` a convention the community uses to check whether the process is running within a pod?

pkg/certificateauthority/handler.go
```diff
@@ -0,0 +1,64 @@
...
+		if !isRunningOnPod() {
```
tkashem: Question:
- If we are not running inside a Pod, why do we even have this reconciler? We can simply check if we are running inside a pod when we create the controller and choose not to instantiate it, correct?

pkg/certificateauthority/caconfigmap.go
awgreene: Fair point - I'll make the switch.

pkg/certificateauthority/handler.go
awgreene: That's true, I'll move the logic there.

pkg/certificateauthority/caconfigmap.go
awgreene: I had gotten the suggestion from @joelanford, I have not seen a community convention.

pkg/certificateauthority/handler.go
awgreene: On second thought -  the `operator-sdk up local` is very useful for quick local development - the change that you're suggesting would make debugging CA problems require spinning up a pod, while the currently implementation provides a helpful message. Thoughts?

awgreene: This would be particularly useful if there is ever an error when reconciling the configmap.

kevinrizza: I think if you're actually working on this code you should be able to configure your environment to handle it. By default just disabling it seems like the right way to do it to me.

awgreene: Okay - I'll make the change.

tkashem: Good point! Does the configmap reconciler need to run on an upstream kubernetes distribution? If the answer is no then I think we should treat it as a configurable feature.

pkg/certificateauthority/caconfigmap.go
```diff
@@ -0,0 +1,59 @@
...
+			Name:      TrustedCaConfigMapName,
+			MountPath: trustedCaMountPath,
+		},
+	}
+}
+
+// getCaOnDisk returns the contents of the Certificate Authority bundle on disk as a byte
+// array or returns the error encountered when attempting to do so.
+func getCaOnDisk() ([]byte, error) {
+	return ioutil.ReadFile(trustedCaMountPath + caBundlePath)
```
ecordell: should this use a `filepath.Join`?

awgreene: After reviewing the Go Doc, this seems like a good idea. The latest commit uses `filepath.Join`.

PR #248, opened 2019-09-06
Fix broken links - OLM Documentation was replaced by doc

camilamacedo86:
The Dir of docs in OLM was changed from Documentation to doc. This change is to fix the broken links. 

- https://github.com/operator-framework/operator-sdk/issues/1887

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,7 +38,7 @@ $ oc apply -f deploy/operator.yaml
 
 ## Marketplace-operator on vanilla Kubernetes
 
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
+# Intellij IDEA
+.idea
```
kevinrizza: This is noise, please remove this from the pull request.

camilamacedo86: It is for do not commit ( ignore ) the files from Golang IDEA as it is done for the other things there. If the dev user IDEA files from IDEA that should not have been trying to be committed.

kevinrizza: I do understand what the change to the `.gitignore` file does. It is just unrelated to the commit message, pr header, and purpose of the pull request. IMO a pull request to this repo should only include changes directly related to that pull request. If you would like to suggest an update to the `.gitignore` please open a separate pull request for that purpose.

camilamacedo86: Ok I catcher your point. Tks for the clarifications.

camilamacedo86: Done ..

PR #249, opened 2019-09-09
Bug 1749643: don't crash on non-hyphenated csc

gallettilance:
Problem:
non-hyphenated CatalogSourceConfigs cause the marketplace operator pod
to crash during migration because the function that infers the
CatalogSource name from the CatalogSourceConfig assumes the presence of
a hyphen.

Solution:
Check that splitting a string on the hyphen character produces a list of
length greater than 1.


pkg/migrator/migrator.go
```diff
@@ -252,8 +252,11 @@ func IsPackageInOpsrc(packageName string, opsrc *v1.OperatorSource) bool {
 // extracts `community` off the name, appends `operators`
 // to it, and returns `community-operators` as output
 func ExtractCsName(cscName string) string {
-	possibleCsName := strings.Split(cscName, "-")[1]
-	return fmt.Sprintf("%s-%s", possibleCsName, "operators")
+	possibleCsName := strings.Split(cscName, "-")
+	if len(possibleCsName) > 1 {
+		return fmt.Sprintf("%s-%s", possibleCsName[1], "operators")
+	}
+	return return fmt.Sprintf("%s-%s", possibleCsName[0], "operators")
```
awgreene: can you explain the `return return`?

kevinrizza: does this work in every case? should we have some more fundamental base case here instead?

gallettilance: Maybe we can open a separate PR to improve the implementation of this function?

gallettilance:
@kevinrizza @awgreene this is ready for another review

awgreene:
This looks good to me.

PR #250, opened 2019-09-10
[deploy] Sync deploy and manifest dirs

awgreene:
The ServiceMonitor yaml defined in the deploy dir did not contain
required  RBAC permissions for the prometheus-k8s ServiceAccount.
This commit introduces the missing RBAC permissions.

### DIFF
--- a/deploy/service_monitor.yaml
+++ b/deploy/service_monitor.yaml
@@ -14,3 +14,34 @@ spec:
   selector:
     matchLabels:
       name: marketplace-operator
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: openshift-marketplace-metrics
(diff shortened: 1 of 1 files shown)


deploy/service_monitor.yaml
```diff
@@ -14,3 +14,34 @@ spec:
   selector:
     matchLabels:
       name: marketplace-operator
+---
```
kevinrizza: If these are needed for the cr, why aren't we applying them first in separate files?

awgreene: I was following the [example set in OLM](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/deploy/ocp/manifests/0.11.0/0000_90_olm_00-service-monitor.yaml#L63-L93).

PR #251, opened 2019-09-10
Race condition in config test

kevinrizza:
A race condition can occur in the `testDisableNonDefault()` test in
the `OperatorHubTests` test suite e2e tests where the test updates the
config object then immediately does a get without waiting for the config
object's controller to handle the update event. In some cases, this can
cause a test failure because the get returns the old object that is yet
to be updated.

This commit updates the test to poll against the object until the config
object has a chance to be updated before continuing with the test.

### DIFF
--- a/test/testsuites/operatorhubtests.go
+++ b/test/testsuites/operatorhubtests.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"testing"
+	"time"
 
 	apiconfigv1 "github.com/openshift/api/config/v1"
 	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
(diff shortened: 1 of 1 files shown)


test/testsuites/operatorhubtests.go
```diff
@@ -178,10 +180,18 @@ func testDisableNonDefault(t *testing.T) {
 
 	_ = updateOperatorHub(t, sources, false)
 
-	// Check if the cluster OperatorHub resource values are as expected
-	cluster := getClusterOperatorHub(t)
-	assert.True(t, len(cluster.Status.Sources) == len(helpers.DefaultSources)+1,
-		"Spurious elements in HubSourceStatus")
+	// Wait for the operatorhub update to complete
+	cluster := &apiconfigv1.OperatorHub{}
+	err = wait.Poll(time.Second*5, time.Minute*1, func() (done bool, err error) {
```
awgreene: Perhaps this should be `PollImmediate`?

test/testsuites/operatorhubtests.go
```diff
@@ -178,10 +180,18 @@ func testDisableNonDefault(t *testing.T) {
...
+		err = test.Global.Client.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, cluster)
+		if err != nil {
+			return false, err
```
awgreene: If this `Get` returns an error, the poll will exit. Is that expected?

PR #252, opened 2019-09-10
Bug 1750654: The phase message of csc won't change when the env changed

awgreene:
Problem:
If a CatalogSourceConfig specifies a source that does not exist the
CatalogSourceConfig will get stuck in the configuring phase. This error
is encountered because the Reconcile method for the CatalogSourceConfig
is not returning an error, causing the handler not to reconcile the
CatalogSourceConfig.

Solution:
Update the CatalogSourceConfig Configuring Reconcile method to return an
error. Also adds e2e tests to prevent regressions in the future.

### DIFF
--- a/test/testsuites/packagestests.go
+++ b/test/testsuites/packagestests.go
@@ -22,7 +22,7 @@ func PackageTests(t *testing.T) {
 
 	t.Run("resolve-missing-source", resolveMissingSource)
 	t.Run("unresolved-missing-source", unresolvedMissingSource)
-	t.Run("non-existing-source", nonExistingSource)
+	t.Run("resolve-non-existing-source", resolveNonExistingSource)
 	t.Run("source-missing-packages", sourceMissingPackages)
 }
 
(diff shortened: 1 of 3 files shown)


test/testsuites/packagestests.go
```diff
@@ -197,6 +213,24 @@ func sourceMissingPackages(t *testing.T) {
 }
 
 func runSourceTest(namespace, source, packages, expectedPhase, expectedMessage string) error {
+	// Create the CatalogSourceConfig.
+	csc, err := createCscRuntimeObject(namespace, source, packages)
+	if err != nil {
+		return err
+	}
+
+	// Check that the CatalogSourceConfig with an non-existing targetNamespace eventually reaches the
```
gallettilance: typo: `a non-existent`

awgreene: Addressed in the latest commit.

test/testsuites/packagestests.go
```diff
@@ -165,19 +165,35 @@ func unresolvedMissingSource(t *testing.T) {
 }
 
 // nonExistingSource ensures that if the provided source does not exist the CatalogSourceConfig
-// will be placed in the Configuring phase with the expected messages.
-func nonExistingSource(t *testing.T) {
+// will be placed in the Configuring phase with the expected messages. Then, the non-existant
```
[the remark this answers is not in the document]
awgreene: Addressed in the latest commit.

awgreene:
[INFO] Acquiring a lease ...
failed to acquire a resource: Post http://boskos.ci/acquire?dest=leased&owner=ci-op-6yvtq2vc-bc8e3&request_id=2942913621423428907&state=free&type=aws-quota-slice: dial tcp 172.30.131.17:80: connect: no route to host

/test e2e-aws-console-olm

pkg/certificateauthority/handler.go
```diff
@@ -42,7 +42,7 @@ func (h *configmapHandler) Handle(ctx context.Context, in *corev1.ConfigMap) err
 	caBundle := in.Data[caBundleKey]
 
 	// Retrieve the Certificate Authority bundle from Disk.
-	// If an error is returned and is not related to a nonexistant file, return an error.
+	// If an error is returned and is not related to a non-existent file, return an error.
```
kevinrizza: nit: we shouldn't include this change in this pr since it seems unrelated.

awgreene: Addressed.

PR #254, opened 2019-09-11
Bug 1731338: OPSRC and CSC child resources name conflict

dinhxuanvu:
Currently, if opsrc and csc with the same name are created, both
are created with Succeeded status despite the latter one doesn't
have a deployment for registry pod as marketplace reclaims the
existitng deployment with the same name that belongs to opsrc/csc
that is created first.

By adding ownership validation via labels, child resource such as
deployment cannot be reclaimed unless it is owned by a correct
opsrc/csc. As a result, whichever opsrc/csc is created latter will
fail to reach Succeeded state as it is unable to create child resouces
due to name conflicts nor reclaim those resources as it doesn't own
those resources.

Signed-off-by: Vu Dinh <vdinh@redhat.com>

### DIFF
--- a/pkg/grpccatalog/grpccatalog.go
+++ b/pkg/grpccatalog/grpccatalog.go
@@ -70,6 +70,11 @@ func (r *GrpcCatalog) EnsureResources(key types.NamespacedName, displayName, pub
 
 	// Update the CatalogSource if it exists else create one.
 	if err == nil {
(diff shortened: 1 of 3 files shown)


kevinrizza:
So this pr will effectively just cause the reconciliation to fail on a `catalogsourceconfig` when the name matches an `operatorsource`? Can we call out that purpose explicitly in the commit message?

PR #256, opened 2019-09-30
Bug 1752063: increase initial delay for appregistry pods

ecordell:
This increases the initial delay for the readiness and liveness probes for the appregistry pods. 

As more entries are added to appregistry, it takes longer to download them on startup. Since we are moving away from appregistry, we're mitigating the issue by increasing the startup time, rather than coming up with solutions that report ready faster.

Before merging, I would like to take a look at the ci cluster and ensure that the pods are not emitting events that signal an error for readiness probes.

**Reviewer Checklist**

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -397,8 +397,8 @@ func (r *registry) newPodTemplateSpec(registryCommand []string, needServiceAccou
 								Command: action,
 							},
 						},
-						InitialDelaySeconds: 5,
-						FailureThreshold:    30,
+						InitialDelaySeconds: 30,
+						FailureThreshold:    10,
 					},
 					LivenessProbe: &core.Probe{
 						Handler: core.Handler{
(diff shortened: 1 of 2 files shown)


cuppett:
@kevinrizza You might need to approve this one.

kevinrizza:
@cuppett It looks like Evan didn't want to merge this yet:

```
Before merging, I would like to take a look at the ci cluster and ensure that the pods are not emitting events that signal an error for readiness probes.
```

which is why I haven't `lgtm`ed yet. I also think these numbers are a guess and Evan also wanted to do some performance testing to determine if they are reasonable. But if we feel like this is an actual issue that needs to be addressed asap I can merge it for now and we can open up a follow up?

njhale:
Looks like there are error events being thrown, but are timeouts:
```
Readiness probe failed: timeout: failed to connect service "localhost:50051" within 1s
```

ecordell:
These values look good on a real cluster. The appregistry pods tend to come up in about 7 seconds right now, so this gives a good buffer.

njhale:
/hold

The failure is pretty consistent. Holding off on retesting for now.

PR #257, opened 2019-10-15
Bug 1757099: Assume available status is true without defaults

kevinrizza:
Problem:
The clusteroperator operator status is never set to Available=True when
there are no default operator sources set. This is because it relies on
a set of reconciliations to return successfully and determines a ratio
based on that value. When no default operator sources are set, there are
no reconciliation events taking place and therefore the ratio is always
set to zero.

Solution:
When no default operator sources are set and no events are firing, set
the Available status to true.

https://bugzilla.redhat.com/show_bug.cgi?id=1757099

### DIFF
--- a/pkg/operatorhub/operatorhub.go
+++ b/pkg/operatorhub/operatorhub.go
@@ -32,6 +32,7 @@ type operatorhub struct {
 type OperatorHub interface {
 	Get() map[string]bool
 	Set(spec configv1.OperatorHubSpec)
+	Disabled() bool
 }
 
 // GetSingleton returns the singleton instance of HubConfig
@@ -44,6 +45,21 @@ func (o *operatorhub) Get() map[string]bool {
 	o.lock.Lock()
 	defer o.lock.Unlock()
 	return o.current
+}
+
+// Disabled returns true if all defaults are disabled
+func (o *operatorhub) Disabled() bool {
+	o.lock.Lock()
+	defer o.lock.Unlock()
+
+	disabled := true
+	for _, disabled := range o.current {
+		if disabled == false {
+			disabled = false
+		}
+	}
+
+	return disabled
 }
 
 // Set sets the current configuration based on the spec. If the spec is empty,
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -13,6 +13,7 @@ import (
 	mktconfig "github.com/operator-framework/operator-marketplace/pkg/apis/config/v1"
 	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
 	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v2"
+	"github.com/operator-framework/operator-marketplace/pkg/operatorhub"
 	log "github.com/sirupsen/logrus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -26,7 +27,7 @@ const (
 
 	// minSyncsBeforeReporting is the minimum number of syncs we wish to see
 	// before reporting that the operator is available
-	minSyncsBeforeReporting = 4
+	minSyncsBeforeReporting = 3
 
 	// successRatio is the ratio of successful syncs / total syncs that we
 	// want to see in order to report that the marketplace operator is not degraded.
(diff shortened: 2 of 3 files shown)


pkg/status/status.go
```diff
@@ -358,8 +359,21 @@ func (s *status) monitorClusterStatus() {
 				break
 			}
 
-			// Wait until the operator has reconciled the minimum number of syncs.
 			_, syncEvents := s.syncRatio.GetSyncs()
+
+			// no default operator sources are present, so assume we are in a good state
+			if len(defaults.GetGlobalDefinitions()) == 0 && syncEvents == 0 {
```
ecordell: `GetGlobalDefaults` returns the defaults built into the container, not the defaults present on the cluster. I don't see how this will fix the bug that occurs when `disableAllDefaultSources` is set.

pkg/status/status.go
```diff
@@ -358,8 +359,21 @@ func (s *status) monitorClusterStatus() {
...
+				log.Info("No default operators. Let's assume we are succeeding!")
+				reason := "NoDefaultOpSrcEnabled"
+				conditionListBuilder := clusterStatusListBuilder()
+				conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionFalse, fmt.Sprintf("Successfully progressed to release version: %s", s.version), reason)
+				conditionListBuilder(configv1.OperatorUpgradeable, configv1.ConditionTrue, upgradeableMessage, reason)
+				statusConditions := conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionTrue, fmt.Sprintf("Available release version: %s", s.version), reason)
+				statusErr = s.setStatus(statusConditions)
+				break
```
ecordell: Do we have tests for this?

pkg/status/status.go
kevinrizza: Yep, you were correct. I misunderstood what the `GetGlobalDefaults` function actually did. please take another look

pkg/status/status.go
kevinrizza: @ecordell added tests.

pkg/operatorhub/operatorhub.go
```diff
@@ -46,6 +47,21 @@ func (o *operatorhub) Get() map[string]bool {
 	return o.current
 }
 
+// Disabled returns true if all defaults are disabled
+func (o *operatorhub) Disabled() bool {
+	o.lock.Lock()
+	defer o.lock.Unlock()
+
+	disabled := true
```
ecordell: This function always returns true, I think you want:
```go 
o.lock.Lock()
defer o.lock.Unlock()

for _, disabled := range o.current {
  if disabled == true {
    return false
  }
}
return true
```

(there should probably be a test for this as well)

test/testsuites/operatorhubtests.go
```diff
@@ -213,6 +214,62 @@ func testDisableNonDefault(t *testing.T) {
 	resetClusterOperatorHub(t, namespace)
 }
 
+// testClusterStatusDefaultsDisabled tests that, when all default operator sources are disabled,
+// the clusterstatus sets Available=True
+func testClusterStatusDefaultsDisabled(t *testing.T) {
```
ecordell: Can you make sure there's a test that covers the case where they're not all disabled? I would've assumed there was one already, but if so it should've caught the issue with `func (o *operatorhub) Disabled()`

pkg/operatorhub/operatorhub.go
kevinrizza: sorry this isn't clear because i named two variables the same thing -- this is intending to check that *all* of the values in that map are set to false.

That's because our choice is either:

1. Some (or all) of the defaults are enabled. In that case, just do the normal behavior. Something will reconcile and we should ensure it's working and report when it is not.
2. NONE of the defaults are enabled, in which case just say that we are available.

So to make this more readable it should look something like:

``` go
o.lock.Lock()
defer o.lock.Unlock()

allDisabled = true
for _, disabled := range o.current {
  if disabled == false {
    allDisabled = false
  }
}

return allDisabled
```

test/testsuites/operatorhubtests.go
kevinrizza: The test where they are not all disabled should be the same as the test where the operatorhub isn't set. We can definitely add a test explicitly checking for that though.

kevinrizza:
closing in favor of https://github.com/operator-framework/operator-marketplace/pull/260

PR #260, opened 2019-10-28
Bug 1757099: Assume available status is true without defaults

Bowenislandsong:
Problem:
The cluster operator status is never set to Available=True when
there are no default operator sources set. This is because it relies on
a set of reconciliations to return successfully and determines a ratio
based on that value. When no default operator sources are set, there are
no reconciliation events taking place and therefore the ratio is always
set to zero.

Solution:
When no default operator sources are set and no events are firing, set
the Available status to true.

https://bugzilla.redhat.com/show_bug.cgi?id=1757099

/cc @ecordell @kevinrizza

### DIFF
--- a/pkg/operatorhub/operatorhub.go
+++ b/pkg/operatorhub/operatorhub.go
@@ -32,6 +32,7 @@ type operatorhub struct {
 type OperatorHub interface {
 	Get() map[string]bool
 	Set(spec configv1.OperatorHubSpec)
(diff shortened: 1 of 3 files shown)


kevinrizza:
lgtm

Bowenislandsong:
@kevinrizza Please see the second commit. I have to reset disabled operators for both tests. Thank you.

PR #271, opened 2019-11-21
Bug 1775245: Schedule registry pods onto linux nodes only

Denton24646:
Adds NodeSelector to all registry pods that ensures they will get scheduled to linux nodes only.

Bugs reporting different registry pods being scheduled to windows nodes in mixed clusters and failing to start

**Reviewer Checklist**

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -32,6 +32,8 @@ const (
 	portNumber                 = 50051
 	portName                   = "grpc"
 	deploymentUpdateAnnotation = "openshift-marketplace-update-hash"
+	linuxNodeSelectorKey       = "beta.kubernetes.io/os"
+	linuxNodeSelectorValue     = "linux"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/registry/registry.go
```diff
@@ -415,6 +415,9 @@ func (r *registry) newPodTemplateSpec(registryCommand []string, needServiceAccou
 					},
 				},
 			},
+			NodeSelector: map[string]string{
+				"beta.kubernetes.io/os": "linux",
```
kevinrizza: nit: should these be variables so that the test can just import them?

kevinrizza:
Also I think you need to remove the parens from the pr header otherwise the automation isn't smart enough to pick it up.

pkg/registry/registry.go
Denton24646: That's a good point - these should be `consts` that are exported

PR #277, opened 2020-01-09
Use DNS for CatalogSource address

kevinrizza:
Problem:
Marketplace needs to support ipv6 clusters. Currently when creating
catalogsources, the marketplace points the catsrc to the clusterIP
address of the registry pod that it spins up. This cluster ip will not
work in an ipv6 only cluster.

Solution:
Use the dns address of the service to wire up the catalogsource to the
registry pod.

**Reviewer Checklist**

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -272,7 +272,7 @@ func (r *registry) ensureService() error {
 	}
 	r.log.Infof("Created Service %s", service.GetName())
 
-	r.address = service.Spec.ClusterIP + ":" + strconv.Itoa(int(service.Spec.Ports[0].Port))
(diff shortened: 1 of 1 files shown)


pkg/grpccatalog/grpccatalog.go
```diff
@@ -108,6 +112,19 @@ func (r *GrpcCatalog) EnsureResources(key types.NamespacedName, displayName, pub
 	return nil
 }
 
+func setTimestampLabel(catsrc *olm.CatalogSource) {
```
ecordell: not 100% sure this is necessary, since we watch events on the GRPC connection and resync the catalogsource if the connection state changes. 

The grpc connection can't live through a pod rollout but this seems like a reasonable safeguard in case some events are missed / issues with the grpc conection watch.

PR #278, opened 2020-01-14
Add marketplace default catalog sources

jpeeler:
OLM-1441

This adds two new appregistry based operator sources.

This allows the open marketplace catalog built by the container pipeline to be a default catalog.

**Reviewer Checklist**

### DIFF
--- a/defaults/04_redhat_marketplace.yaml
+++ b/defaults/04_redhat_marketplace.yaml
@@ -1,0 +1,13 @@
+apiVersion: "operators.coreos.com/v1"
+kind: "OperatorSource"
+metadata:
+  name: "redhat-marketplace"
+  namespace: "openshift-marketplace"
+  labels:
+    opsrc-provider: marketplace
+spec:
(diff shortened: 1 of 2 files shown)


shawn-hurley:
Don't we want to add this to the config object as a default? 

Also, do we really want to make the stage registry a default or am I misunderstanding that?

defaults/04_redhat_marketplace.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: "operators.coreos.com/v1"
+kind: "OperatorSource"
+metadata:
+  name: "redhat-marketplace"
+  namespace: "openshift-marketplace"
+  labels:
+    opsrc-provider: redhat
```
shawn-hurley: I think that this needs to be `marketplace` or something similar so that we can tell it is different than the default RedHat catalog

ecordell:
looks like the test has an error: index out of range

PR #280, opened 2020-01-30
Bug 1787660: Add CLI flag to set ClusterOperator name.

benluddy:
If the flag is set to a non-empty string, and config.openshift.io/v1
isn't available, then the operator exits with an error
status. Previously, the absence of configv1 would log a warning and
continue running without ever updating ClusterOperator status. If this
occurred during an OpenShift cluster upgrade, the upgrade would time
out waiting for marketplace to report success via its ClusterOperator.

Outside of an OpenShift cluster, the new flag can be omitted (or set
to the empty string) to disable ClusterOperator status updates
altogether.


pkg/operatorsource/handler.go
```diff
@@ -4,7 +4,7 @@ import (
 	"context"
 
 	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/shared"
-	"github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	v1 "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
```
kevinrizza: What's the point of this alias?

benluddy: That is weird. I have an after-save hook to tell `gopls` to organize imports, but I didn't notice that it had done that.

PR #282, opened 2020-02-26
Bug 1801294: Alert to make custom opsrc/csc users aware of deprecation

anik120:
In 4.5 we are deprecating the OperatorSource and CatalogSourceConfig APIs.
This PR introduces an alert in the logs, and also blocks upgrade to future
versions in presence of Opsrcs/CSCs by sending a Upgradeable: False message
to CVO.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/catalogsourceconfig/catalogsourceconfig_controller.go
+++ b/pkg/controller/catalogsourceconfig/catalogsourceconfig_controller.go
@@ -91,6 +91,7 @@ type ReconcileCatalogSourceConfig struct {
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileCatalogSourceConfig) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	log.Printf("Reconciling CatalogSourceConfig %s/%s\n", request.Namespace, request.Name)
+	log.Warning("DEPRECATION NOTICE: The CatalogSourceConfig API is deprecated in future versions. Please visit this link for futher details: https://docs.openshift.com/container-platform/4.4/release_notes/ocp-4-4-release-notes.html#ocp-4-4-marketplace-apis-deprecated")
 	// Reconcile kicked off, message Sync Channel
 	r.syncSender.SendSyncMessage(nil)
 
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -1,6 +1,7 @@
 package status
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"sync"
@@ -14,11 +15,13 @@ import (
 	mktconfig "github.com/operator-framework/operator-marketplace/pkg/apis/config/v1"
 	v1 "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
 	v2 "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v2"
+	"github.com/operator-framework/operator-marketplace/pkg/defaults"
 	"github.com/operator-framework/operator-marketplace/pkg/operatorhub"
 	log "github.com/sirupsen/logrus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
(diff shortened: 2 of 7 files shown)


pkg/status/status.go
```diff
@@ -417,3 +449,33 @@ func (NoOpReporter) StartReporting() <-chan struct{} {
 func (NoOpReporter) ReportMigration() error {
 	return nil
 }
+
+// checkOperatorUpgradeablity checks for the presence of CatalogSourceConfigs,
+// and OperatorSources in the cluster. Since both the APIs are scheduled to be
+// deprecated/are already deprecated, checkOperatorUpgradeablity returns false
+// if it detects the presence of any CSC or Opsrc in the cluster to indicate
+// that the cluster is not upgradable to future versions.
+func checkOperatorUpgradeablity() (bool, error) {
```
kevinrizza: Shouldn't we be ignoring the default operator sources though? Those will be migrated to catalog sources automatically, so we don't want to block upgrade for those

pkg/controller/catalogsourceconfig/catalogsourceconfig_controller.go
```diff
@@ -91,6 +91,7 @@ type ReconcileCatalogSourceConfig struct {
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileCatalogSourceConfig) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	log.Printf("Reconciling CatalogSourceConfig %s/%s\n", request.Namespace, request.Name)
+	log.Warning("DEPRECATION NOTICE: The CatalogSourceConfig API is deprecated in future versions. Please visit this link for futher details: https://docs.openshift.com/container-platform/4.5/release_notes/ocp-4-5-release-notes.html#ocp-4-5-marketplace-apis-deprecated")
```
kevinrizza: Same comment as the other, I think we don't want to throw warnings for defaults since we will migrate them cleanly without admin intervention

anik120: @kevinrizza do you mean we don't want to throw warnings for default Opsrcs? My understanding was that we want warnings for ALL CSCs.

pkg/status/status.go
anik120: @kevinrizza that is right. Updated to ignore default operator sources.

kevinrizza:
/approve

@shawn-hurley Can you put super quick eyes on this? It all seems reasonable to me but I would like to double check that we're 100% good with preventing upgrade due to these APIs.

pkg/controller/catalogsourceconfig/catalogsourceconfig_controller.go
shawn-hurley: @adellape can you verify that this would be the correct header?

adellape: Discussed with Anik/Kevin and decided it should be a 4.4 target: https://docs.openshift.com/container-platform/4.4/release_notes/ocp-4-4-release-notes.html#ocp-4-4-marketplace-apis-deprecated

anik120: @shawn-hurley updated the PR with the 4.4 release note target.

piyush-nimbalkar:
@anik120  If the OperatorSource API is getting deprecated, what are the users supposed to do to migrate? What is the alternative?
Is there is any documentation on the why and how of this? Could not access the link in code as it is not public.

anik120:
> @anik120 If the OperatorSource API is getting deprecated, what are the users supposed to do to migrate? What is the alternative?
Is there is any documentation on the why and how of this? Could not access the link in code as it is not public.

@piyush-nimbalkar the doc is being created right now. It is not live yet, which is why you couldn't access the doc. It'll be live with the 4.4 release, and will contain all the information you need on how to migrate your custom OperatorSources over.

piyush-nimbalkar:
Thank you @anik120 
In the meanwhile, should we not use OperatorSource? Asking this because I was starting something new using OperatorSource. If that is getting deprecated, would like to know what should I be using.

anik120:
> Thank you @anik120
> In the meanwhile, should we not use OperatorSource? Asking this because I was starting something new using OperatorSource. If that is getting deprecated, would like to know what should I be using.
@piyush-nimbalkar what exactly do you mean by `starting something new with using OperatorSource`? 

If you're just looking to build a catalog of operators to make them available in your cluster you can refer to this doc : https://github.com/operator-framework/operator-registry/tree/release-4.1#building-a-catalog-of-operators-using-operator-registry
The docker file this doc refers to: https://github.com/operator-framework/operator-registry/blob/release-4.1/upstream-example.Dockerfile

PR #286, opened 2020-03-05
Bug 1809200: enable prometheus metrics over https

exdx:
Enable the prometheus server that runs alongside marketplace to report metrics over https by using provided certificate and keys from the marketplace-trusted-ca volume that is auto-mounted in the marketplace operator deployment.  

Metrics were provided only via http previously which is insecure. See BZ for more details. 

**Reviewer Checklist**


exdx:
Conceptually, I think this is what we want, but I'm a little confused by the volume that is provided by the marketplace-trusted-ca. I see in https://github.com/operator-framework/operator-marketplace/blob/master/manifests/09_operator.yaml#L85 that we get a .pem file that we can use. 

However to enable https the server needs both the cert and the private key - do we have access to the key in the .pem file? Is there a way to unpack the .pem file to get both the cert and the key? Need a little clarity here.

cmd/manager/main.go
```diff
@@ -63,6 +68,7 @@ func main() {
 		"", "the directory where the default OperatorSources are stored")
 	var clusterOperatorName string
 	flag.StringVar(&clusterOperatorName, "clusterOperatorName", "", "the name of the OpenShift ClusterOperator that should reflect this operator's status, or the empty string to disable ClusterOperator updates")
+
```
awgreene: Do we need this newline?

cmd/manager/main.go
```diff
@@ -45,7 +46,11 @@ const (
 	updateNotificationSendWait = time.Duration(10) * time.Minute
 )
 
-var version = flag.Bool("version", false, "displays marketplace source commit info.")
+var (
+	version     = flag.Bool("version", false, "displays marketplace source commit info.")
+	tlsKeyPath  = flag.String("tls-key", "/etc/pki/ca-trust/extracted/pem/ca-bundle.crt", "Path to use for private key (requires tls-cert)")
+	tlsCertPath = flag.String("tls-cert", "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem", "Path to use for certificate key (requires tls-key)")
```
awgreene: Please use the constants defined in the `certificateauthority` package for the defaults.

pkg/metrics/metrics.go
```diff
@@ -42,8 +42,8 @@ var (
 	)
 )
 
-// init enables marketplace to serve prometheus metrics.
-func init() {
+// SetupPrometheus enables marketplace to serve prometheus metrics.
```
awgreene: Could we call this ServePrometheusMetrics?

pkg/metrics/metrics.go
```diff
@@ -42,8 +42,8 @@ var (
...
+func SetupPrometheus(useTLS bool, cert, key string) {
```
awgreene: We may want to update this function to return the error.

exdx:
Addressed feedback from here and slack. I think my first comment still stands though - will this actually work with the cert and key defaults as they are now? The "key" is referring to the actual configmap key to the .pem, not the actual private key. Unless they are both supplied in that volume.

exdx:
Evan provided helpful feedback in slack. 
Relevant OpenShift doc https://docs.openshift.com/container-platform/4.3/authentication/certificates/service-serving-certificate.html

exdx:
https://docs.openshift.com/container-platform/4.3/authentication/certificates/service-serving-certificate.html

exdx:
Updated per evan's comments in slack. I am using the service-ca controller to generate a cert and key via the annotation on the service. I have updated the operator deployment to mount the generated secret into the container. I have also added the location of the cert and key as runtime arguments to marketplace (as in OLM) and removed them as default values of the flags themselves (since this would break upstream users). 

I have also updated the prometheus service monitor to use https, but I'm not entirely sure that its correct? specifically in the `tlsConfig` the `caFile` is a reference to the location of the .crt file in the volume? I'm not entirely sure how the ServiceMonitor has access to this. Besides this open question I think this is just about ready.

PR #294, opened 2020-03-20
Updating the links to crd files

amalkurup89:

The links to two of the CRD files were deprecated. So, this change is to update those links to the correct files

Correcting the documentation, so as to ease the reference for future readers or users


benluddy:
/ok-to-test

Thanks!

PR #296, opened 2020-03-24
[release-4.3] Bug 1813995: support serving marketplace prometheus metrics over https

exdx:

**Reviewer Checklist**


njhale:
@exdx It looks like there's an issue w/ the options being passed into the marketplace-operator:

```
flag provided but not defined: -clusterOperatorName
Usage of marketplace-operator:
  -defaultsDir string
    	the directory where the default OperatorSources are stored
  -kubeconfig string
    	Paths to a kubeconfig. Only required if out-of-cluster.
  -master string
    	The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.
  -registryServerImage string
    	the image to use for creating the operator registry pod (default "quay.io/openshift/origin-operator-registry")
  -tls-cert string
    	Path to use for certificate (requires tls-key)
  -tls-key string
    	Path to use for private key (requires tls-cert)
  -version
    	displays marketplace source commit info.
```

exdx:
Strange, the test that fails is testing a 4.1.x -> 4.2 upgrade. Not related to the BZ. 
```
// testDatastoreAndInstalledCatalogSourceConfigsCleanedUp ensures that after a cluster is migrated
// from openshift 4.1.x to openshift 4.2.0, the following stale objects are cleaned up:
// Stale CatalogSourceConfigs created during operator installation in a openshift 4.1.x cluster.
// Datastore CatalogSourceConfigs created by OperatorSources in a openshift 4.1.x cluster.
```
/retest

manifests/11_service_monitor.yaml
```diff
@@ -7,7 +7,18 @@ metadata:
     name: marketplace-operator
 spec:
   endpoints:
-  - port: metrics
+    - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      interval: 30s
+      metricRelabelings:
+        - action: drop
```
fedepaol: Hey, landed here only because I need to do the same thing to a metric endpoint of mine. I suspect this relabeling section is not necessary, only the scheme part.

cmd/manager/main.go
```diff
@@ -69,6 +76,22 @@ func main() {
 		os.Exit(0)
 	}
 
+	// set TLS to serve metrics over a secure channel if cert is provided
+	// cert is provided by default by the marketplace-trusted-ca volume mounted as part of the marketplace-operator deployment
+	var useTLS bool
+	if *tlsCertPath != "" && *tlsKeyPath == "" || *tlsCertPath == "" && *tlsKeyPath != "" {
```
crawford, line 82: This is a very confusing line of code. Not only did I have to double check Golang's operator precedence, but I misread the alternating comparison operators while trying to understand this. Normally, I would shrug and move on, but for something related to TLS, I'm going to ask that this be rewritten in the simpler form:

```go
if (*tlsCertPath == "") != (*tlsKeyPath == "") {
```

PR #298, opened 2020-04-01
[deprecation] Prometheus alert for custom resources

anik120:

This PR introduces a Promethues alert to acompany the log and the
CVO status update in presence of a custom OperatorSource
or any CatalogSourceConfig.

**Reviewer Checklist**

### DIFF
--- a/manifests/12_prometheus_rule.yaml
+++ b/manifests/12_prometheus_rule.yaml
@@ -65,3 +65,13 @@ spec:
           for: 15m
           labels:
             severity: warning
+    - name: custom-resource-detected
+      rules:
+        - alert: CustomResourceDetected
+          annotations:
+            summary: Indicates the presence of a custom OperatorSource or a CatalogSourceConifg in the cluster
+            message: The cluster has custom OperatorSource/CatalogSourceConfig, which are deprecated from OCP 4.5. Upgrade to OCP 4.5 will not be possible unless those resources are removed. Please visit https://docs.openshift.com/container-platform/4.4/release_notes/ocp-4-4-release-notes.html#ocp-4-4-marketplace-apis-deprecated for further details.
+          expr: |
+            custom_resource_total > 0
+          labels:
+            severity: warning
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -3,8 +3,9 @@ package metrics
 import (
 	"crypto/tls"
 	"fmt"
-	"github.com/operator-framework/operator-marketplace/pkg/filemonitor"
 	"net/http"
+
+	"github.com/operator-framework/operator-marketplace/pkg/filemonitor"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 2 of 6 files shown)


awgreene:
@anik120 you misspelled "accompany" in your commit message.

pkg/metrics/metrics.go
```diff
@@ -45,6 +55,13 @@ var (
 		},
 		[]string{opSrcLabel},
 	)
+	customResourceGaugeVec = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "custom_resource_total",
+			Help: "A counter that stores information about the presense of a custom OperatorSource or any CatalogSourceConfig in the cluster.",
```
awgreene, line 61: This is a Gauge, not a Counter, which are different types in Prometheus.

pkg/metrics/metrics.go
```diff
@@ -127,3 +149,13 @@ func registerMetrics() error {
 func GetRoundTripper() http.RoundTripper {
 	return roundTripper
 }
+
+// RegisterCustomResource increases the count of the custom_resource_total metric
+func RegisterCustomResource(resourceType string) {
```
awgreene, line 154: Can this be named IncCustomResource since "Register" has it's own meaning in the prometheus client?

pkg/metrics/metrics.go
```diff
@@ -127,3 +149,13 @@ func registerMetrics() error {
...
+	customResourceGaugeVec.With(prometheus.Labels{ResourceTypeLabel: resourceType}).Inc()
+}
+
+// DeregisterCustomResource decreases the count of the custom_resource_total metric
+func DeregisterCustomResource(resourceType string) {
```
awgreene, line 159: Can this be named DecCustomResource since "Register" has it's own meaning in the prometheus client?

pkg/metrics/metrics.go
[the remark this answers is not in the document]
anik120: Right, I was using a counter before I decided to change to a gauge instead but forgot to change the comment. Thank you for noticing it.

anik120: Right, I was using a counter before I decided to change to a gauge instead but forgot to change the comment. Thank you for noticing it

pkg/metrics/metrics.go
anik120: @awgreene I like to think of the abstraction seen outside this particular package and how the name affects readability. 
For example, from the CSC/Opsrc controller, I `metrics.RegisterCustomResource`, which reads as `Register this custom resource with "metrics"`.
`metrics.IncCustomResource` when being read outside this package seems too technical (vs readable like an English sentence), and something that the reader might have to go into to understand the purpose of the function.

anik120:
@awgreene addressed your comments. Thanks for the review!

pkg/metrics/metrics.go
awgreene: I don't agree that you should consider cases of "being read outside this package" because Go code is shared in context of the package being imported.

My point was that Prometheus includes a "Register" function in it's package which the `metrics` package makes use of. However - this is a pretty mute point in a soon to be deprecated API so I'm not going to hold up the PR on this.

manifests/12_prometheus_rule.yaml
```diff
@@ -65,3 +65,13 @@ spec:
           for: 15m
           labels:
             severity: warning
+    - name: custom-resource-detected
```
ecordell, line 68: I think this could be better named, since `custom-resource` is a generic term for an object in an api defined by a CRD.

`custom-source-detected` and `custom_source_total`?

manifests/12_prometheus_rule.yaml
```diff
@@ -65,3 +65,13 @@ spec:
...
+      rules:
+        - alert: CustomResourceDetected
```
lilic, line 70: Hey, noticed this alert in an openshift cluster. 

The alertname is not very clear, to me custom resource detected can mean any custom resource, having a prefix of MarketplaceCustomResource would be a bit more clear.

Also this is a notification rather than an alert. See https://prometheus.io/docs/practices/alerting/#what-to-alert-on

Also the metric name is wrong, _total indicates a counter rather than a gauge like you have https://prometheus.io/docs/practices/naming/#metric-names

cc @awgreene @anik120 thanks :)

PR #302, opened 2020-04-16
[deprecation]: Deprecate CatalogSourceConfig API

anik120:

This PR removes the CatalogSourceConfig API in it's entirety.

**Reviewer Checklist**


pkg/operatorsource/configuring.go
```diff
@@ -105,9 +105,9 @@ func (r *configuringReconciler) Reconcile(ctx context.Context, in *v1.OperatorSo
 	}
 
 	// Now that we have updated the datastore, let's check if the opsrc is new.
-	// If it is, let's force a resync for CatalogSourceConfig.
+	// If it is, let's force a resync for CatalogSources.
 	if isResyncNeeded {
```
kevinrizza, line 109: This whole section of logic seems unnecessary now, right? I don't think this comment is correct, because the resyncer updates the cached app-registry info that the catalogsourceconfig controller uses to know about what metadata was pulled. Since they don't exist anymore, the whole cache sync just isn't needed anymore

test/testsuites/watchtests.go
```diff
@@ -25,35 +25,11 @@ const (
...
-	err := deleteCheckRestoreChild(t, "Deployment", v2.CatalogSourceConfigKind)
-	assert.NoError(t, err, cscErrMsg)
-}
-
-// testRestoreCscService tests that when a Service that is owned by a CatalogSourceConfig
-// is restored upon deletion.
-func testRestoreCscService(t *testing.T) {
-	err := deleteCheckRestoreChild(t, "Service", v2.CatalogSourceConfigKind)
-	assert.NoError(t, err, cscErrMsg)
-}
```
kevinrizza, line 55: Not blocking, just had a comment about the cache sync that exists between the two controllers -- I think we can just rip out any reference to the cache. But if that is a non trivial amount of work, then let's at least update the comments to reflect that more accurately

pkg/operatorsource/configuring.go
anik120: @kevinrizza that's what I thought too, but when I removed this the catalogsource that was being created from the opsrc was missing the packages it is supposed to have. I discovered that after this test failed when this logic was not there: https://github.com/operator-framework/operator-marketplace/blob/master/pkg/operatorsource/configuring_test.go#L27

anik120: @kevinrizza turns out that was because of an `Expect()` in the test, which I have removed, so the resync logic has been removed. FYI we still need to `WriteMetaDataToDataStore()`.
Updated PR.

cmd/manager/main.go
```diff
@@ -174,6 +174,18 @@ func main() {
...
 
+	// migrate away from Marketplace API
+	clientGo, err := client.New(cfg, client.Options{Scheme: mgr.GetScheme()})
+	if err != nil {
+		log.Error(err, "Failed to instantiate client for migrator")
+		os.Exit(1)
+	}
+	migrator := migrator.New(clientGo)
+	err = migrator.Migrate()
+	if err != nil {
```
ecordell, line 185: this can be a followup pr, but if the migration fails we should probably retry it a couple of times.

(failing to migrate is not the worst, the crd will just stick around)

anik120:
/hold cancel 
Cleaned up e2e tests

awgreene:
Looks like an e2e test is failing @anik120 
```
--- FAIL: TestMarketplace (945.42s)
    --- PASS: TestMarketplace/cluster-operator-status-test-group (65.29s)
        --- PASS: TestMarketplace/cluster-operator-status-test-group/cluster-operator-status-on-startup-test-suite (0.03s)
    	client.go:57: resource type OperatorSource with namespace/name (openshift-marketplace/test-operators-for-co) created
        --- PASS: TestMarketplace/cluster-operator-status-test-group/cluster-operator-status-on-custom-opsrc-creation-test-suite (25.05s)
        --- PASS: TestMarketplace/cluster-operator-status-test-group/cluster-operator-status-on-custom-opsrc-deletion-test-suite (40.05s)
    	client.go:75: resource type OperatorSource with namespace/name (openshift-marketplace/test-operators-for-co) successfully deleted
    --- FAIL: TestMarketplace/opsrc-test-group (669.18s)
    	client.go:57: resource type OperatorSource with namespace/name (openshift-marketplace/test-operators) created
        --- FAIL: TestMarketplace/opsrc-test-group/proxy-test-suite (0.06s)
            --- FAIL: TestMarketplace/opsrc-test-group/proxy-test-suite/opsrc-registry-includes-proxy-variables (0.06s)
```

cmd/manager/main.go
anik120: @ecordell till now I haven't seen any error that could be non-deterministic. If we fail once, we're most likely going to fail all the time. I'm not sure if blocking the operator because we're trying multiple times is better than failing fast and reporting the error.

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,107 @@
...
+	catsrc := olm.CatalogSource{}
+	err := kubeClient.Get(context.TODO(), client.ObjectKey{
+		Name:      csc.Name,
+		Namespace: csc.Spec.TargetNamespace},
+		&catsrc)
+	if err != nil {
+		return err
+	}
+	labels := catsrc.Labels
+	delete(labels, builders.CscOwnerNameLabel)
```
ecordell, line 74: This removes the labels, but I thought there were actual ownerrefs on the catalogsources? Is that not true?

anik120: @ecordell looks like we just set the labels: https://github.com/operator-framework/operator-marketplace/blob/master/pkg/operatorsource/configuring.go#L124-L129

kevinrizza: We don't use ownerrefs because catalogsourceconfigs are created across namespaces

PR #304, opened 2020-04-24
Bug 1827676: Fix panic when default Catsrc polling is not set

anik120:

This PR fixes the marketplace operator panic when the default
CatalogSource is configured without spec.updatestrategy.registry.interval
set.

**Reviewer Checklist**

### DIFF
--- a/pkg/defaults/catsrcHelpers.go
+++ b/pkg/defaults/catsrcHelpers.go
@@ -134,14 +134,24 @@ func AreCatsrcSpecsEqual(spec1 *olm.CatalogSourceSpec, spec2 *olm.CatalogSourceS
 	if spec1 == nil || spec2 == nil {
 		return false
 	}
-	if strings.EqualFold(string(spec1.SourceType), string(spec2.SourceType)) &&
-		strings.EqualFold(spec1.ConfigMap, spec2.ConfigMap) &&
-		strings.EqualFold(spec1.Address, spec2.Address) &&
-		strings.EqualFold(spec1.DisplayName, spec2.DisplayName) &&
-		strings.EqualFold(spec1.Publisher, spec2.Publisher) &&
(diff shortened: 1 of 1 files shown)


pkg/defaults/catsrcHelpers.go
```diff
@@ -139,8 +139,13 @@ func AreCatsrcSpecsEqual(spec1 *olm.CatalogSourceSpec, spec2 *olm.CatalogSourceS
 		strings.EqualFold(spec1.Address, spec2.Address) &&
 		strings.EqualFold(spec1.DisplayName, spec2.DisplayName) &&
 		strings.EqualFold(spec1.Publisher, spec2.Publisher) &&
-		strings.EqualFold(spec1.Image, spec2.Image) &&
-		spec1.UpdateStrategy.RegistryPoll == spec1.UpdateStrategy.RegistryPoll {
+		strings.EqualFold(spec1.Image, spec2.Image) {
+		if spec1.UpdateStrategy != nil && spec2.UpdateStrategy != nil {
```
kevinrizza, line 143: what happens when just one is nil? shouldn't we check for that as well?

pkg/defaults/catsrcHelpers.go
```diff
@@ -139,8 +139,13 @@ func AreCatsrcSpecsEqual(spec1 *olm.CatalogSourceSpec, spec2 *olm.CatalogSourceS
...
+			if spec1.UpdateStrategy.RegistryPoll == spec1.UpdateStrategy.RegistryPoll {
```
exdx, line 144: I think it would be more clear if we had the polling nil check done separately from the string checks. If the strings are not equal then we can return early, otherwise we can proceed to check the polling

anik120:
@kevinrizza @exdx updated the PR to alleviate both of your concerns. PTAL now. Thanks!

PR #300, opened 2020-04-10
[operatorhub] Accept CatalogSource as a default source

anik120:
**Description of the change:**

This PR enhances the OperatorHub API's capablity to accept a
CatalogSource as a default source along with OperatorSource.

**Motivation for the change:**
In Openshift 4.5, we want to be able to support both OperatorSources 
and CatalogSources as default source types. 

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage 
- [ ] Sufficient end-to-end test coverage
- [ ] Docs updated or added to `/docs` 
- [ ] Commit messages sensible and descriptive

PR #308, opened 2020-05-06
Bug 1831818: Remove Obsolete default OperatorSource

anik120:
In #300, the OperatorHub API was enhanced to accept both OperatorSource
and CatalogSource as defaults. However, when an existing OperatorSource
in a cluster was switched to a CatalogSource, the old OperatorSource
persisted. This PR fixes the issue, and removes the obsolete OperatorSource,
so that there's only one source for a catalog of operators.

**Reviewer Checklist**

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -198,6 +198,10 @@ func main() {
 		exit(err)
 	}
 
+	err = defaults.RemoveObsoleteOpsrc(clientGo)
+	if err != nil {
+		log.Error(err, "[defaults] Could not remove obsolete default OperatorSource/s")
+	}
 	// statusReportingDoneCh will be closed after the operator has successfully stopped reporting ClusterOperator status.
 	statusReportingDoneCh := statusReporter.StartReporting()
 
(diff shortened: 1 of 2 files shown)


cmd/manager/main.go
```diff
@@ -198,6 +198,10 @@ func main() {
 		exit(err)
 	}
 
+	err = defaults.RemoveObsoleteOpsrc(clientGo)
```
kevinrizza, line 201: Won't this only clean things up on restart? Do we want to make this part of the active defaults logic instead?

anik120: @kevinrizza that is a good point. What do you mean by `active defaults logic` though?

ecordell: Can't we only switch an operator source to a catalogsource after a restart? (because the definition is in the pod, the pod will need to be updated)

ecordell:
/approve

This looks good, I will leave it to @kevinrizza to lgtm since you had a comment

kevinrizza:
Misunderstood the way this api worked. On further investigation, this seems reasonable.

/lgtm

ecordell:
Note - this doesn't need to be cherry-picked

PR #322, opened 2020-08-06
Bug 1862481: Correctly count  non-default Marketplace CRs

awgreene:
Problem: It is possible for the marketplace operator to report that
deprecated CRs are present on cluster despite no CRs being present.

Solution: Marketplace already searches for non-default CRs on cluster
when reporting status. Marketplace will now use this search to report
the number of deprecated CRs on cluster.

### DIFF
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -15,6 +15,7 @@ import (
 	mktconfig "github.com/operator-framework/operator-marketplace/pkg/apis/config/v1"
 	v1 "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
 	"github.com/operator-framework/operator-marketplace/pkg/defaults"
+	"github.com/operator-framework/operator-marketplace/pkg/metrics"
 	"github.com/operator-framework/operator-marketplace/pkg/operatorhub"
 	log "github.com/sirupsen/logrus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


pkg/status/status.go
```diff
@@ -439,25 +440,27 @@ func (NoOpReporter) StartReporting() <-chan struct{} {
...
 	opsrcs := &v1.OperatorSourceList{}
+	nonDefaultOpsrcCount := 0
 	if err := kubeClient.List(context.TODO(), &client.ListOptions{}, opsrcs); err != nil {
-		return false, err
+		return nonDefaultOpsrcCount, err
 	}
 	for _, opsrc := range opsrcs.Items {
 		if !defaults.IsDefaultSource(opsrc.Name) {
-			return true, nil
+			nonDefaultOpsrcCount++
```
kevinrizza, line 462: Why do we need to keep a count at all? Why can't we just have knowledge of if they exist on the cluster?

awgreene: I do not have a strong preference, but we lose information regarding how many Marketplace CRs by going binary. If you can explain potential benefits, I'll be happy to make the change.

anik120: I actually think it's beneficial to have a count coz it enhances the UX for an admin. 

If it is binary: 
* Sees the alert
* Deletes an opsrc 
* Sees the alert still exists (but I already deleted a custom resource)

If it is a count: 
* Sees the alert
* Deletes a CR and sees the count go down 
* Alert tells the admin there's more CRs to be deleted.

awgreene:
Ci issue:
> time="2020-08-09T09:49:38Z" level=error msg="failed to remove file" error="could not run remote command: pod e2e-aws-console-olm does not have a host assigned"

/retest

jonesbr17:
@awgreene would it be possible to get this cherry-picked to release-4.4 as well? Thanks.

jonesbr17:
@awgreene Any updates on this? This bug is still present in 4.4. Can this be cherry-picked?

awgreene:
Good morning @jonesbr17, this bugfix was backported to 4.4 in #326.

PR #312, opened 2020-06-03
Removing deprecated links and section related to CatalogSourceConfig

amalkurup89:
**Description of the change:** Removing deprecated links and section related to CatalogSourceConfig

**Motivation for the change:**  Improving docs

PR #323, opened 2020-08-06
Bug 1866168: [deprecation] Remove OperatorSource API

anik120:

This PR removes the OperatorSource API from the cluster

**Reviewer Checklist**


pkg/metrics/metrics.go
```diff
@@ -123,36 +83,6 @@ func ServePrometheus(useTLS bool, cert, key string) error {
 // registerMetrics registers marketplace prometheus metrics.
 func registerMetrics() error {
```
njhale, line 84: Marketplace no longer registers _any_ metrics, so is there a reason to leave the metrics port open? compliance?

cmd/manager/main.go
```diff
@@ -153,7 +150,7 @@ func main() {
 	}
 
 	// Setup all Controllers
-	if err := controller.AddToManager(mgr, options.ControllerOptions{SyncSender: statusReporter}); err != nil {
```
njhale, line 156: Should the migration logs below reference `OperatorSource` now?

pkg/migrator/migrator.go
```diff
@@ -13,8 +13,8 @@ import (
 )
 
 const (
-	//CscCRDName is the CatalogSourceConfig CRD name
-	CscCRDName = "catalogsourceconfigs.operators.coreos.com"
+	//OpsrcCRDName is the OperatorSource CRD name
+	OpsrcCRDName = "operatorsources.operators.coreos.com"
```
njhale, line 17: Is there ever a case where we need to migrate away from both `CatalogSourceConfig` _and_ `OperatorSource` in a single upgrade?

Makefile
```diff
@@ -21,27 +21,20 @@ osbs-build:
...
 	# gomock and mockgen are never out of sync.
-	go install ./vendor/github.com/golang/mock/mockgen
+	# go install ./vendor/github.com/golang/mock/mockgen
 	
-	@echo making sure directory for mocks exists
-	mkdir -p $(MOCKS_DIR)
+	# @echo making sure directory for mocks exists
+	# mkdir -p $(MOCKS_DIR)
 
 	# $(mockgen) -destination=<Path/file where the mock is generated> -package=<The package that the generated mock files will belong to> -mock_names=<Original Interface name>=<Name of Generated mocked Interface> <Go package path of the original interface> <comma seperated list of the interface you want to mock>
```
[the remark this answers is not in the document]
anik120: I added this comment as part of [this PR](https://github.com/operator-framework/operator-marketplace/pull/147/files#r269754633). Leaving it in there for the same reason I had then.

Makefile
```diff
@@ -21,27 +21,20 @@ osbs-build:
...
+unit: unit-test
 
 unit-test:
 	go test -v ./pkg/...
 
 generate-mocks:
 	# Build mockgen from the same version used by gomock. This ensures that 
 	# gomock and mockgen are never out of sync.
-	go install ./vendor/github.com/golang/mock/mockgen
+	# go install ./vendor/github.com/golang/mock/mockgen
```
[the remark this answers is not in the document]
anik120: There are interfaces that can be mocked, but no unit test cases were written for them expect for OperatorSource related interfaces.

cmd/manager/main.go
```diff
@@ -153,7 +150,7 @@ func main() {
...
+	if err := controller.AddToManager(mgr, options.ControllerOptions{}); err != nil {
```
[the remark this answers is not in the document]
anik120: We do have the [OperatorHub controller](https://github.com/operator-framework/operator-marketplace/blob/master/pkg/operatorhub/handler.go#L28) still

pkg/controller/options/options.go
```diff
@@ -1,9 +1,4 @@
 package options
 
-import (
-	"github.com/operator-framework/operator-marketplace/pkg/status"
-)
-
 type ControllerOptions struct {
```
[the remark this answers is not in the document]
anik120: Same argument as [this comment](https://github.com/operator-framework/operator-marketplace/pull/323#discussion_r468016473)

pkg/metrics/metrics.go
anik120: I left it in there predicting we might want metrics for our default CatalogSources. Ripping them out now seems premature.

cmd/manager/main.go
anik120: Yes they should, missed this one. Updating it

pkg/migrator/migrator.go
anik120: That is a good point. Do we allow migration from suppose 4.3 to 4.6 directly? Or do users have to follow the migration chain, i.e 4.3->4.4->4.5->4.6 ?

cmd/manager/main.go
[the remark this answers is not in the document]
njhale: Sorry, this looked like a no-op until I investigated a little more.

Makefile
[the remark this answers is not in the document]
kevinrizza: But why not just remove everything related to mockgen if we're no longer using it?

manifests/03_operatorsource.crd.yaml
```diff
@@ -1,143 +0,0 @@
-apiVersion: apiextensions.k8s.io/v1beta1
```
kevinrizza, line 1: Shouldn't we also update all of these manifests in the `/deploy/upstream` folder?

https://github.com/operator-framework/operator-marketplace/tree/master/deploy/upstream

Makefile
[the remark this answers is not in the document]
anik120: I guess we can always go back and look at the git history if we ever do decide to mock interfaces for new test cases. Removing them.

manifests/03_operatorsource.crd.yaml
anik120: That is correct. Missed this one. Updating the PR with `/deploy/upstream` udpated.

manifests/05_role.yaml
```diff
@@ -26,7 +26,6 @@ rules:
   - operators.coreos.com
   resources:
   - catalogsources
-  - catalogsourceconfigs
   - operatorsources
```
awgreene, line 29: Can we remove the `operatorsources` resource as well? Given that we removed the OpSrc CRD from the manifests dir, there shouldn't be anything to query.

pkg/migrator/migrator.go
awgreene: I don't believe that we allow skipping minor OpenShift versions, but we should check before merging this as customers may be using CSCs or OpSrcs and this change will result in data loss.

pkg/migrator/migrator.go
```diff
@@ -40,39 +40,39 @@ func New(client client.Client) Migrator {
 }
 func (m *migrator) Migrate() error {
 
-	cscs := &v2.CatalogSourceConfigList{}
-	err := m.client.List(context.TODO(), &client.ListOptions{}, cscs)
+	opsrcs := &v1.OperatorSourceList{}
```
awgreene, line 43: Given that the OpSrc CRD was removed from the manifests dir, shouldn't CVO delete the OpSrc CRD from the cluster? If CVO removes the CRD, these CatSrc's will be removed, which shouldn't be a problem if Marketplace recreates them on startup.

pkg/status/status.go
```diff
@@ -372,24 +264,17 @@ func NewReporter(cfg *rest.Config, mgr manager.Manager, namespace string, name s
...
 	}
 
 	return &reporter{
-		configClient: configClient,
-		rawClient:    rawClient,
-		namespace:    namespace,
-		version:      version,
-		syncRatio:    syncRatio,
-		// Add a buffer to prevent dropping syncs
-		syncCh:              make(chan error, 25),
```
awgreene, line 392: Do we need the receiver on [line 405](https://github.com/operator-framework/operator-marketplace/pull/323/files#diff-a5666326aefdf43a8e89fd4871449fa0L405)

pkg/status/status.go
```diff
@@ -225,14 +186,6 @@ func (r *reporter) syncChannelReceiver() {
 		select {
 		case <-r.stopCh:
 			return
-		case err := <-r.syncCh:
-			if err == nil {
-				r.syncRatio.ReportSyncEvent()
-			} else {
-				r.syncRatio.ReportFailedSync()
-			}
-			failedSyncs, syncs := r.syncRatio.GetSyncs()
```
awgreene, line 234: It seems like we could remove some of the other methods that existed when this was more complicated, such as `ensureClusterOperator` and `setStatusCondition`, etc...

manifests/03_operatorsource.crd.yaml
awgreene: Can you grep for all instances of `OperatorSource` and `CatalogSourceConfig` and make sure that no count is returned?

pkg/status/status.go
anik120: Not sure what that is linking to: https://github.com/operator-framework/operator-marketplace/blob/8ffe135acd11523609bc2dd3ef1c5312a6111f11/pkg/status/status.go
The receiver has been ripped off since we're not keeping track of any syncs anymore.

pkg/status/status.go
anik120: I think we should write one function per task even if the system isn't very complicated. I am all for code readability whether the system is simple or complex

pkg/migrator/migrator.go
kevinrizza: You can't skip minor versions

pkg/migrator/migrator.go
anik120: The migrator is more for any non-default operatorsources that weren't cleared out by the admin. 
CVO will block the upgrade if there are non-default operatorsources in the cluster before upgrade, but the migrator's job is to clean up the resources if that block from CVO was overridden somehow. 

All the migrator does, is simply remove the owner refs from the child resources(i.e CatalogSource in this case) and deleted the operatorsource.

pkg/status/status.go
awgreene: https://github.com/operator-framework/operator-marketplace/blob/8ffe135acd11523609bc2dd3ef1c5312a6111f11/pkg/status/status.go#L290

manifests/05_role.yaml
anik120: Since the migrator needs to list the non-default operatorsources in the cluster ( (ref to this comment)[https://github.com/operator-framework/operator-marketplace/pull/323#discussion_r468836213] ) we need permission to list and delete the operatorsources.

pkg/status/status.go
awgreene: Right - but I feel like a lot of the complexity in this code is still present in the methods that were left behind. The methods themselves should be simplified.

pkg/migrator/migrator.go
awgreene: Am I missing something? This code will only be available when 4.6 is released.In 4.6, openshift will no longer have the OpSrc and CatSrc CRDs from the cluster. How will marketplace remove a CR for a CRD that doesn't exist?

pkg/status/status.go
anik120: Absolutely, good catch!

pkg/migrator/migrator.go
awgreene: Deleting a CRD when a CR has a finalizer will prevent the CRD from being deleted as outlined in this [issue](https://github.com/kubernetes/kubernetes/issues/60538).

pkg/migrator/migrator.go
```diff
@@ -40,39 +40,39 @@ func New(client client.Client) Migrator {
...
 	if err != nil {
 		return err
 	}
 	allErrors := []error{}
-	for _, csc := range cscs.Items {
-		err = removeCSCOwnerRefFromCatalogSource(&csc, m.client)
+	for _, opsrc := range opsrcs.Items {
+		err = removeOpsrcOwnerRefFromCatalogSource(&opsrc, m.client)
 		if err != nil {
 			allErrors = append(allErrors, err)
```
awgreene, line 52: Might be worth checking if the OpSrc API exists before running this code so we don't populate 404 errors.

pkg/migrator/migrator.go
awgreene, line 43: Might be worth checking if the OpSrc API exists before running this code so we don't populate 404 errors.

awgreene: This is a nit since we should only hit this error when someone overrides CVO.

anik120: Updated the call site to not log the error if the error is `IsNotFound`.

pkg/status/status.go
anik120: This work can be done as part of [this bug](https://bugzilla.redhat.com/show_bug.cgi?id=1857680)

manifests/03_operatorsource.crd.yaml
anik120: OperatorSource: 
```$ grep -nr 'operatorsource' .
./Makefile:7:OPERATORSOURCE_MOCK_PKG := operatorsource_mocks
Binary file ./build/_output/bin/operator-marketplace-local matches
Binary file ./build/_output/bin/marketplace-operator matches
./pkg/apis/operators/v1/operatorsource_types.go:14:	OpSrcFinalizer = "finalizer.operatorsources.operators.coreos.com"
./pkg/apis/operators/v1/operatorsource_types.go:19:// related to OperatorSource type should be added to operatorsource.go file.
./pkg/migrator/migrator.go:17:	OpsrcCRDName = "operatorsources.operators.coreos.com"
./docs/troubleshooting.md:14:3. [Operators(s) in an OperatorSource fail to show up in the UI](#operators-in-an-operatorsource-fail-to-show-up-in-the-ui) 
./docs/troubleshooting.md:37:`oc describe opsrc <Name-of-OperatorSource> -n openshift-marketplace`. [This section](#operatorSource-failing-to-connect-to-the-datastore) and [the following section](#operators-in-an-operatorsource-fail-to-show-up-in-the-ui) of the doc also contains helpful information about debugging an OperatorSource. 
./docs/troubleshooting.md:109:A common reason for an OperatorSource to not reconcile correctly is the use of the wrong [endpoint](https://github.com/operator-framework/operator-marketplace/blob/master/deploy/examples/community.operatorsource.cr.yaml#L8). Please make sure that the endpoint is specified as `https://quay.io/cnr`.
./docs/e2e-testing.md:36:    // Create the operatorsource.
./vendor/github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver/resolver.go:77:	// create a map of operatorsourceinfo (subscription+catalogsource data) to the original subscriptions
./README.md:12:The operator manages two CRDs: [OperatorSource](./deploy/upstream/03_operatorsource.crd.yaml) and [CatalogSourceConfig](./deploy/upstream/02_catalogsourceconfig.crd.yaml).
./README.md:73:$ kubectl apply -f deploy/upstream/07_upstream_operatorsource.cr.yaml
./README.md:142:[upstream-community-operators]: deploy/upstream/07_upstream_operatorsource.cr.yaml
./README.md:143:[community-operators]: deploy/examples/community.operatorsource.cr.yaml
./manifests/05_role.yaml:29:  - operatorsources
./deploy/upstream/05_role.yaml:19:  - operatorsources
```

CatalogSourceConfigs: 
```$ grep -nr 'catalogsourceconfigs' .
Binary file ./build/_output/bin/operator-marketplace-local matches
Binary file ./build/_output/bin/marketplace-operator matches
./pkg/apis/operators/v2/catalogsourceconfig_types.go:15:	CSCFinalizer = "finalizer.catalogsourceconfigs.operators.coreos.com"
```
For the README updates, operatorsource could be included in this PR: https://github.com/operator-framework/operator-marketplace/pull/312

PR #335, opened 2020-08-27
Bug 1873546: Increase registry delay for large payloads

kevinrizza:
Fixes community and certified operators pods crash looping because their startup time exceeds their readiness probe timeout. Increase that timeout from 30s to 60s.

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -400,7 +400,7 @@ func (r *registry) newPodTemplateSpec(registryCommand []string, needServiceAccou
 								Command: action,
 							},
 						},
-						InitialDelaySeconds: 30,
+						InitialDelaySeconds: 60,
 						FailureThreshold:    10,
 					},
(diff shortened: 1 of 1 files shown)


sdodson:
Clearing stale cherry-pick-approved.

markmc:
@kevinrizza I've updated the PR description to give a bit more information. Would have liked to see that sort of detail in the commit message

Left you a question in the bz; holding back on cherry-pick-approved while you look into that. Thanks

markmc:
> @kevinrizza I've updated the PR description to give a bit more information. Would have liked to see that sort of detail in the commit message
> 
> Left you a question in bz #1873012; holding back on cherry-pick-approved while you look into that. Thanks

Answered my own question. At least one of the reports of this issue was with 30s. My only concern is whether 60s (+ 10s period and threshold of 10) will be enough. Approving.

PR #336, opened 2020-08-27
Bug 1873250: Recreate default CatalogSources if deleted/edited

anik120:

This PR implementes a watch for default CatalogSources, to recreate them
if they are deleted or edited.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/catalogsource/catalogsource_controller.go
+++ b/pkg/controller/catalogsource/catalogsource_controller.go
@@ -1,0 +1,130 @@
+package catalogsource
+
+import (
+	"context"
+	"time"
+
+	olm "github.com/operator-framework/operator-marketplace/pkg/apis/olm/v1alpha1"
+	"github.com/operator-framework/operator-marketplace/pkg/controller/options"
+	"github.com/operator-framework/operator-marketplace/pkg/defaults"
+	"github.com/operator-framework/operator-marketplace/pkg/operatorhub"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// Add creates a new CatalogSource Controller and adds it to the Manager. The Manager will set fields on the Controller
+// and Start it when the Manager is Started.
+func Add(mgr manager.Manager, _ options.ControllerOptions) error {
+	return add(mgr, newReconciler(mgr))
+}
+
+func newReconciler(mgr manager.Manager) reconcile.Reconciler {
+	client := mgr.GetClient()
+	return &ReconcileCatalogSource{
+		client: client,
+	}
+}
+
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+
+	c, err := controller.New("catalogsource-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	_, defaultCatalogsources := defaults.GetGlobalDefinitions()
+	pred := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if _, ok := defaultCatalogsources[e.MetaOld.GetName()]; ok {
+				return true
+			}
+			return false
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			if _, ok := defaultCatalogsources[e.Meta.GetName()]; ok {
+				// If DeleteStateUnknown is true it implies that the Delete event was missed
+				// and we can ignore it.
+				if e.DeleteStateUnknown {
+					return false
+				}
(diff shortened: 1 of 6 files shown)


pkg/controller/catalogsource/catalogsource_controller.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	if instance.DeletionTimestamp != nil {
+		if !operatorhub.GetSingleton().Get()[defaultCatsrcDef.Name] {
+			createNewCatsrcInstance(r.client, defaultCatsrcDef)
+		}
+	}
+
+	if !defaults.AreCatsrcSpecsEqual(&defaultCatsrcDef.Spec, &instance.Spec) {
+		if err := r.client.Delete(context.TODO(), instance); err != nil {
```
exdx, line 117: I'm not sure if we delete here that it will be gone by the time the `createNewCatsrcInstance()` is called in the next line -- does it make more sense to return here with a `Result{requeue: true}` here instead and have the above logic recreate it? 

Or alternatively, we could just update it directly if the specs don't match instead of delete/recreate. Using the SSA client would ensure this is a seamless process.

pkg/controller/catalogsource/catalogsource_controller.go
```diff
@@ -0,0 +1,138 @@
...
+
+		}
+	}
+	return reconcile.Result{}, nil
+}
+
+func createNewCatsrcInstance(client client.Client, catsrc olm.CatalogSource) error {
+	err := client.Create(context.TODO(), &catsrc)
+	if err != nil {
+		log.Warnf("Could not recreate default CatalogSource %s. Error: %s", catsrc.GetName(), err.Error())
```
exdx, line 135: nit: I think we should return the `err` after the log and return nil below.

pkg/controller/catalogsource/catalogsource_controller.go
```diff
@@ -0,0 +1,138 @@
...
+
+// ReconcileCatalogSource reconciles a CatalogSource object
+type ReconcileCatalogSource struct {
+	// This client, initialized using mgr.Client() above, is a split client
+	// that reads objects from the cache and writes to the apiserver
+	client client.Client
+}
+
+func (r *ReconcileCatalogSource) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	log.Infof("Reconciling default CatalogSource %s", request.Name)
```
exdx, line 92: nit: generally OLM logs are in lowercase

pkg/controller/catalogsource/catalogsource_controller.go
```diff
@@ -0,0 +1,138 @@
...
+			if !operatorhub.GetSingleton().Get()[defaultCatsrcDef.Name] {
+				createNewCatsrcInstance(r.client, defaultCatsrcDef)
+			}
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	if instance.DeletionTimestamp != nil {
```
exdx, line 110: I see what the check here is but it seems like a race against the GC controller. If its marked for deletion, but not gone yet, recreate it. I wonder if this is a necessary as it would only apply in the case where it was deleted right after the `Get()`, otherwise you would be recreating an object still on the cluster. 

But maybe other controllers use this pattern, I'm not sure.

pkg/controller/catalogsource/catalogsource_controller.go
anik120: Updated to return `Result{requeue:true, requeueAfter: time.Seconds * 5}` 

Deleting and recreating seems much cleaner, we used to do that for OperatorSources too

pkg/controller/catalogsource/catalogsource_controller.go
anik120: Done.

pkg/controller/catalogsource/catalogsource_controller.go
exdx: GC = garbage collector

anik120: @exdx I put this in because the reconcile request could be received before the object is actually removed (even though it has been marked for deletion). The comments [here](https://github.com/operator-framework/operator-marketplace/blob/release-4.5/pkg/controller/operatorsource/operatorsource_controller.go#L89-L108) might make that clearer. 

If we don't put in this clause, a sequence of event might look like this: 
1) Object has been marked for deletion 
2) Reconciler sends a get request, and gets the object, checks if the spec has been tampered with, and sees nothing wrong and does nothing 
3) Object is deleted (and no reconciliation request is sent again coz it has already been sent)

pkg/controller/catalogsource/catalogsource_controller.go
exdx: LGTM

pkg/controller/catalogsource/catalogsource_controller.go
exdx: I see the logic you're describing, makes sense. But I'm concerned about the case where the object is marked for deletion, we issue the get request on the next line and it succeeds, and _then_ the GC controller deletes it immediately afterwards (so the `createNew()...` doesn't get called). Basically a race against when the GC controller actually deletes the object. 

Maybe its safer to requeue in the case where its marked for deletion?

anik120: I see what you mean. Update to re-queue the request after 5 secs instead of calling `createNewCatsrcInstance()`

exdx:
/lgtm 
I think the flow now is more intuitive. Basically `GET` the catalog at the beginning of the reconciliation loop, and if it can't be found, create a new one. 

Further on, if we encounter issues with the catalog (the spec doesn't equal the expected spec, or its marked for deletion) delete it and requeue, so the first `GET` will fail eventually and we will recreate it.

ecordell (approved):
/approve

I had some comments but don't think they need to block.

PR #338, opened 2020-09-06
manifests/10_clusteroperator: Drop namespace from cluster-scoped resource

wking:
ClusterOperator objects are [cluster-scoped and not namespaced][1].  Dropping the namespace here reverts the ClusterOperator namespace addition from 5a41001689 (#83).

[1]: https://github.com/openshift/api/blob/7ac89ba6b97136f09e68e06c32638dae78cad516/config/v1/types_cluster_operator.go#L9

### DIFF
--- a/manifests/10_clusteroperator.yaml
+++ b/manifests/10_clusteroperator.yaml
@@ -2,7 +2,6 @@ apiVersion: config.openshift.io/v1
 kind: ClusterOperator
 metadata:
   name: marketplace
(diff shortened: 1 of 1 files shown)


wking:
No need to retest.  This won't land without a bug before 4.6 forks from master, and there are no real consequences to justify a bug.

timflannagan:
Duplicate of #401 now?

/close

PR #340, opened 2020-09-10
Bug 1877564: Sync remote app-registry

anik120:
The syncing of remote app-registry was removed by #302 since it was
    tied to the syncing of the child CatalogSourceConfig which was removed
    in #302. This PR puts the syncing of remote app-registry back in, minus
    the syncing of the non-existing child catalogSourceConfig.

**Reviewer Checklist**


eparis:
/hold
since we played games with bugzilla instead of giving QE the opportunity to test in 4.6 before we merge to 4.5 I am holding this PR. The hold should be removed when either QE tests before we merge to 4.5 or when QE approves they will be able to contain testing such a fix in the extremely quick z-stream timeline.

anik120:
> since we played games with bugzilla instead of giving QE the opportunity to test in 4.6

@eparis What's the procedure to get a bug fix in for a bug that's only valid for n-1th release and not the nth release? We've got a few of these kind of bugs in before, but I recall having the option of setting the parent bug to `verified` stating the feature being fixed has been removed from the nth release. 

Is there a `/bugzilla ignore-parent` command or equivalent? Or do we let our customers wait around for the bureaucratic process to take it's time? 

Background: We've deprecated an API in 4.5, and another in 4.6, so there's a potential we might land in other situations like these where customers complain about a bug that's only valid for the n-1th or n-2th release till the time we're sticking with bugzilla.

PR #342, opened 2020-09-15
Fix upstream deployment

awgreene:
Problem: The `registryServerImage` runtime argument was recently removed
from marketplace but the marketplace deployment defined for the upstream
build did not remove the argument which caused the upstream marketplace
pod to crash.

Solution: Update the deployment manifest so it does not include the
`registryServerImage` runtime argument.


cliveseldon:
Any reason this is not being merged? @awgreene

ryandawsonuk:
@kevinrizza would you be able to approve this?

PR #347, opened 2020-10-06
Bug 1885376: Remove condition around marketplace OperatorAvailable status update

ankitathomas:
marketplace-operator install fails intermittently, with marketplace never setting its status as available.
Marketplace checks for a preexisting false available condition for making this update, which can 
incorrectly lead to the install getting stuck. Removing this wrapping condition to allow marketplace to
set its availability to true without this issue.

### DIFF
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -125,7 +125,7 @@ func (r *reporter) setStatusCondition(statusCondition configv1.ClusterOperatorSt
 func (r *reporter) updateStatus(previousStatus *configv1.ClusterOperatorStatus) error {
 	var err error
 	if compareClusterOperatorStatusConditionArrays(previousStatus.Conditions, r.clusterOperator.Status.Conditions) {
-		log.Debugf("[status] Previous and current ClusterOperator Status are the same, the ClusterOperator Status will not be updated.")
+		log.Infof("[status] Previous and current ClusterOperator Status are the same, the ClusterOperator Status will not be updated.")
 	} else {
 		log.Debugf("[status] Previous and current ClusterOperator Status are different, attempting to update the ClusterOperator Status.")
 
@@ -225,14 +225,11 @@ func (r *reporter) monitorClusterStatus() {
 			}
 
 			// Report that marketplace is available
-			if cohelpers.IsStatusConditionFalse(r.clusterOperator.Status.Conditions, configv1.OperatorAvailable) {
-				reason := "OperatorAvailable"
-				conditionListBuilder := clusterStatusListBuilder()
-				conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionFalse, fmt.Sprintf("Successfully progressed to release version: %s", r.version), reason)
-				statusConditions := conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionTrue, fmt.Sprintf("Available release version: %s", r.version), reason)
-				statusErr = r.setStatus(statusConditions)
-				break
-			}
+			reason := "OperatorAvailable"
+			conditionListBuilder := clusterStatusListBuilder()
+			conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionFalse, fmt.Sprintf("Successfully progressed to release version: %s", r.version), reason)
+			statusConditions := conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionTrue, fmt.Sprintf("Available release version: %s", r.version), reason)
+			statusErr = r.setStatus(statusConditions)
 		}
 	}
 }


pkg/status/status.go
```diff
@@ -225,14 +225,11 @@ func (r *reporter) monitorClusterStatus() {
...
 			// Report that marketplace is available
-			if cohelpers.IsStatusConditionFalse(r.clusterOperator.Status.Conditions, configv1.OperatorAvailable) {
-				reason := "OperatorAvailable"
-				conditionListBuilder := clusterStatusListBuilder()
-				conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionFalse, fmt.Sprintf("Successfully progressed to release version: %s", r.version), reason)
-				statusConditions := conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionTrue, fmt.Sprintf("Available release version: %s", r.version), reason)
-				statusErr = r.setStatus(statusConditions)
-				break
-			}
+			reason := "OperatorAvailable"
```
ecordell, line 228: why not just create it in the succeeded state to start with? all this really confirms is that the our `timer` works, right? (which seems like a proxy for "is the pod alive" which we'll already have a signal for)

kevinrizza: +1

It feels like this whole reporting mechanism should be ripped out and we can just have a function that does this.

anik120: @ecordell @kevinrizza if we rip out the whole thing, we will essentially rip out the ability to report status based on the availability of the default catalogsources (the way we used to with default CatalogSources). 

Are we saying that we don't forsee a requirement coming in that'll ask that of marketplace ever?

anik120: Also, we still want a thread to keep checking if the operator is up, and if not[ report `OperatorExited` ](https://github.com/operator-framework/operator-marketplace/blob/master/pkg/status/status.go#L190 )right?

anik120: > all this really confirms is that the our timer works, right? 

This thread running an infinite loop actually does three things: 
1) On startup, creates the clusteroperator and sets the message to `Determining Status` and condition to `Available:False` `Progressing: True` `Degraded: False`
2) Once the startup is done it sets the condition to `Available:True` `Progressing: False` `Degraded: False`
3) If the deployment is deleted it sets the clusteroperator condition to `Available:False` with `Reason: OperatorExited`.

I'm guessing we want the deployment to be monitored and the clusteroperator status to be set as `Available:False` if the deployment is deleted.

anik120 (commented):
@ankitathomas after the PR is finalized, it's probably a good idea to retest a few times (5-10?) and report the status of those tests (if any are failing with `clusteroperator Determining Status`) in the comments to make sure we've solved the problem before we merge this.

pkg/status/status.go
ecordell: The current loop is really :

1) On startup, creates the clusteroperator and sets the message to `Determining Status` and condition to `Available:False` `Progressing: True` `Degraded: False`
2) 20 seconds later, regardless of what else has happened, sets `Available:True` `Progressing: False` `Degraded: False`
3) If the deployment is deleted it sets the clusteroperator condition to `Available:False` with `Reason: OperatorExited`.

I'm just saying that if there's nothing (aside from the stopchannel) that really prevents us from going available anymore, let's just make this:

1) On startup, creates the clusteroperator and sets the message to Available:True` `Progressing: False` `Degraded: False`
2) If the deployment is deleted it sets the clusteroperator condition to `Available:False` with `Reason: OperatorExited`.

>  where do we have the signal for that?

I just mean that the pod will be failing and there are already cluster alerts for failing pods.

ankitathomas:
Test run 1:
test okd-e2e-aws failed due to amazon connectivity error:
Post "https://iam.amazonaws.com/": read tcp 10.128.4.236:43018->52.94.225.3:443: read: connection reset by peer"

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/operator-framework_operator-marketplace/347/pull-ci-operator-framework-operator-marketplace-master-okd-e2e-aws/1313567790354927616

anik120:
Test run 2: 

test e2e-aws failed due to error: 

```fail [k8s.io/kubernetes@v1.19.0/test/e2e/kubectl/kubectl.go:524]: Unexpected error:
    <exec.CodeExitError>: {
        Err: {
            s: "error running /usr/bin/kubectl --server=https://api.ci-op-wn6s5c4z-8d2e4.origin-ci-int-aws.dev.rhcloud.com:6443 --kubeconfig=/var/run/secrets/ci.openshift.io/multi-stage/kubeconfig --namespace=e2e-kubectl-8324 run -i --image=docker.io/library/busybox:1.29 --restart=Never success -- /bin/sh -c exit 0:\nCommand stdout:\n\nstderr:\nerror: timed out waiting for the condition\n\nerror:\nexit status 1",
        },
        Code: 1,
    }
    error running /usr/bin/kubectl --server=https://api.ci-op-wn6s5c4z-8d2e4.origin-ci-int-aws.dev.rhcloud.com:6443 --kubeconfig=/var/run/secrets/ci.openshift.io/multi-stage/kubeconfig --namespace=e2e-kubectl-8324 run -i --image=docker.io/library/busybox:1.29 --restart=Never success -- /bin/sh -c exit 0:
    Command stdout:
    
    stderr:
    error: timed out waiting for the condition
    
    error:
    exit status 1
```
Rest of the tests are green. 

/retest all

ankitathomas:
Test run 3:
Test e2e-aws-serial failed due to errors relating to kube proxy:
``` bash
curl -q -s --connect-timeout 1 http://localhost:10249/proxyMode
command terminated with exit code 7 
```
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/operator-framework_operator-marketplace/347/pull-ci-operator-framework-operator-marketplace-master-e2e-aws/1313625080806248448

ecordell:
/lgtm
/approve

I still think this can be improved, but it's fine to merge as-is. I don't see this adding any new failures to the system, so we can gather data faster about this fixing the issue by merging and seeing it across all platform CI.

pkg/status/status.go
awgreene: At the time that this was created, there was an e2e test which ensured that operators don't immediately report available. This was over a year ago and might have changed.

awgreene: I reached out to Trevor King and this test does not seem to exist anymore.

PR #354, opened 2020-10-23
Bug 1838352: Don't report OperatorExited to ClusterOperator

awgreene:
Problem: The Marketplace operator should not report that it is in a bad state to Telemeter when it exits gracefully.

Solution: No longer report that the operator has exited when the operator shuts down gracefully.

### DIFF
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -188,17 +188,7 @@ func (r *reporter) monitorClusterStatus() {
 	for {
 		select {
 		case <-r.stopCh:
-			// If the stopCh is closed, set all ClusterOperatorStatus conditions to false.
-			reason := "OperatorExited"
-			msg := "The operator has exited"
-			conditionListBuilder := clusterStatusListBuilder()
-			conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionFalse, msg, reason)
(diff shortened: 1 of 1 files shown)


pkg/status/status.go
```diff
@@ -188,17 +188,7 @@ func (r *reporter) monitorClusterStatus() {
...
-			msg := "The operator has exited"
-			conditionListBuilder := clusterStatusListBuilder()
-			conditionListBuilder(configv1.OperatorProgressing, configv1.ConditionFalse, msg, reason)
-			conditionListBuilder(configv1.OperatorAvailable, configv1.ConditionFalse, msg, reason)
-			statusConditions := conditionListBuilder(configv1.OperatorDegraded, configv1.ConditionFalse, msg, reason)
-			statusErr := r.setStatus(statusConditions)
-			if statusErr != nil {
-				log.Error("[status] " + statusErr.Error())
-			}
+			log.Error("[status] Operator no longer reporting status")
```
kevinrizza, line 191: Do we necessarily want to log an error here at all? It seems like this should just be an INFO message if this gets logged just because the operator is stopping gracefully.

kevinrizza (commented):
Otherwise this lgtm

pkg/status/status.go
awgreene: Good point - updated.

PR #358, opened 2020-11-02
Bug 1891993: clean up child resources for old opsrcs

ankitathomas:
Clean up service and deployments associated with old opsrcs.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -4,9 +4,14 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"github.com/operator-framework/operator-marketplace/pkg/builders"
+	v1 "k8s.io/api/apps/v1"
+	v12 "k8s.io/api/core/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"net/http"
 	"os"
 	"runtime"
+	"strings"
 	"time"
 
 	"github.com/operator-framework/operator-marketplace/pkg/metrics"
@@ -186,6 +191,11 @@ func main() {
 		log.Error(err, "[migration] Error in migrating Marketplace away from OperatorSource API")
 	}
 
+	err = cleanUpOldOpsrcResources(clientGo)
+	if err != nil {
+		log.Error(err, "OperatorSource child resource cleanup failed")
+	}
+
 	// Handle the defaults
 	err = ensureDefaults(cfg, mgr.GetScheme())
 	if err != nil {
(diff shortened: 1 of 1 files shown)


dinhxuanvu (approved):
/lgtm

pkg/migrator/migrator.go
```diff
@@ -91,6 +95,37 @@ func deleteOpsrc(opsrc *v1.OperatorSource, client client.Client) error {
 	if err != nil {
 		return err
 	}
+
+	// delete child resources
+	objMeta := metav1.ObjectMeta{
```
ecordell, line 100: this will fix the issues for clusters upgrading from 4.5.z -> 4.6.z, but not clusters that are updating from 4.6.0 / 4.6.1 -> 4.6.z, right?

dinhxuanvu: I think this fix on marketplace in combination with the changes on OLM should resolve the issue on 4.6.z update.

kevinrizza: I don't believe that's the case Evan, this migrator runs whenever the marketplace operator starts up in 4.6, so it would fix the problem for both update cases.

kevinrizza: Ignore my last comment, I didn't realize this was iterating over the list of operator sources.

IMO we should actually just look at the deployments and services in the marketplace namespace with the right annotations and forcefully delete them.

anik120: @kevinrizza I don't know if that's any different than what's being done here. This solution looks fine to me unless I'm missing something.

anik120: Ignore my last comment, I'm +1 with @ecordell and @kevinrizza it'll be safer to just  wipe the namespace clean so that we don't have to depend on any OLM changes.

pkg/defaults/defaults.go
```diff
@@ -182,23 +179,3 @@ func populateDefsConfig(dir string) (map[string]v1.OperatorSource, map[string]ol
 	}
 	return opsrcDefinitions, catsrcDefinitions, config, nil
 }
-
-// RemoveObsoleteOpsrc removes any existing default OperatorSource
-// in the cluster that has been switched into a default CatalogSource
-func RemoveObsoleteOpsrc(kubeClient client.Client) error {
```
anik120, line 188: This was added when the Operatorhub was enhanced to accept both operatorsources and CatalogSources. After that change, when an operatorsource was switched to catalogsource, the `obsolete` operatorsources were still hanging around, [so it was necessary to `RemoveObsoleteOpsrc`](https://github.com/operator-framework/operator-marketplace/pull/308).

With that context, it makes sense for 4.7 to remove this since we don't have opsrc in 4.7, and therefore no opportunity to switch opsrc to catsrc (and therefore no necessity to `RemoveObsoleteOpsrc`).
However, if we're going to backport this all the way till 4.5, we might want to keep this around. Otherwise removing this will re-introduce [this](https://bugzilla.redhat.com/show_bug.cgi?id=1831818) bug.

anik120 (changes requested):
Good work so far @ankitathomas. Had a couple of thoughts.

pkg/defaults/defaults.go
ankitathomas: It would make sense to have this removed on a separate PR. I'll create one for it

kevinrizza:
I'm not sure that I love the idea of this function sitting directly in main.go, (seems like it would be better in the migrator as another step in addition to the other removal stuff) but I don't want to block this fix over that minor problem.

/approve

anik120:
>I'm not sure that I love the idea of this function sitting directly in main.go, (seems like it would be better in the migrator as another step in addition to the other removal stuff) but I don't want to block this fix over that minor problem.

FWIW I had the same thought, but didn't wanna block this PR for the same reason.

PR #359, opened 2020-11-05
Bug 1868770: Allow catsrc with default catsrc name in disconnected env

anik120:

With the introduction of #336, when a CatalogSource with the same name as that of a
default CatalogSource name was being created, the new CatalogSource was either being
restored to the default Spec, or being removed if the CatalogSource was disabled via
OperatorHub. This made is impossible to create a CatalogSource with the same name as
that of the default CatalogSource in a disconnected cluster enviroment. This PR fixes
the issue by allowing for creation/update/deleteion of a CatalogSource with the same
name as that of a default CatalogSource, if it has been disabled via OperatorHub first.
When the default CatalogSource is renabled via OperatorHub, the default CatalogSource
Spec is restored by overriding the custom created CatalogSource.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/catalogsource/catalogsource_controller.go
+++ b/pkg/controller/catalogsource/catalogsource_controller.go
@@ -89,18 +89,22 @@ type ReconcileCatalogSource struct {
 }
 
 func (r *ReconcileCatalogSource) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	log.Infof("Reconciling default CatalogSource %s", request.Name)
 
 	_, defaultCatalogsources := defaults.GetGlobalDefinitions()
 	defaultCatsrcDef := defaultCatalogsources[request.Name]
+
+	if operatorhub.GetSingleton().Get()[defaultCatsrcDef.Name] {
+		log.Info("%s disabled. Not taking any action", defaultCatsrcDef.Name)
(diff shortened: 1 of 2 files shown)


exdx:
Looks good, just wondering if we should document this bit somewhere 

> When the default CatalogSource is renabled via OperatorHub, the default CatalogSource
Spec is restored by overriding the custom created CatalogSource.

This could catch some customers by surprise, if they are enabling default catalogs for _some_ default catalogs but not their custom catalog specifically, which has the same name and would get deleted  

/lgtm

anik120:
> Looks good, just wondering if we should document this bit somewhere
 
@adellape fyi

kevinrizza:
/hold

Holding this so that another higher priority bug can merge first

anik120:
/hold cancel

Looks like this is next in line, all other outstanding PRs have merged.

reggie-k:
Hmmm what immediate workaround can I use?
Running in a disconnected env, just upgraded from 4.5 to 4.6, this is a terrible regression for me.
OperatorHub disconnected worked like a charm  on 4.5, was used a lot in our org.

reggie-k:
Ok, I just created the same catalog with a custom name, works like a charm again

PR #361, opened 2020-11-09
Format logs info

jianzhangbjz:

Just format the log output.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/catalogsource/catalogsource_controller.go
+++ b/pkg/controller/catalogsource/catalogsource_controller.go
@@ -94,7 +94,7 @@ func (r *ReconcileCatalogSource) Reconcile(request reconcile.Request) (reconcile
 	defaultCatsrcDef := defaultCatalogsources[request.Name]
 
 	if operatorhub.GetSingleton().Get()[defaultCatsrcDef.Name] {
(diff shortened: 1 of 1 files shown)


kevinrizza:
@jianzhangbjz looks like the bugzilla attached to this pr isn't related to the actual fix? The bz in your header references a bug that has already been verified.

Also it seems like this is something that most likely won't be approved for backport the openshift patch maintainers, can you just remove the bugzilla reference so this will still land in master?

anik120:
> @jianzhangbjz looks like the bugzilla attached to this pr isn't related to the actual fix? The bz in your header references a bug that has already been verified.

Also it seems like this is something that most likely won't be approved for backport the openshift patch maintainers, can you just remove the bugzilla reference so this will still land in master?

In that case this won't be backported to 4.6/4.5

jianzhangbjz:
> can you just remove the bugzilla reference so this will still land in master?

@kevinrizza Yes, sure, thanks for your reminder. Removed it.

jianzhangbjz:
@kevinrizza Could you help to cherry-pick it to release-4.6? Thanks!

PR #364, opened 2020-11-09
Bug 1896160: Marketplace should report when it is degraded

awgreene:
Problem: Marketplace does not always report the degraded
ClusterStatusConditionType when it is reporting its status to CVO.

Solution: Marketplace should report when it is degraded and ensure that
the degraded condition is present.


anik120 (approved):
/lgtm

ankitathomas (approved):
/lgtm

PR #365, opened 2020-11-09
Bug 1892382: Fix operator exited message

awgreene:

PR #371, opened 2020-12-18
Bug 1909217: Do not report degraded when Marketplace exits gracefully

awgreene:
Problem: The Marketplace operator should not report that it is in a bad
state to Telemeter when it exits gracefully.

Solution: No longer report that the operator has exited when the
operator shuts down gracefully.


awgreene:
/hold

I want to run some additional testing.

ashcrow:
https://github.com/operator-framework/operator-marketplace/pull/365

PR #372, opened 2020-12-18
Bug 1908431: Preserve custom catsrc w/ default catsrc name on restart

anik120:
In #362, marketplace's catalogsource reconciler was updated to allow custom
Catalogsources with the same name as one of the default Catalogsources, when
the default Catalogsource was disabled first with the operatorhub api. However,
when marketplace was restarted, the custom catalogsource was being force deleted
by marketplace's operatorhub controller.
This PR adds a special annotation to the default CatalogSources, and updates the
operator to delete any catalogsource with the same name as one of the default only
if the CatalogSource has that special annotation.

**Reviewer Checklist**


benluddy:
Continued in #373.

PR #373, opened 2020-12-23
Bug 1908431: Preserve custom catsrc w/ default catsrc name on restart

benluddy:
Superseding #372.

In #362, marketplace's catalogsource reconciler was updated to allow
custom Catalogsources with the same name as one of the default
Catalogsources, when the default Catalogsource was disabled first with
the operatorhub api. However, when marketplace was restarted, the
custom catalogsource was being force deleted by marketplace's
operatorhub controller.

This PR adds a special annotation to the default CatalogSources, and
updates the operator to delete any catalogsource with the same name as
one of the default only if the CatalogSource has that special
annotation. In addition, the behavior of the marketplace operator as
it relates to reconciling the default CatalogSources should no longer
be tied to its process lifecycle.

### DIFF
--- a/pkg/controller/catalogsource/catalogsource_controller.go
+++ b/pkg/controller/catalogsource/catalogsource_controller.go
@@ -2,14 +2,13 @@ package catalogsource
 
 import (
 	"context"
-	"time"
 
 	olm "github.com/operator-framework/operator-marketplace/pkg/apis/olm/v1alpha1"
+	v1 "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
 	"github.com/operator-framework/operator-marketplace/pkg/controller/options"
 	"github.com/operator-framework/operator-marketplace/pkg/defaults"
 	"github.com/operator-framework/operator-marketplace/pkg/operatorhub"
 	log "github.com/sirupsen/logrus"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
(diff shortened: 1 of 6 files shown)


pkg/defaults/catsrcHelpers.go
```diff
@@ -64,7 +64,9 @@ func processCatsrc(client wrapper.Client, def olm.CatalogSource, disable bool) e
 	}
 
 	if disable {
-		err = ensureCatsrcAbsent(client, def, cluster)
+		if cluster.Annotations[defaultCatsrcAnnotationKey] == defaultCatsrcAnnotationValue {
```
njhale, line 67: This probably requires that the default `CatalogSource` manifests contain the annotation: https://github.com/operator-framework/operator-marketplace/tree/master/defaults

pkg/defaults/defaults.go
```diff
@@ -38,6 +38,11 @@ var (
 	defaultConfig = make(map[string]bool)
 )
 
+const (
+	defaultCatsrcAnnotationKey   string = "source"
```
njhale, line 42: We need to namespace this.

pkg/defaults/catsrcHelpers.go
```diff
@@ -106,8 +114,8 @@ func ensureCatsrcPresent(client wrapper.Client, def olm.CatalogSource, cluster *
 		return nil
 	}
 
-	if AreCatsrcSpecsEqual(&def.Spec, &cluster.Spec) {
-		logrus.Infof("[defaults] CatalogSource %s default and on cluster specs are same", def.Name)
+	if cluster.Annotations[defaultCatsrcAnnotationKey] == defaultCatsrcAnnotationValue && AreCatsrcSpecsEqual(&def.Spec, &cluster.Spec) {
```
njhale, line 117: So it looks like we stomp the "custom default" catalogs when the defaults are enabled again. For consistency, should we update the annotation as well in this situation?

I imagine the current situation looks like this:
1. disable default
2. default deleted
3. create "custom default" without annotation
4. enable default
5. "custom default" has spec stomped, but doesn't get the annotation
6. disable default
7. default is not deleted because it doesn't have the annotation (?)

benluddy: Yes, thanks for catching this. I discarded a separate patch for this issue that did what you're suggesting and forgot to fold that into this one.

pkg/defaults/catsrcHelpers.go
benluddy: The annotation is added at runtime instead of requiring it to be present on the canonical default manifests.

benluddy:
The console tests appear to have been broken since Monday (based on [the history of the equivalent OLM job](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-operator-framework-operator-lifecycle-manager-master-e2e-aws-console-olm)).

pkg/controller/catalogsource/catalogsource_controller.go
```diff
@@ -89,39 +88,8 @@ type ReconcileCatalogSource struct {
...
-	}
-
-	if !defaults.AreCatsrcSpecsEqual(&defaultCatsrcDef.Spec, &instance.Spec) {
-		if err := r.client.Delete(context.TODO(), instance); err != nil {
-			log.Warnf("Could not set default CatalogSource %s's spec back to desired default state. Error in deleting updated CatalogSource: %s", defaultCatsrcDef.GetName(), err.Error())
-		}
-		return reconcile.Result{Requeue: true, RequeueAfter: time.Second * 5}, nil
-	}
-	return reconcile.Result{}, nil
+	return reconcile.Result{}, defaults.New(map[string]v1.OperatorSource{}, defaultCatalogsources, operatorhub.GetSingleton().Get()).Ensure(r.client, request.Name)
```
kevinrizza, line 92: 👍

benluddy:
Backport to 4.6 is [here](https://github.com/operator-framework/operator-marketplace/pull/375).

benluddy:
https://github.com/openshift/console/pull/7835 should address the `e2e-aws-console-olm` job failure.

/test e2e-aws-upgrade

PR #375, opened 2021-01-13
Bug 1915905: Preserve custom catsrc w/ default catsrc name

benluddy:
In #362, marketplace's catalogsource reconciler was updated to allow
custom Catalogsources with the same name as one of the default
Catalogsources, when the default Catalogsource was disabled first with
the operatorhub api. However, when marketplace was restarted, the
custom catalogsource was being force deleted by marketplace's
operatorhub controller.

This PR adds a special annotation to the default CatalogSources, and
updates the operator to delete any catalogsource with the same name as
one of the default only if the CatalogSource has that special
annotation. In addition, the behavior of the marketplace operator as
it relates to reconciling the default CatalogSources should no longer
be tied to its process lifecycle.

### DIFF
--- a/pkg/controller/catalogsource/catalogsource_controller.go
+++ b/pkg/controller/catalogsource/catalogsource_controller.go
@@ -2,14 +2,13 @@ package catalogsource
 
 import (
 	"context"
-	"time"
 
 	olm "github.com/operator-framework/operator-marketplace/pkg/apis/olm/v1alpha1"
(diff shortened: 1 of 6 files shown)


benluddy:
/test e2e-aws-operator

Filed https://github.com/operator-framework/operator-marketplace/issues/376 to track the flaky end-to-end test.

hasbro17:
@operator-framework/patch-managers PTAL

PR #382, opened 2021-02-11
Bug 1927456: update defaults to 4.7

kevinrizza:


bparees:
overriding the valid-bz flag, this is a simple change, easily verified, and we don't want to hold up 4.7 while we verify the change in 4.8.  We will ensure the change does go into 4.8 as well (actually for 4.8 we'll change to the 4.8 catalog).

PR #391, opened 2021-04-16
Bug 1950417: Migrate to Go modules, bump k8s dependencies to v0.20.6 and sdk to v0.19.x

timflannagan:

**Reviewer Checklist**


timflannagan:
Oh man the openshift/release for this repository needs some work.

/hold

timflannagan:
Related to those image failures: https://github.com/operator-framework/operator-marketplace/pull/392.

Updating configuring for this repository in https://github.com/openshift/release/pull/17802.

timflannagan:
Oh, right we need to bump the Go build root image before we can get real CI feedback as the more up-to-date k8s APIs use features in the more recent core Go packages.

cmd/manager/main.go
```diff
@@ -4,140 +4,152 @@ import (
...
 	"runtime"
-	"strings"
-	"time"
 
-	"github.com/operator-framework/operator-marketplace/pkg/metrics"
-	"github.com/operator-framework/operator-marketplace/pkg/migrator"
+	"github.com/sirupsen/logrus"
 
 	apiconfigv1 "github.com/openshift/api/config/v1"
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/signals"
```
timflannagan, line 21: Ended up vendoring the signal handling package from OLM into pkg/signals and removing the OLM dependency entirely. IIRC, there's a controller runtime signal handling library too that we may want to explore.

cmd/manager/main.go
```diff
@@ -4,140 +4,152 @@ import (
...
+		clusterOperatorName string
+		tlsKeyPath          string
+		tlsCertPath         string
+		version             bool
+	)
 	flag.StringVar(&clusterOperatorName, "clusterOperatorName", "", "the name of the OpenShift ClusterOperator that should reflect this operator's status, or the empty string to disable ClusterOperator updates")
+	flag.StringVar(&defaults.Dir, "defaultsDir", "", "the directory where the default CatalogSources are stored")
+	flag.BoolVar(&version, "version", false, "displays marketplace source commit info.")
+	flag.StringVar(&tlsKeyPath, "tls-key", "", "Path to use for private key (requires tls-cert)")
+	flag.StringVar(&tlsCertPath, "tls-cert", "", "Path to use for certificate (requires tls-key)")
```
timflannagan, line 80: Minor change: just moved these from global variables to the inside the scope of the main function.

cmd/manager/main.go
```diff
@@ -4,140 +4,152 @@ import (
...
-	if *tlsCertPath != "" && *tlsKeyPath == "" || *tlsCertPath == "" && *tlsKeyPath != "" {
-		log.Warn("both --tls-key and --tls-crt must be provided for TLS to be enabled, falling back to non-https")
-	} else if *tlsCertPath == "" && *tlsKeyPath == "" {
-		log.Info("TLS keys not set, using non-https for metrics")
-	} else {
-		log.Info("TLS keys set, using https for metrics")
-		useTLS = true
-	}
-	err := metrics.ServePrometheus(useTLS, *tlsCertPath, *tlsKeyPath)
+	err := metrics.ServePrometheus(tlsCertPath, tlsKeyPath)
```
timflannagan, line 91: Another minor change: moved the `useTLS` logic in its own function in the metrics library and updated the `ServePrometheus` function to remove that `useTLS` parameter and instead call that helper function in its flow.

cmd/manager/main.go
```diff
@@ -4,140 +4,152 @@ import (
...
 	// Create a new Cmd to provide shared dependencies and start components
 	// Even though we are asking to watch all namespaces, we only handle events
 	// from the operator's namespace. The reason for watching all namespaces is
 	// watch for CatalogSources in targetNamespaces being deleted and recreate
 	// them.
-	mgr, err := manager.New(cfg, manager.Options{Namespace: ""})
+	mgr, err := manager.New(cfg, manager.Options{
+		Namespace:              "",
+		HealthProbeBindAddress: ":8080",
+		MetricsBindAddress:     "0",
```
timflannagan, line 125: This is an artifact of controller-runtime making a breaking change from defaulting to port 8383 to 8080, which conflicts with the port that we expose for readiness and liveness probes. Setting this value to `0` disables metrics, so we'll need to double-check if that's the behavior we want, vs. updating the port where we serve our health checks.

cmd/manager/main.go
```diff
@@ -147,64 +159,72 @@ func main() {
...
-	go http.ListenAndServe(":8080", nil)
+	logrus.Info("Setting up health checks")
+	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+		logrus.Error(err, "unable to set up health check")
+		os.Exit(1)
+	}
+	if err := mgr.AddReadyzCheck("healthz", healthz.Ping); err != nil {
+		logrus.Error(err, "unable to set up ready check")
+		os.Exit(1)
+	}
```
timflannagan, line 188: Updated the liveness and readiness probes to use the controller-runtime healthz package instead of handling it ourselves.

cmd/manager/main.go
```diff
@@ -147,64 +159,72 @@ func main() {
...
 
 	// Wait until this instance becomes the leader.
-	log.Info("Waiting to become leader.")
-	err = leader.Become(context.TODO(), "marketplace-operator-lock")
+	logrus.Info("Waiting to become leader.")
+	err = leader.Become(context.TODO(), leaderElectionConfigMapName)
```
timflannagan, line 192: Unlikely to change, but I moved the name of the leader election ConfigMap to a package-level constant to avoid hardcoding this value. It's worth re-evaluating whether we want to continue using the leader-for-life implementation, like we're using here, vs. the controller-runtime leader-for-lease implementation that's configurable in the manager instance.

cmd/manager/main.go
```diff
@@ -266,30 +287,31 @@ func ensureDefaults(cfg *rest.Config, scheme *kruntime.Scheme) error {
...
-	deploy := &v1.DeploymentList{}
-	svc := &v12.ServiceList{}
-	o := &client.ListOptions{}
-	if err := o.SetLabelSelector(strings.Join([]string{builders.OpsrcOwnerNameLabel, builders.OpsrcOwnerNamespaceLabel}, ",")); err != nil {
-		return err
+
+	deploy := &appsv1.DeploymentList{}
+	svc := &corev1.ServiceList{}
+	o := []client.ListOption{
+		client.MatchingLabels{builders.OpsrcOwnerNameLabel: builders.OpsrcOwnerNamespaceLabel},
```
timflannagan, line 294: This is needed as there were breaking changes in the function signatures in the controller-runtime Client interface. The new signature expects a list of options vs. a single instance of a ListOption.

pkg/controller/operatorhub/operatorhub_controller.go
```diff
@@ -49,16 +49,10 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	// We only care if the event came from the cluster config.
 	pred := predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
-			if e.Meta.GetName() == operatorhub.DefaultName {
-				return true
-			}
-			return false
+			return e.Meta.GetName() == operatorhub.DefaultName
```
timflannagan, line 52: This applies to the other changes in this package, but just simplifying the logic for these predicate functions to avoid having to make uneeded additional comparisons.

test/helpers/helpers.go
```diff
@@ -220,6 +225,27 @@ func UpdateRuntimeObject(client test.FrameworkClient, obj runtime.Object) error
...
+			err = client.Get(context.TODO(), key, clone)
+			if err != nil {
+				return err
+			}
+
+			obj.SetResourceVersion(clone.GetResourceVersion())
+		}
+		return err
+	})
+}
```
timflannagan, line 247: Open to suggestions on this implementation and whether there's something easier to implement without mixing client-go and controller-runtime mechanisms. This was mainly needed to cut down on flakes as we don't have any retry on conflict mechanism when we're attempting to update any of the APIs or CatalogSource resources that we manage.

test/testsuites/clusteroperatorstatustests.go
```diff
@@ -19,14 +19,11 @@ import (
 // defines the status of the marketplace operator has the correct status upon initialization. It
 // also confirms that the ClusterOperator's RelatedObjects contains the expected list of objects.
 func ClusterOperatorStatusOnStartup(t *testing.T) {
-	ctx := test.NewTestCtx(t)
+	ctx := test.NewContext(t)
```
timflannagan, line 22: The `NewTextCtx` has been deprecated in favor of `NewContext`.

test/testsuites/defaultcatsrctests.go
```diff
@@ -143,69 +151,68 @@ func testDefaultCatsrcWhileDisabled(t *testing.T) {
...
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "redhat-operators",
+			Name:      catSrcName,
 			Namespace: namespace,
 		},
 		Spec: olm.CatalogSourceSpec{
-			SourceType:  "grpc",
+			SourceType: "grpc",
+			// TODO(tflannag): Do we care that this image will cause an ImagePullBackOff?
```
timflannagan, line 159: Just mirroring the TODO comment here - in the context of this test I don't think we care as we're just ensuring the reconciliation behavior but using this image will result in the CatalogSource Pod being deployed with an image we can't authenticate to and the result is the Pod lands in the `ImagePullBackOff` state and we don't verify that Pod is healthy after updating.

test/testsuites/defaultcatsrctests.go
```diff
@@ -143,69 +151,68 @@ func testDefaultCatsrcWhileDisabled(t *testing.T) {
...
 }
 
 // checkForCatsrc checks if CatalogSource is present, and is not being removed after some time has passed
 func checkForCatsrc(name, namespace string) (olm.CatalogSource, error) {
 	client := test.Global.Client
-	// Wait for a minute
-	wait.Poll(time.Second*5, time.Minute*1, func() (done bool, err error) {
-		return false, nil
-	})
+	time.Sleep(60 * time.Second)
```
timflannagan, line 215: Another minor detail, but there's likely a better implementation than just blindly waiting a minute for changes to propagate.

cmd/manager/main.go
timflannagan: Also, it looks like we don't have a test that ensures the Prometheus ServiceMonitor resource hasn't regressed, so we may need to call that out in a TODO comment. I verified we're still serving Prometheus metrics ourselves locally:

```bash
#! /bin/bash

set -o errexit
set -o nounset
set -o pipefail

token=$(oc -n openshift-monitoring sa get-token prometheus-k8s)
host=$(oc -n openshift-monitoring get routes prometheus-k8s -o jsonpath={.spec.host})

# Note on the usage:
# Curl is particularly annoying to pass variables to parameters, so in order to
# a query successfully you need to ensure that the `""` and `''` are properly formatted
# in order to avoid a parsing error. You either need to avoid using the same string
# deliminitor entirely or character escape them.
#
# The following shows this issue as an example:
# $ ./scripts/run-prom-query.sh 'rate(container_cpu_usage_seconds_total{image!="", container_name!="POD"}[5m])'
curl -k -s -G -H "Authorization: Bearer $token" --data-urlencode "query=$1" https://${host}/api/v1/query | faq -f json
```
```bash
$ ./scripts/run-prom-query.sh 'sum(kube_pod_container_resource_requests_cpu_cores{namespace="openshift-marketplace"}) by (pod, namespace, node)' | wc -l
63
```

timflannagan: It looks like in the past we were carrying a PrometheusRule YAML manifest, but I didn't dig into whether we still care about serving our own custom metrics as it looks like they were mainly centered around (the now deprecated) app registry implementation.

When poking the Prometheus rules endpoint locally, it looks like we're not registering any custom rules anymore.

cmd/manager/main.go
timflannagan: I need to verify this locally as well, but something I'm realizing now is whether we'll need to bump the initial delay on the health check probes to accommodate the healthz.Ping check.

scripts/run-e2e-job.sh
```diff
@@ -5,7 +5,7 @@ set -o nounset
 set -o pipefail
 
 ROOT_DIR=$(dirname "${BASH_SOURCE[0]}")/..
-SDK_VERSION=v0.3.0
+SDK_VERSION=${SDK_VERSION:=v0.19.1}
 
 # Get operator-sdk binary.
 wget -O /tmp/operator-sdk https://github.com/operator-framework/operator-sdk/releases/download/${SDK_VERSION}/operator-sdk-${SDK_VERSION}-x86_64-linux-gnu && chmod +x /tmp/operator-sdk
```
timflannagan, line 11: Also, we may want to migrate towards having a dedicated built root Dockerfile if we're requiring operator-sdk being installed when running e2e tests, rather than curling a specific release binary when running this script.

ankitathomas (commented):
Joe is still missing from the reviewers list, but this looks good overall, great work!
/lgtm

anik120 (changes requested):
Thanks for taking this on and addressing a mammoth of a tech debt @timflannagan 🎉

test/e2e/main_test.go
```diff
@@ -6,13 +6,13 @@ import (
 	olm "github.com/operator-framework/operator-marketplace/pkg/apis/olm/v1alpha1"
 	"github.com/operator-framework/operator-marketplace/test/helpers"
 	"github.com/operator-framework/operator-marketplace/test/testgroups"
-	"github.com/operator-framework/operator-sdk/pkg/test"
+	f "github.com/operator-framework/operator-sdk/pkg/test"
```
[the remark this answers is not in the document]
timflannagan: That's just what was in the operator-sdk samples/documentation/etc. that I was relying on when going through the release changelogs. I don't have a strong opinion either way about how we name a third-party testing import.

cmd/manager/main.go
timflannagan: Ah, good to know - either way, the infrastructure is still there we just need to update the metric registration handler. Just wanted to call out this behavior anyways.

go.mod
```diff
@@ -0,0 +1,59 @@
+module github.com/operator-framework/operator-marketplace
```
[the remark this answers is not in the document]
timflannagan: Until today, I had it organized like that, and I still have a backup branch locally that has that mass number of commits. At the end of the day, it was still way too difficult to parse through commits that also have a ton of vendor changes, so I elected for a single commit that did the full migration.

cmd/manager/main.go
timflannagan: This is for the metrics that controller runtime exposes, which to my understanding, weren't being exposed anyways when I poke around a <4.8 marketplace instance. I can play around with getting these metrics enabled and exposed.

test/testsuites/operatorhubtests.go
```diff
@@ -439,8 +440,8 @@ func checkCatsrcIsDeleted(name, namespace string) error {
 	err := wait.Poll(time.Second*5, time.Minute*1, func() (done bool, err error) {
 		def := &olm.CatalogSource{}
 		err = client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: namespace}, def)
-		if !errors.IsNotFound(err) || def.Name != "" {
-			return false, fmt.Errorf("default CatalogSource still present on the cluster")
+		if errors.IsAlreadyExists(err) || def.Name != "" {
```
benluddy, line 443: When would GET return 409?

test/helpers/helpers.go
```diff
@@ -220,6 +223,26 @@ func UpdateRuntimeObject(client test.FrameworkClient, obj runtime.Object) error
 	)
 }
 
+func UpdateCatalogSourceWithRetries(client test.FrameworkClient, obj *olmv1alpha1.CatalogSource) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		err := client.Update(context.TODO(), obj)
+		if err != nil && apierrors.IsConflict(err) {
```
benluddy, line 229:
```suggestion
		if apierrors.IsConflict(err) {
```

cmd/manager/main.go
```diff
@@ -4,140 +4,146 @@ import (
...
 
-	log.Print("Registering Components.")
-
+	logrus.Info("Registering Components.")
+	logrus.Info("setting up scheme")
 	// Setup Scheme for all defined resources
 	if err := apis.AddToScheme(mgr.GetScheme()); err != nil {
-		exit(err)
+		logrus.Error(err)
+		os.Exit(1)
```
benluddy, line 131: May as well replace all of these with https://pkg.go.dev/github.com/sirupsen/logrus?utm_source=godoc#Fatalf while you're changing this.

cmd/manager/main.go
```diff
@@ -4,140 +4,146 @@ import (
...
+	flag.StringVar(&tlsKeyPath, "tls-key", "", "Path to use for private key (requires tls-cert)")
+	flag.StringVar(&tlsCertPath, "tls-cert", "", "Path to use for certificate (requires tls-key)")
 	flag.Parse()
 
 	// Check if version flag was set
-	if *version {
-		fmt.Print(sourceCommit.String())
-		// Exit immediately
+	if version {
+		logrus.Infof(sourceCommit.String())
```
benluddy, line 74: Probably harmless, but `sourceCommit` isn't supposed to be interpreted as a format string.

```suggestion
		logrus.Info(sourceCommit.String())
```
or

```suggestion
		logrus.Infof("%s", sourceCommit.String())
```

cmd/manager/main.go
benluddy: Nice!

> A little copying is better than a little dependency.

https://go-proverbs.github.io/

test/testsuites/operatorhubtests.go
timflannagan: Good catch - I think I was looking to reverse the double negative for easier readability, but I'm just going to revert that change entirely as I'm not 100% convinced that the current check is correct.

benluddy:
/hold

> level=fatal msg=failed to initialize the cluster: Cluster operator marketplace is not available

cmd/manager/main.go
```diff
@@ -147,64 +144,64 @@ func main() {
...
 	clientGo, err := client.New(cfg, client.Options{Scheme: mgr.GetScheme()})
 	if err != nil && !k8sErrors.IsNotFound(err) {
-		log.Error(err, "Failed to instantiate client for migrator")
-		os.Exit(1)
+		logrus.Fatal(err, "Failed to instantiate the client for migrator")
 	}
 	migrator := migrator.New(clientGo)
 	err = migrator.Migrate()
 	if err != nil {
-		log.Error(err, "[migration] Error in migrating Marketplace away from OperatorSource API")
```
benluddy, line 191: It looks like some of these log calls were accidentally changed to fatal (https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/operator-framework_operator-marketplace/391/pull-ci-operator-framework-operator-marketplace-master-e2e-aws/1389397643536371712/artifacts/e2e-aws/gather-extra/artifacts/pods/openshift-marketplace_marketplace-operator-6c6799c7cc-9r89t_marketplace-operator.log).

anik120: Fyi we can even get rid of the migrator now, can be a separate PR too but I'm not opposed to getting it done in this PR (since we need a separate bz for it if we want this in 4.8)

timflannagan: @benluddy Thanks, that was a poor replacement job. I updated a couple of instances from Fatal back to Error.

@anik120 I'm down with removing it in this PR or adding a TODO comment saying this can be removed. I feel like removing the migrator should be verified separately though, so we may want to pursue the latter option.

timflannagan: Actually, is the migrator removable? Are we still backwards compatible with earlier versions of OCP without it?

anik120: This migrator was for migrating any remaining OperatorSources to CatalogSources in 4.5(or 4.6, one of those). Now that we don't have the OperatorSource, all it does it tries to get the OperatorSources in the cluster, but encounters the error that there's no CRD called OperatorSource, and logs `[migration] Error in migrating Marketplace away from OperatorSource API` and exits out. 

> I feel like removing the migrator should be verified separately though, so we may want to pursue the latter option.

Totally understandable if we don't feel comfortable doing this now. FWIW though, given what I've said above, it's a pretty innocuous change, at least from the dev team's perspective.

benluddy: Let's keep that separate, this is already a substantial PR.

timflannagan:
Most of those e2e tests look like permafailures due to the CSI alert firing.

[This may warrant further investigation](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/operator-framework_operator-marketplace/391/pull-ci-operator-framework-operator-marketplace-master-e2e-aws-upgrade/1389573989793271808):

```
ail [github.com/openshift/origin/test/e2e/upgrade/upgrade.go:152]: during upgrade to registry.build02.ci.openshift.org/ci-op-hpimhtr2/release@sha256:bd393f79138ae68b4bb433873c4c5c5b4fe3cb66462ad4f6b5f4d913969c55ab
Unexpected error:
    <*errors.errorString | 0xc0038a0750>: {
        s: "Cluster did not complete upgrade: timed out waiting for the condition: deployment openshift-marketplace/marketplace-operator is Progressing=False: ProgressDeadlineExceeded: ReplicaSet \"marketplace-operator-85b87bfd4b\" has timed out progressing.",
    }
    Cluster did not complete upgrade: timed out waiting for the condition: deployment openshift-marketplace/marketplace-operator is Progressing=False: ProgressDeadlineExceeded: ReplicaSet "marketplace-operator-85b87bfd4b" has timed out progressing.
occurred
```

timflannagan:
Yeah, it looks like there's an alert firing in the marketplace namespace for a Pod crash looping in the upgrade tests:

```
alert KubePodCrashLooping fired for 5780 seconds with labels: {container="marketplace-operator", endpoint="https-main", job="kube-state-metrics", namespace="openshift-marketplace", pod="marketplace-operator-85b87bfd4b-kp9st", service="kube-state-metrics", severity="warning"}
```

timflannagan:
Ah, yeah looks like an issue with the liveness probe failing after upgrading:

```
Liveness probe error: Get "http://10.129.0.76:8080/healthz": context deadline exceeded (Client.Timeout exceeded while awaiting headers) body:
--
```

I spun up a 4.7 CRC cluster and updated the marketplace deployment to use a custom image and wasn't able to reproduce the crash looping Pod. However, I did see that using the controller-runtime health check implementations can cause the marketplace container to fail liveness and readiness checks frequently, presumably as we don't have any real probe configurations (e.g. an initial delay) configured in the deployment. I'm going to revert back to the original implementation that instantly reports back a 200 status.

anik120 (approved):

PR #395, opened 2021-04-22
Bug 1946790: Update clusteroperator status conditions on startup

anik120:

Previously, the conditions for clusteroperator marketplace was set to
available:false, progressing:true, upgradable:true, degraded:false with the
message "Progressing towards release version..." on startup of the marketplace
operator. This was done because a test in the openshift/origin test suites checked
to make sure operators did not set their condition to available:true immediately
on startup. That test has however been removed.
This PR sets the conditions for clusteroperator marketplace to available:true,
progressing:false, degreaded:false and upgradable:true on startup instead of
waiting for a particular time period.

**Reviewer Checklist**

### DIFF
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -24,12 +24,9 @@ const (
 	// coStatusReportInterval is the interval at which the ClusterOperator status is updated
 	coStatusReportInterval = 20 * time.Second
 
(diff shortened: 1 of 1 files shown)


ankitathomas (approved):
/lgtm

PR #399, opened 2021-05-06
Remove dead code

ecordell:
Initial pass, just to see what things are left.

I'm not sure I'm fully grokking what the configmap controller is doing.


njhale:
I think this is stale :)

PR #402, opened 2021-05-19
Bug 1936585: Configure alerts for Non-Ready Default CatalogSources

anik120:

With the [introduction of the `catalogsource_ready` metric in olm](https://github.com/operator-framework/operator-lifecycle-manager/pull/2152), alerts can be
fired for the default CatalogSources marketplace deploys if they are in	a non-ready state.
This PR introduces prometheus alerts for any default CatalogSources that have
been in a Non-Ready state for more than 10 mins.

**See it in action:**

Screenshot from 2021-05-19 15-32-32

`CommunityOperatorsCatalogError` alert being fired on a deployment of marketplace with an intentional bad image for `community-operators`

**Reviewer Checklist**

### DIFF
--- a/manifests/12_prometheus_rule.yaml
+++ b/manifests/12_prometheus_rule.yaml
@@ -1,0 +1,46 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: marketplace-alert-rules
+  namespace: openshift-marketplace
+  labels:
+    prometheus: alert-rules
+    role: alert-rules
+spec:
+  groups:
+    - name: marketplace.community_operators.rules
+      rules:
+        - alert: CommunityOperatorsCatalogError
+          annotations:
+            message: Default OperatorHub source "community-operators" is in Non-Ready state for more than 10 mins.
+          expr: catalogsource_ready{name="community-operators",exported_namespace="openshift-marketplace"} == 0
+          for: 10m
+          labels:
+            severity: warning
+    - name: marketplace.certified_operators.rules
+      rules:
+        - alert: CertifiedOperatorsCatalogError
+          annotations:
+            message: Default OperatorHub source "certified-operators" is in Non-Ready state for more than 10 mins.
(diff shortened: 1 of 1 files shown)


anik120:
@jianzhangbjz not sure if you guys do this the way I do, but fyi if you need to build a version of marketplace that is different from master (to deploy a CatalogSource with a bad image for testing, for example), the `Dockerfile` in this repo is outdated. I have a PR open to update the Dockerfile that can be used instead https://github.com/operator-framework/operator-marketplace/pull/400/files

hasbro17 (commented):
/lgtm

manifests/12_prometheus_rule.yaml
```diff
@@ -0,0 +1,46 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: marketplace-alert-rules
+  namespace: openshift-marketplace
+  labels:
+    prometheus: alert-rules
+    role: alert-rules
```
timflannagan, line 8: Any significance to these labels? Mainly curious as I didn't see anything consistent with our labeling in the root list of manifests.

/lgtm

timflannagan: It looks like we sometimes use the `name: marketplace-operator` in some manifests. We may need to take some time at some point in the next release and populate the k8s recommended labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/.

anik120: @timflannagan that is a good point. I'm not sure why/if these labels are necessary tbh, for this PR I mainly stuck to what we were doing previously , eg in 4.4 https://github.com/operator-framework/operator-marketplace/blob/release-4.4/manifests/12_prometheus_rule.yaml#L6-L8
Mainly to make sure I'm not adding any regression here. 

That being said, I don't think putting it on our to-do list to explore the significance of these labels is a bad idea.

PR #403, opened 2021-05-24
WIP: make leader election lock namespace configurable

bparees:


bparees:
we decided to just avoid this operator entirely in hypershift and we're talking about doing away w/ it in standalone OCP, so i guess we can close this.

PR #413, opened 2021-06-25
Start without defaults on ARM

yselkowitz:

Remove the defaultDir argument from the operator manifest on ARM.

﻿The defaults currently point to 4.8 indexes which do not yet exist on
ARM, which result in unnecessary alerts and the need for a post-install
command to silence them.  Removing the defaultDir argument from the
operator launch command on ARM should avoid this.

This should be reverted once 4.9 indexes are available.

**Reviewer Checklist**

### DIFF
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -10,6 +10,9 @@ USER marketplace-operator
 COPY --from=builder /go/src/github.com/operator-framework/operator-marketplace/build/_output/bin/marketplace-operator /usr/bin
 ADD manifests /manifests
 ADD defaults /defaults
+USER root
+RUN if test "$(arch)" = "aarch64"; then sed -i '/defaultsDir/d' /manifests/09_operator.yaml; fi
+USER marketplace-operator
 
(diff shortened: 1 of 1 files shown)


yselkowitz:
Please allow this to be tested, but let's not merge yet pending further consulation on this approach.
/hold

yselkowitz:
We decided against this approach, and will just doc note it with a post-install step to silence the alerts.

yselkowitz:
While we could (and depending on timing might still) doc note this for users, in order to get clean results in CI and not have these alerts mask the detection of other potential problems, it looks like we'll need this after all.  With the understanding that we'll need to revert this once the default operator indexes have moved to 4.9 and are available on ARM, please consider this.
/hold cancel

yselkowitz:
Probably a flake
/test e2e-aws

yselkowitz:
Flake
/test e2e-aws-upgrade

yselkowitz:
Upgrade failure is https://bugzilla.redhat.com/show_bug.cgi?id=1955300  4.9 is the initial release for ARM, so upgrade test is irrelevant, and failure is clearly unrelated to PR.  Could I get an override please?

PR #414, opened 2021-06-29
Bug 1958888: Use client-go's leader election implementation

timflannagan:
Update the cmd/manager/main.go package and use client-go's leader
election implementation instead of operator-sdk's implementation.

This type of leader election implementation is acquired through a
leasing mechanism vs. operator-sdk's leader for life, which can produce
a failed upgrade in some edge cases as the new operator version cannot
acquire leader election.

Alternatives:
- Using controller-runtime's implementation, but that
would likely include a sizable refactor in the driver function when
setting up the manager instance here, as that implementation only
attempts to gain leader election when mgr.Start() is called. This is
problematic in the case that we have several pre-requisites before
starting any controllers/informer caches/etc., like loading the default
CatalogSource YAML manifests into a cache.
- Avoid needing a leader election mechanism by spinning up a sidecar
container that recycles Pods when a new rollout is detected.

**Reviewer Checklist**


timflannagan:
AWS installations should be fixed now.

/retest

cmd/manager/main.go
```diff
@@ -105,96 +137,105 @@ func main() {
...
-	logrus.Info("The operator exited gracefully, exit code 0")
-	os.Exit(0)
+	id := os.Getenv("POD_NAME")
+	if id == "" {
+		logger.Info("failed to determine $POD_NAME falling back to hostname")
+		id, err = os.Hostname()
+		if err != nil {
+			logger.Fatal(err)
+		}
+	}
```
timflannagan, line 205: Just a quick note about this behavior - I was reading the operator-lib's leader election package [godoc](https://github.com/operator-framework/operator-lib/blob/main/leader/doc.go#L44-L52) and saw the following was documented:

```
Leader for Life requires that all candidate Pods be in the same Namespace. It
uses the downwards API to determine the pod name, as hostname is not reliable.
You should run it configured with:
env:
  - name: POD_NAME
    valueFrom:
      fieldRef:
        fieldPath: metadata.name
```

We already specify the POD_NAME in the marketplace-operator deployment, so we _should_ have access to that value, and if not, fall back to getting this value by querying for pod hostname. I was mainly testing this on a crc cluster (e.g. single node) and ran into some of those inconsistencies between os.Hostname() and pod metadata.Name values.

cmd/manager/main.go
```diff
@@ -47,52 +57,74 @@ func printVersion() {
...
+		version                 bool
 	)
-	flag.StringVar(&clusterOperatorName, "clusterOperatorName", "", "the name of the OpenShift ClusterOperator that should reflect this operator's status, or the empty string to disable ClusterOperator updates")
-	flag.StringVar(&defaults.Dir, "defaultsDir", "", "the directory where the default CatalogSources are stored")
+	flag.StringVar(&clusterOperatorName, "clusterOperatorName", "", "configures the name of the OpenShift ClusterOperator that should reflect this operator's status, or the empty string to disable ClusterOperator updates")
+	flag.StringVar(&defaults.Dir, "defaultsDir", "", "configures the directory where the default CatalogSources are stored")
 	flag.BoolVar(&version, "version", false, "displays marketplace source commit info.")
 	flag.StringVar(&tlsKeyPath, "tls-key", "", "Path to use for private key (requires tls-cert)")
 	flag.StringVar(&tlsCertPath, "tls-cert", "", "Path to use for certificate (requires tls-key)")
+	flag.StringVar(&leaderElectionNamespace, "leader-namespace", "openshift-marketplace", "configures the namespace that will contain the leader election lock")
```
timflannagan, line 90: Another quick note - this was namely added to accommodate the hypershift use case, where you may be housing locks outside of the namespace the actual operator deployments are being run.

cmd/manager/main.go
```diff
@@ -25,20 +26,29 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
 const (
-	leaderElectionConfigMapName = "marketplace-operator-lock"
+	// TODO(tflannag): Should this be configurable?
+	defaultLeaderElectionConfigMapName = "marketplace-operator-lock"
+	defaultRetryPeriod                 = 30 * time.Second
+	defaultRenewDeadline               = 60 * time.Second
+	defaultLeaseDuration               = 90 * time.Second
```
timflannagan, line 51: These values mirror the ones that were recently updated in the downstream package-server-manager in order to meet the SNO GA requirements regarding API server downtime: https://github.com/openshift/operator-framework-olm/blob/master/cmd/package-server-manager/util.go#L14-L22.

cmd/manager/main.go
```diff
@@ -105,96 +137,105 @@ func main() {
...
-	logrus.Info("Starting the Cmd.")
+	leaderCtx, leaderCancel := context.WithCancel(context.Background())
+	defer leaderCancel()
 
-	// Handle the defaults
-	err = ensureDefaults(cfg, mgr.GetScheme())
-	if err != nil {
-		logrus.Fatal(err)
-	}
+	run := func(ctx context.Context) {
```
timflannagan, line 158: The `OnStartedLeading` leader election callback function allows us to specify what should be run once the leader election lease has been acquired, so wrapping the controller/manager specific logic into an inline function, and then calling that inlined function in that callback function allows us to almost completely emulate the previous startup/initialization/etc. logic before migrating to a leaser-for-lease implementation.

There are likely cleaner implementations, like moving this to it's own dedicated function, but going down that path lead to a much larger refactoring of the main executable package and didn't want to inflate the number of changes in this PR.

timflannagan: One other things to note as the diff is pretty difficult to parse is that health and liveness probe endpoints are setup _before_ the `run` function is called during the `OnStartedLeading` callback function. It looks like we're blindly setting both of those probes to a 200 status anyways so the current behavior should be retained.

cmd/manager/main.go
```diff
@@ -25,20 +26,29 @@ import (
...
+	"k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
 const (
-	leaderElectionConfigMapName = "marketplace-operator-lock"
+	// TODO(tflannag): Should this be configurable?
```
njhale, line 47: I wouldn't since:

1. today there's no ask to change these between supported configurations of OpenShift; e.g. single-node, regular
2. in the long term, we plan for this controller to be subsumed by downstream OLM

cmd/manager/main.go
```diff
@@ -47,52 +57,74 @@ func printVersion() {
...
 
+func setupScheme() *kruntime.Scheme {
+	scheme := kruntime.NewScheme()
+
+	utilruntime.Must(apis.AddToScheme(scheme))
+	utilruntime.Must(olmv1alpha1.AddToScheme(scheme))
+	utilruntime.Must(v1beta1.AddToScheme(scheme))
+	utilruntime.Must(corev1.AddToScheme(scheme))
+
+	if configv1.IsAPIAvailable() {
```
njhale, line 68: Are the `configv1` schemes actually **required** for the controller to run, or are they just optional?

cmd/manager/main.go
```diff
@@ -105,100 +137,115 @@ func main() {
...
-	logrus.Info("Waiting to become leader.")
-	err = leader.Become(context.TODO(), leaderElectionConfigMapName)
-	if err != nil {
-		logrus.Fatal(err, "Failed to retry for leader lock")
-	}
-	logrus.Info("Elected leader.")
+	ctx := signals.Context()
+	stopCh := ctx.Done()
+
+	leaderCtx, leaderCancel := context.WithCancel(context.Background())
```
njhale, line 155: why not parent `leaderCtx` w/ the stop signal context: `leaderCtx, leaderCancel := context.WithCancel(ctx)`?

cmd/manager/main.go
```diff
@@ -105,100 +137,115 @@ func main() {
...
+		if clusterOperatorName != "" {
+			logger.Info("setting up the marketplace clusteroperator status reporter")
+			statusReporter, err = status.NewReporter(cfg, mgr, namespace, clusterOperatorName, os.Getenv("RELEASE_VERSION"), stopCh)
+			if err != nil {
+				logger.Fatal(err)
+			}
+		}
 
-	logrus.Info("Starting the Cmd.")
+		// Populate the global default OperatorSources definition and config
```
njhale, line 169: nit: `OperatorSources` have been deprecated (and removed?)

cmd/manager/main.go
```diff
@@ -105,100 +137,115 @@ func main() {
...
+		ReleaseOnCancel: true,
+		LeaseDuration:   defaultLeaseDuration,
+		RenewDeadline:   defaultRenewDeadline,
+		RetryPeriod:     defaultRetryPeriod,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(ctx context.Context) {
+				logger.Infof("became leader: %s", id)
+				run(leaderCtx)
+			},
+			OnStoppedLeading: func() {
```
njhale, line 233: @timflannagan could you explain what it means for this controller to give up leadership? Should it stop managing CatalogSources?

If so, I'm having trouble finding where that's happening.

njhale:
Do we need at least one upgrade edge that understands both the old leader election method and the new one?

or is there a reason why we don't need to be worried about an older version not relinquishing leadership to something that doesn't understand its election method?

cmd/manager/main.go
timflannagan: This is a pre-existing comment, and a diff is only produced due to me inlining some of the code. I don't think there's a ton of value in the comment itself, outside of whether operatorsources have been deprecated/removed, which appears to be the case, so I'm fine with removing that anyways.

cmd/manager/main.go
timflannagan: I'm not entirely sure tbh - for the operatorhub controller they're absolutely necessary as the OperatorHub Go types are housed in [o/api](https://github.com/openshift/api/blob/master/config/v1/types_operatorhub.go). I'd like to hold off on modifying this behavior as much as possible as we'll need to backport these fixes.

cmd/manager/main.go
timflannagan: Both the old and new leader election implementations are centered around a ConfigMap locking mechanism. In the previous implementation, which used operator-sdk's leader-for-life mechanism, was stalling upgrades entirely when the node holding the lock being unreachable, and the result is that a new marketplace-operator pod is unable to preempt the previous pod as the ConfigMap leader election lease cannot be garbage collected.

Now, preemption is handled using the leader-for-lease implementation, where a pod/node will obtain a lock, become a leader, and periodically will attempt to renew this lease. In the case of multiple pods all attempting to obtain that lease and become a leader, one of those pods will obtain leadership, and in the next preemption poll, the current leader will need to attempt to re-acquire leadership and if that process is unsuccessful, another will be elected leader and lock ownership will transition to that pod's metadata.Name (i.e. their identity).

njhale (approved):
/lgtm

PR #419, opened 2021-08-20
Bug 1997811: Bump k8s and controller-runtime dependencies to v0.21.x and v0.9.x

timflannagan:
# Overview

The marketplace operator is currently building and testing with an older Kubernetes version (v0.20.6) of clients and APIs. This can pose a maintenance burden going forward as we'll be deploying the marketplace operator using an unmaintained Kubernetes minor version as we progress future in the 4.x OCP lifecycle, and upgrading to OCP 4.9 will be needed for upgrading from EUS to EUS. Simply upgrading the controller runtime and k8s.io dependencies to a more recent (v0.21.x, v0.22.x) versions and making the requisite code changes is not possible as we're parking on a stale v0.19.x operator-sdk version. Currently, the marketplace operator uses the operator-sdk testing harness, which was deprecated in v0.19.x and removed entirely in v1.x in favor of controller runtime's envtest implementation. The operator-sdk testing harness is not compatible with client function signatures changes in controller-runtimes v0.7.x release, and bumping the operator-sdk dependency to a supported v1.x version involves several breaking changes. This PR is centered around the removal of the operator-sdk dependency entirely, replacing any direct package imports, and removing the testing harness entirely.

## Minor Changes

- Various unused packages have been removed.
- The root list of CVO manifests have been regenerated.
- Avoid disabling module support (`GO111MODULE=off`) when building the marketplace operator binary in the build/build.sh bash script.
- Move the scripts/update-manifests.sh bash script to hack.
- Avoid registering the v1beta1 CRD scheme in the main executable package.

### Avoid manually vendoring the OLM API types

<https://github.com/operator-framework/operator-marketplace/pull/329> introduced these manually vendored APIs as the operator-sdk version at the time didn't support Go modules.

This removes those vendored packages in favor of the operator-framework/api dependency, which currently houses the CatalogSource type definitions the marketplace operator needs.

Any existing package imports in the code base are replaced to use the v0.10.5 api dependency instead of the (now) deleted manually vendored API packages.

### Handle controller runtime v0.7.x breaking changes

The [controller-runtime v0.7.x release](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.7.0) introduced a couple of breaking changes:

- `mgr.Start(...)` now takes a context.Context as a parameter instead of a stop channel.
- Any `Reconcile` method now takes a context.Context parameter.
- Any function or methods that passes around a `runtime.Object` needs to be updated to instead handle a `client.Object` and `client.ObjectList` instead.

### Removing the operator-sdk dependency

Any code paths that directly import the github.com/operator-framework/operator-sdk package import have been removed.

The `k8sutil.GetWatchNamespace()` has been directly vendored in the pkg/apis/operators/shared package.

Avoid printing the operator-sdk version when starting up the marketplace operator executable.

Migrate the e2e testing packages to use Ginkgo instead of the operator-sdk testing harness. The majority of the OLM-related repositories (and the Kubernetes ecosystem as a whole) use Ginkgo for e2e testing. Any relevant marketplace e2e test has been migrated to a testing package in the test/e2e package.

Update the scripts/ directory and remove any operator-sdk-related bash scripts that are no longer needed. The root Makefile has been also updated to remove targets that are no longer needed. The `e2e-job` Makefile target, which is what the o/release ci-operator configuration is responsible for running, has been updated to simply inline the `go test ... ./test/e2e/...` command instead of utilizing a wrapper bash script to setup and run e2e tests.


pkg/controller/catalogsource/catalogsource_controller.go
```diff
@@ -87,12 +89,12 @@ type ReconcileCatalogSource struct {
 	client client.Client
 }
 
-func (r *ReconcileCatalogSource) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+func (r *ReconcileCatalogSource) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
```
joelanford, line 92: Is it necessary/helpful/reasonable to pass this context down to anything called from this function (e.g. `GetSingleton().Get()` or `Ensure()`)?

pkg/defaults/catsrcHelpers.go
```diff
@@ -51,9 +51,9 @@ func getCatsrcDefinition(fileName string) (*olm.CatalogSource, error) {
 
 // processCatsrc will ensure that the given CatalogSource is present or not on
 // the cluster based on the disable flag.
-func processCatsrc(client wrapper.Client, def olm.CatalogSource, disable bool) error {
+func processCatsrc(client wrapper.Client, def olmv1alpha1.CatalogSource, disable bool) error {
 	// Get CatalogSource on the cluster
-	cluster := &olm.CatalogSource{}
+	cluster := &olmv1alpha1.CatalogSource{}
 	err := client.Get(context.TODO(), wrapper.ObjectKey{
```
joelanford, line 57: Pass a `context.Context` into `processCatsrc`?

joelanford (commented):
/lgtm

Only nit would be to see if we could further reduce or completely remove usage of `context.TODO()`. Non-blocking though.

pkg/controller/catalogsource/catalogsource_controller.go
timflannagan: I did a cursory look through the repository and removed as many usages of `context.TODO()` outside of testing code that were easily replaceable. The status reported package, which is responsible for setting up a go routine to monitor the marketplace ClusterOperator resource (and as aside, I think we'd want to move that handlement to its own controller down the line) but the changeset for updating all the call sites to properly handle signals seemed riskier in the full context of these changes, so I held off there.

One other quick note, and something we can tackle outside of this PR, is we still embed OperatorSource logic in the various top-level functions/methods/etc. I wonder if it's worth the easier maintenance for 4.9+ to make the requisite changes. It likely would be a source for cherrypick failures when backporting bug fixes, but the surface area would be low. I'm not super worried about breaking other clients that depend on marketplace at this point either.

pkg/controller/configmap/configmap_controller.go
```diff
@@ -97,34 +96,27 @@ type ReconcileConfigMap struct {
...
-			// Return and don't requeue
-			return reconcile.Result{}, nil
-		}
-		// Error reading the object - requeue the request.
-		return reconcile.Result{}, err
+	if err := r.client.Get(ctx, request.NamespacedName, caConfigMap); err != nil {
+		// Requested object was not found, could have been deleted after reconcile request.
+		// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+		// Return and don't requeue
+		return reconcile.Result{}, client.IgnoreNotFound(err)
```
joelanford, line 116: :+1:

joelanford (approved):
/lgtm

PR #423, opened 2021-09-08
pkg/controller: Remove the CA ConfigMap controller

timflannagan:
Remove the controller responsible for ensuring that the ConfigMap
containing the custom CA and system certificates in proxy-enabled
environments. This controller was responsible for restarting the
marketplace operator deployment when the desired specification (the
ConfigMap merged CA bundle contents) don't match what's current listed
on disk in the underlying Pod resource.

The CVO is responsible for deploying an empty ConfigMap that contains
the "config.openshift.io/inject-trusted-cabundle" label key (and a value
of "true") and the CNO component is responsible for injecting the
user-provided/system CA certificates into a single merged CA bundle.

Updates made to ConfigMap resources are updated automatically, and the
projected content is eventually updated when kubelet performs it's next
periodic sync [1].

[1] https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically

**Reviewer Checklist**


timflannagan:
This is likely contenious.

/hold

PR #424, opened 2021-09-08
Remove unused packages

timflannagan:
Remove the pkg/builder packages as they're no longer called anywhere in
the codebase since the removal of the OperatorSource and
CatalogSourceConfig API types.

Remove the unused pkg/metrics/roundtripper.go package as we no longer
need to create custom metrics around the previous app-registry
implementation of serving catalog content.

**Reviewer Checklist**


njhale:
Trimming this down is the next best thing to deprecating it.

/approve

PR #456, opened 2021-11-30
Bug 2034409: Update defaults to point to 4.10 images

kevinrizza:


kevinrizza:
/hold

looks like this test is permafailing

njhale:
Looks like `e2e-aws-console-olm` is permafailing due to a hardcoded dependency on [a member of the redhat-operators catalog](https://github.com/openshift/console/blob/f240986d7751abcf3292ed6522b3b78c5c87a5c5/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/create-namespace.spec.ts#L26) that is not currently present in the 4.10 version of the index. 

This will continue to fail until the test is updated to either:
- reference an operator that's present in the index
- not depend on the default indexes at all (preferred)

njhale:
I opened a BZ against the console to address this: https://bugzilla.redhat.com/show_bug.cgi?id=2035409

If we run short on time before the release we can force merge then disable this job in `openshift/release` until the test responsible is fixed.

jhadvig:
> Looks like e2e-aws-console-olm is permafailing due to a hardcoded dependency on a member of the redhat-operators catalog that is not currently present in the 4.10 version of the index.

@njhale could you please recommend an operator selector that we could use instead? The failing suite is for creating a nemaspace from operator install page. So I *think* it should be pretty easy to replace the selector for one that is present in the 4.10

ccing @dtaylor113

njhale:
> So I think it should be pretty easy to replace the selector for one that is present in the 4.10

@jhadvig I **strongly** advise against doing so, if we can help it, since there's **no guarantee** that the operator will be present in the catalog shipped with 4.11, or 4.12, or any future minor release of OpenShift. I think we should only pursue this route if removing the dependency isn't possible due to time constraints.

To remove the dependency will involve creating a test fixture CatalogSource containing a known set of operators. The easiest way to create this set would be to push the 4.9 catalog image to a new image repo. You can then reference that image in the text fixture CatalogSource.

jhadvig:
ccing @harishsurf

dtaylor113:
Hi, unfortunately we currently have to depend on real catalog and operators.  We currently are using 2 or 3 operators in this fashion until we can create our own CatalogSource & e2e Operator dedicated to be used in e2e testing.

In `create-namespace.spec.ts` you can change from :
```
-    const operatorSelector = 'advanced-cluster-management-redhat-operators-openshift-marketplace';
```
... to ...
```
+    const operatorSelector = 'businessautomation-operator-redhat-operators-openshift-marketplace';
```
and also delete the line
`- cy.byTestID('Select a Namespace-radio-input').check();`

This worked locally for me.

perdasilva:
@kevinrizza maybe we need to rebase this to pick up the test changes?

anik120:
/hold cancel 

Looks like all the tests have passed 🎉

PR #457, opened 2022-01-06
Bug 1927478: Add catalog source pod spec overrides

perdasilva:
Signed-off-by: perdasilva <perdasilva@redhat.com>

Updates the catalog source definitions to use the new `grpcPodConfig` to override tolerations, node-selector and priorityClassName.

https://bugzilla.redhat.com/show_bug.cgi?id=1927478

**Reviewer Checklist**


awgreene (changes requested):
I am ready to `lgtm` this, but I would like to clarify one thing before I do so.

We have added these tolerations to the default catalogSource manifests, which users will not be able to override, is this expected behavior?

awgreene:
> I am ready to `lgtm` this, but I would like to clarify one thing before I do so.
> 
> We have added these tolerations to the default catalogSource manifests, which users will not be able to override, is this expected behavior?

I suppose a user could disable these catalogs and then recreate them with their own tolerations.

anik120:
/hold cancel 

/lgtm 
looks all good now :) 

> I suppose a user could disable these catalogs and then recreate them with their own tolerations.

I can confirm that this is true.

PR #462, opened 2022-03-24
chore(manifests): add openshift capability annotation

njhale:

Register CVO-managed manifests under the optional `marketplace` capability.

See also: 
- [Example w/ cluster-baremetal-operator](https://github.com/openshift/cluster-baremetal-operator/pull/249/files#diff-fd1ca771b89b4ffca6e3473d81aa894d2ba00f3e24a987a814a1ca41426ac1b9)
- [OpenShift "capabilities" EP](https://github.com/openshift/enhancements/blob/master/enhancements/installer/component-selection.md)

We want OpenShift users to be able to disable/enable the marketplace operator without putting their cluster into an unmanaged state.

Jira: [OLM-2474](https://issues.redhat.com/browse/OLM-2474)

### DIFF
--- a/manifests/01_namespace.yaml
+++ b/manifests/01_namespace.yaml
@@ -7,6 +7,7 @@ metadata:
     include.release.openshift.io/single-node-developer: "true"
     openshift.io/node-selector: ""
     workload.openshift.io/allowed: "management"
+    capability.openshift.io/name: "marketplace"
   labels:
     openshift.io/cluster-monitoring: "true"
   name: "openshift-marketplace"
--- a/manifests/04_service_account.yaml
+++ b/manifests/04_service_account.yaml
@@ -7,3 +7,4 @@ metadata:
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
+    capability.openshift.io/name: "marketplace"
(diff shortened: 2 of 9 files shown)


njhale:
cc @bparees

perdasilva:
@njhale do we want to add some documentation to readme as well to say where this annotation and value come from?

timflannagan:
> where is the OperatorHub CRD itself defined? i'd expect it to be excluded when this cap is not enabled, also.

It looks like is defined in o/api: https://github.com/openshift/api/blob/master/config/v1/0000_03_config-operator_01_operatorhub.crd.yaml

bparees:
ok so https://github.com/openshift/api/pull/1150 should be marking it as a capability unless there is another reason to have that api defined when the marketplace operator is disabled

(it'll be interesting to see what happens to the console when marketplace is disabled)

njhale:
> ok so [openshift/api#1150](https://github.com/openshift/api/pull/1150) should be marking it as a capability unless there is another reason to have that api defined when the marketplace operator is disabled
> 
> (it'll be interesting to see what happens to the console when marketplace is disabled)

@bparees just pushed that change actually: https://github.com/openshift/api/pull/1150/files#diff-c53669fc670c26b116439080f132967cca87dec5864c511ed7772de465e0bf7fR9

njhale:
@timflannagan are these prow-job failures due to the same resource starvation on AWS we're seeing in OLM?

timflannagan:
It seems like our AWS dev account should be slightly more stable now.

/retest

timflannagan:
The o/api PR just landed - we should just be waiting on this marketplace PR to land now, but we could use clusterbot to build a custom release image that we pass to the openshift installer to verify these changes work out-of-band to this PR merging?

timflannagan:
https://github.com/openshift/release/pull/27304 moves these prowjobs to GCP.

PR #468, opened 2022-04-13
fix(manifests): add singleton operatorhub cr

njhale:
- Relocate the singleton OperatorHub CR instance from openshift/cluster-config-operator
- add the marketplace capability annotatation to CR instance

This consolidates ownership of OperatorHub-related manifests to the marketplace-operator
and will help prevent issues with resource application order.

Signed-off-by: Nick Hale <njohnhale@gmail.com>

### DIFF
--- a/manifests/0000_03_marketplace-operator_02_operatorhub.cr.yaml
+++ b/manifests/0000_03_marketplace-operator_02_operatorhub.cr.yaml
@@ -1,0 +1,13 @@
+# This name of this file must appear after that of the file containing the OperatorHub CRD when ordered lexicographically.
+# The OperatorHub CRD file is sourced from the following vendor directory: ./vendor/github.com/openshift/api/config/v1
+apiVersion: config.openshift.io/v1
+kind: OperatorHub
+metadata:
+  name: cluster
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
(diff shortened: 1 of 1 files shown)


manifests/0000_03_operatorhub.cr.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: config.openshift.io/v1
```
wking, line 1: We want a filename that sorts after `0000_03_config-operator_01_operatorhub.crd.yaml` (which is what's currently vendored) and `0000_03_marketplace-operator_01_operatorhub.crd.yaml` (where openshift/api#1178 is taking the CRD).  Maybe `0000_03_marketplace-operator_02_operatorhub.cr.yaml`?

manifests/0000_03_marketplace-operator_02_operatorhub.cr.yaml
```diff
@@ -0,0 +1,13 @@
+# This name of this file must appear after that of the file containing the OperatorHub CRD when ordered lexicographically.
+# The OperatorHub CRD file is sourced from the following vendor directory: ./vendor/github.com/openshift/api/config/v1
+apiVersion: config.openshift.io/v1
+kind: OperatorHub
+metadata:
+  name: cluster
```
njhale, line 6: I confirmed that CVO won't have a problem with the (temporary) duplicate in openshift/cluster-config-operator. The PR to remove that is also queued up for after this merges: openshift/cluster-config-operator#245

PR #469, opened 2022-04-13
Bug 2010375: Clarify reason/steps to diagnose in *CatalogError prom alert

anik120:

**Reviewer Checklist**

### DIFF
--- a/manifests/12_prometheus_rule.yaml
+++ b/manifests/12_prometheus_rule.yaml
@@ -13,47 +13,13 @@ metadata:
     role: alert-rules
 spec:
   groups:
-    - name: marketplace.community_operators.rules
+    - name: operator.marketplace.rules
       rules:
-        - alert: CommunityOperatorsCatalogError
+        - alert: OperatorHubSourceError
           annotations:
-            summary: Community-operators not ready for 10 minutes
-            description: Fires whenever the community-operators source is not ready for more than 10 mins.
-            message: Default OperatorHub source "community-operators" is in Non-Ready state for more than 10 mins.
-          expr: catalogsource_ready{name="community-operators",exported_namespace="openshift-marketplace"} == 0
-          for: 10m
-          labels:
-            severity: warning
-    - name: marketplace.certified_operators.rules
-      rules:
-        - alert: CertifiedOperatorsCatalogError
-          annotations:
-            summary: Certified-operators not ready for more than 10 minutes
-            description: Fires whenever the certified-operators source is not ready for more than 10 mins.
-            message: Default OperatorHub source "certified-operators" is in Non-Ready state for more than 10 mins.
-          expr: catalogsource_ready{name="certified-operators",exported_namespace="openshift-marketplace"} == 0
-          for: 10m
-          labels:
-            severity: warning
-    - name: marketplace.redhat_operators.rules
-      rules:
-        - alert: RedhatOperatorsCatalogError
-          annotations:
-            summary: Redhat-operators not ready for more than 10 minutes
-            description: Fires whenever the redhat-operators source is not ready for more than 10 mins.
-            message: Default OperatorHub source "redhat-operators" is in Non-Ready state for more than 10 mins.
-          expr: catalogsource_ready{name="redhat-operators",exported_namespace="openshift-marketplace"} == 0
-          for: 10m
-          labels:
-            severity: warning
-    - name: marketplace.redhat_marketplace.rules
-      rules:
-        - alert: RedhatMarketplaceCatalogError
-          annotations:
-            summary: Redhat-marketplace not ready for more than 10 minutes
-            description: Fires whenever the redhat-marketplace source is not ready for more than 10 mins.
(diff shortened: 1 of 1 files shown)


manifests/12_prometheus_rule.yaml
```diff
@@ -17,9 +17,8 @@ spec:
       rules:
         - alert: CommunityOperatorsCatalogError
           annotations:
-            summary: Community-operators not ready for 10 minutes
-            description: Fires whenever the community-operators source is not ready for more than 10 mins.
-            message: Default OperatorHub source "community-operators" is in Non-Ready state for more than 10 mins.
+            summary: Community-operators source is in non-ready for more than 10 minutes.
+            description: Operators shipped via this source are not available to install until the issue is fixed. Inspect the status of the pod owned by community-operators source in the openshift-marketplace namespace (oc get pods -n openshift-marketplace) to diagnose and repair. {{ "{{ with $console_url := \"console_url\" | query }}{{ if ne (len (label \"url\" (first $console_url ) ) ) 0}} For more information refer to {{ label \"url\" (first $console_url ) }}/k8s/ns/openshift-marketplace/pods.{{ end }}{{ end }}" }}
```
wking, line 21: Worth additional impact clarification like "There is no effect on already installed operators"?  Or maybe "Installed operators dependent on this source will not receive updates until the issue is fixed"?  Or whatever the impact on installed operators actually is.

wking: Looks like you could use `oc -n openshift-marketplace get pods -l olm.catalogSource=community-operators` to trim the results down to just this pod, and similarly for the other alerts:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.10-e2e-aws-serial/1514248134916575232/artifacts/e2e-aws-serial/gather-extra/artifacts/pods.json | jq -r '.items[].metadata | select(.namespace == "openshift-marketplace") | .name + " " + .labels["olm.catalogSource"]'
certified-operators-q6cdz certified-operators
community-operators-8lrw9 community-operators
marketplace-operator-7c5bf65dd6-wptp9 
redhat-marketplace-vs6rr redhat-marketplace
redhat-operators-nc492 redhat-operators
```

manifests/12_prometheus_rule.yaml
```diff
@@ -13,47 +13,13 @@ metadata:
...
-            severity: warning
-    - name: marketplace.redhat_marketplace.rules
-      rules:
-        - alert: RedhatMarketplaceCatalogError
-          annotations:
-            summary: Redhat-marketplace not ready for more than 10 minutes
-            description: Fires whenever the redhat-marketplace source is not ready for more than 10 mins.
-            message: Default OperatorHub source "redhat-marketplace" is in Non-Ready state for more than 10 mins.
-          expr: catalogsource_ready{name="redhat-marketplace",exported_namespace="openshift-marketplace"} == 0
+            summary: The {{ $labels.name }} source is in non-ready state for more than 10 minutes.
```
wking, line 20: From [the `summary` docs][1]:

> Use a static string or, if dynamic expansion is needed, aim for expanding into the same string for alerts that are typically grouped together into one notification. In that way, it can be used as a common “headline” for all alerts in the notification template.

So I'd probably leave `{{ $labels.name }}`.  But I haven't broken multiple catalogs and then gone to look at the UX; maybe it's fine with the expanded name.

[1]: https://github.com/monitoring-mixins/docs/pull/10/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R186-R189

manifests/12_prometheus_rule.yaml
```diff
@@ -13,47 +13,13 @@ metadata:
...
+            description: Operators shipped via this source are not available for installation until the issue is fixed. Operators already installed from this source will not receive updates until issue is fixed. Inspect the status of the pod owned by {{ $labels.name }} source in the openshift-marketplace namespace (oc -n openshift-marketplace get pods -l olm.catalogSource={{ $labels.name }}) to diagnose and repair
```
wking, line 21: I'd change `this source` to `{{ $labels.name }}` or `the {{ $labels.name }} source` or some such here.

manifests/12_prometheus_rule.yaml
wking: Is there often a need to distinguish between multiple instances of this alert?  I guess that depends on how often catalogs break from source-specific reasons vs. how often they break for all-source reasons?

anik120: @wking I don't believe we have any instances where there are all-source reasons, unless for eg all the sources use images in a registry that's currently unreachable, or there's a bug that breaks all the CatalogSources. i.e there aren't any instance where all sources are down because of a single source, if that's what you were asking. So single source breaking is much more common place.

dinhxuanvu (commented):
/lgtm

PR #472, opened 2022-05-02
chore(modules): bump openshift/api for new operatorhub crd

njhale:

- Drop unnecessary replaces directives
- Bump openshift/api to include the new OperatorHub CRD manifest (w/
  marketplace capability annotation)
- Update ./vendor

The current version of OpenShift API being vendored contains an old v1beta1 CRD manifest, missing the marketplace capability annotation.

Blocks openshift/cluster-config-operator#245


tylerslaton (approved):
Wow, those dependencies were over a year old. Do we know if there will be any unintended side effects from this? Should probably keep a close eye on the E2E suite just in case. Regardless, this looks good to me. 

/lgtm

njhale:
> Do we know if there will be any unintended side effects from this?

There could be. Hopefully our e2e suite will catch this. Already seeing okd-e2e-gcp failing, but I'm not confident it's related.

I'll be on the lookout for subsequent failures.

/retest-required

PR #478, opened 2022-06-23
Bug 2088541: Add psa anotations to namespace to suppress warnings

perdasilva:
Signed-off-by: perdasilva <perdasilva@redhat.com>

Adds psa annotations for the openshift-marketplace namespace to suppress PSA warnings in 4.11

**Reviewer Checklist**

### DIFF
--- a/manifests/01_namespace.yaml
+++ b/manifests/01_namespace.yaml
@@ -10,4 +10,6 @@ metadata:
     capability.openshift.io/name: "marketplace"
   labels:
     openshift.io/cluster-monitoring: "true"
+    pod-security.kubernetes.io/audit: baseline
+    pod-security.kubernetes.io/warn: baseline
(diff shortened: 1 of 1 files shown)


perdasilva:
> Interesting...
> 
> ```
> Jun 23 16:35:52.000 W ns/openshift-marketplace pod/certified-operators-dchw9 node/ci-op-stnpvw2q-c2bac-2zf6f-master-2 reason/Unhealthy Startup probe failed: command timed out
> ```

That is weird @dinhxuanvu any ideas what could be the issue? Maybe networking wasn't setup properly on the cluster deployment?

timflannagan:
Looking at that (still running) e2e-gpc-serial test, it appears the test was successful so maybe we're in the clear here? We're still waiting for the test logs to get populated, which is where I found that warning log message initially.

PR #480, opened 2022-07-26
Add PSA enforcement label

kevinrizza:
Add the label `pod-security.kubernetes.io/enforce: baseline` to the openshift manifests so that PSA enforcement doesn't cause catalog pods to fail to be scheduled.


anik120 (approved):
/lgtm

Looks good to unblock tests temporarily, till https://github.com/operator-framework/operator-lifecycle-manager/pull/2820 is in the works.

PR #484, opened 2022-08-18
(psa) make marketplace namespace restricted

anik120:

Pod Security Admission labels PR.
enforce restricted.
enforce-version: v1.24

Following instructions to set PSA enforce level for all `openshift-*` namespaces. 

**Reviewer Checklist**

### DIFF
--- a/manifests/01_namespace.yaml
+++ b/manifests/01_namespace.yaml
@@ -10,4 +10,6 @@ metadata:
     capability.openshift.io/name: "marketplace"
   labels:
     openshift.io/cluster-monitoring: "true"
+    pod-security.kubernetes.io/enforce: restricted
+    pod-security.kubernetes.io/enforce-version: v1.24
   name: "openshift-marketplace"


manifests/01_namespace.yaml
```diff
@@ -10,4 +10,6 @@ metadata:
     capability.openshift.io/name: "marketplace"
   labels:
     openshift.io/cluster-monitoring: "true"
+    pod-security.kubernetes.io/enforce: restricted
+    pod-security.kubernetes.io/enforce-version: v1.24
```
camilamacedo86, line 14: /hold 

We just label it to enforce baseline because the pod requires more permissions.
We do not need to enforce it as restricted because it is done by default

bparees: all pods running in this NS are running as restricted-v2, so setting PSA enforcement to restricted should be fine.

```
$ oc project openshift-marketplace
Now using project "openshift-marketplace" on server "https://api.bparees.devcluster.openshift.com:6443".

$ oc get pods -o yaml | grep scc
      openshift.io/scc: restricted-v2
      openshift.io/scc: restricted-v2
      openshift.io/scc: restricted-v2
      openshift.io/scc: restricted-v2
      openshift.io/scc: restricted-v2
```

camilamacedo86: So, we get these changes merged. 👍 
/hold cancel

camilamacedo86: /lgtm

anik120: /hold cancel

anik120:
/hold 
I just realized with all the deprecation that happened in this repository, we might not have enough e2e tests to confirm that the marketplace deployment is up and running. I'll come back Monday and confirm workloads in this namespace is PSA restricted compliant.

anik120:
/hold cancel 

Looks like we do have enough tests, and the metadata I was worried about [was already present](https://github.com/operator-framework/operator-marketplace/blob/master/manifests/09_operator.yaml#L23-L26)

PR #491, opened 2022-11-30
OCPBUGS-4758: Unenforce PSA Restrictions

awgreene:
OpenShift v4.12 had previously enforced PSA restrictions by defaults in and namespace prefixed with `openshift-`. This change has been delayed until OpenShift v4.13 and users should be allowed to run catalogs without restricted permissions in the openshift-marketplace namespace.

### DIFF
--- a/defaults/01_redhat_operators.cr.yaml
+++ b/defaults/01_redhat_operators.cr.yaml
@@ -11,6 +11,8 @@ spec:
   displayName: "Red Hat Operators"
   publisher: "Red Hat"
   priority: -100
+  grpcPodConfig:
+    securityContextConfig: restricted
   updateStrategy:
     registryPoll:
       interval: 10m
--- a/manifests/01_namespace.yaml
+++ b/manifests/01_namespace.yaml
@@ -10,6 +10,8 @@ metadata:
     capability.openshift.io/name: "marketplace"
   labels:
     openshift.io/cluster-monitoring: "true"
-    pod-security.kubernetes.io/enforce: restricted
-    pod-security.kubernetes.io/enforce-version: v1.24
+    pod-security.kubernetes.io/enforce: baseline
+    pod-security.kubernetes.io/enforce-version: v1.25
+    pod-security.kubernetes.io/audit: restricted
+    pod-security.kubernetes.io/warn: restricted
   name: "openshift-marketplace"
--- a/defaults/02_certified_operators.yaml
+++ b/defaults/02_certified_operators.yaml
@@ -11,6 +11,8 @@ spec:
   displayName: "Certified Operators"
   publisher: "Red Hat"
   priority: -200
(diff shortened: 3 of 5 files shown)


bparees:
lgtm

manifests/01_namespace.yaml
```diff
@@ -10,6 +10,6 @@ metadata:
     capability.openshift.io/name: "marketplace"
   labels:
     openshift.io/cluster-monitoring: "true"
-    pod-security.kubernetes.io/enforce: restricted
+    pod-security.kubernetes.io/enforce: privileged
```
joelanford, line 13: Have we checked to see if this could be `baseline` instead of `privileged` and still support `legacy` catalog sources?

```suggestion
    pod-security.kubernetes.io/enforce: baseline
```

joelanford: Also, let's add the following?
```
pod-security.kubernetes.io/audit: restricted
pod-security.kubernetes.io/warn: restricted
```

manifests/01_namespace.yaml
```diff
@@ -10,6 +10,6 @@ metadata:
...
     pod-security.kubernetes.io/enforce-version: v1.24
```
joelanford, line 14: Do we need to bump the `enforce-version` to 1.25?

anik120:
+1 to all of Joe's comments otherwise lgtm

bparees:
/override ci/prow/e2e-gcp-serial

see discussion in:
https://coreos.slack.com/archives/C01CQA76KMX/p1670604148848869?thread_ts=1670594104.317349&cid=C01CQA76KMX

not all failures are due to that GCP quota issue, but none of them are related to this change and until the gcp quota issue goes away, we're pretty stuck and more runs won't help.

awgreene:
Following up on Ben's recommendation to skip this test until GCP is fixed.
/override ci/prow/e2e-gcp-serial

defaults/01_redhat_operators.cr.yaml
```diff
@@ -11,6 +11,8 @@ spec:
   displayName: "Red Hat Operators"
   publisher: "Red Hat"
   priority: -100
+  grpcPodConfig:
+    securityContextConfig: legacy
```
bparees, line 15: why are we setting this to legacy, and not restricted?  These catalogs should be able to run restricted, no?

awgreene: Yikes, yeah they should be set to restricted, great catch.

awgreene:
Reapplying the override.
/override ci/prow/e2e-gcp-serial

awgreene:
The e2e-gcp-serial tests failed due to a slight disruption of services, which has been seen in:
- Other PRs
- Doc change only PRs

/override ci/prow/e2e-gcp-serial

PR #504, opened 2023-01-18
OCPBUGS-5466: Default CatalogSource aren't always reverted to default settings

dtfranz:
**Problem:**
The `defaults` pkg for CatalogSources contains a helper func to check equality of two CatalogSources. It makes case-insensitive comparisons of a few fields and checks some others as well, but it doesn't perform a full diff. This results in some changes to the spec not being detected, resulting in the default fields not being maintained if changed by a user.

**Solution:**
Perform a full `reflect.DeepEqual` comparison of the two CatalogSourceSpecs, while still maintaining backwards-compatibility for the case-insensitive fields by `DeepCopy`ing both specs and setting those fields to lower-case.

OCPBUGS-5466

### DIFF
--- a/pkg/defaults/catsrcHelpers.go
+++ b/pkg/defaults/catsrcHelpers.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


awgreene (approved):
/approve

PR #520, opened 2023-05-24
OCPBUGS-9357: remove a race condition

everettraven:
No longer create the default `CatalogSource`s before the controller manager has actually started and reconciled an `OperatorHub` CR. This makes the `ensureDefaults()` function unused and therefore I have removed it.

- What currently happens is that the default `CatalogSource`s are created even if the `OperatorHub` marks them as disabled and then when the `OperatorHub` CR is reconciled they are all removed. This presents a problem in disconnected clusters where the defaults are disabled and a `CatalogSource` with the same name as a default is already present on cluster. In this scenario, the operator-marketplace will take ownership of the same-named `CatalogSource` and remove it when the `OperatorHub` CR is reconciled. Making sure the `OperatorHub` CR is reconciled before anything happens prevents this from happening.

**Reviewer Checklist**

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -21,7 +21,6 @@ import (
 	"github.com/operator-framework/operator-marketplace/pkg/controller/options"
 	"github.com/operator-framework/operator-marketplace/pkg/defaults"
 	"github.com/operator-framework/operator-marketplace/pkg/metrics"
-	"github.com/operator-framework/operator-marketplace/pkg/operatorhub"
 	"github.com/operator-framework/operator-marketplace/pkg/signals"
 	"github.com/operator-framework/operator-marketplace/pkg/status"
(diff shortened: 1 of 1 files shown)


anik120:
> /lgtm
> Catalog source operators and OLM operators are working properly.
> oc exec olm-operator-5877fdd8b7-6qj5b -n openshift-operator-lifecycle-manager -- olm --version
> OLM version: 0.0.0-ce46f5b97e4838c4508f6dc471cabc1199a8f0b9
> git commit: ce46f5b97e4838c4508f6dc471cabc1199a8f0b9

fyi @bandrade the proper test for this would be to see if with a build of operator-marketplace, when one of the default CatalogSources are deleted, if it is recreated or not. If it gets recreated, test is green. 
Additionally, another test would be to create an OperatorHub CR that disables one of the default catalogsources, and if the catalogsource is deleted and stays deleted after ~5 mins, then the test would be successful.

PR #521, opened 2023-05-24
Draft: OCPBUGS-10238: CatalogSource Default Enforcement Revert

dtfranz:
Reverts the new behavior of default catalogsource enforcement so that instead of strictly enforcing the spec of default catalogsources, we instead add a condition which indicates a non-upgradeable status.

### DIFF
--- a/pkg/defaults/catsrcHelpers.go
+++ b/pkg/defaults/catsrcHelpers.go
@@ -12,6 +12,7 @@ import (
 	wrapper "github.com/operator-framework/operator-marketplace/pkg/client"
 	"github.com/sirupsen/logrus"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/yaml"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/defaults/catsrcHelpers.go
```diff
@@ -181,5 +206,21 @@ func AreCatsrcSpecsEqual(spec1 *olmv1alpha1.CatalogSourceSpec, spec2 *olmv1alpha
 	spec1Copy.Image = strings.ToLower(spec1Copy.Image)
 	spec2Copy.Image = strings.ToLower(spec2Copy.Image)
 
-	return reflect.DeepEqual(spec1Copy, spec2Copy)
+	deepComparison := reflect.DeepEqual(spec1Copy, spec2Copy)
+
+	if !strings.EqualFold(string(spec1.SourceType), string(spec2.SourceType)) ||
```
dinhxuanvu, line 211: According to the bug, the main issue is changes on `spec.updateStrategy`. So perhaps you can simply reset that field on both objects and continue to use `DeepEqual`. Having said that, comparing only the fields you want is prudent but it is ugly code-wise.

PR #528, opened 2023-08-04
OCPBUGS-15158: Scope Marketplace Operator RBAC

awgreene:
Problem: The Marketplace Operator historically was responsible for making content available on cluster and serving it through pods. The Marketplace Operator has since evolved to simply manage the state of default catalogSource CRs based on specifications in the OperatorHub resource.

Solution: Remove any RBAC that is no longer necessary for the Marketplace Operator to serve its purpose.

### DIFF
--- a/manifests/05_role.yaml
+++ b/manifests/05_role.yaml
@@ -9,90 +9,40 @@ metadata:
     capability.openshift.io/name: "marketplace"
 rules:
 - apiGroups:
-  - ""
+  - config.openshift.io
   resources:
-  - configmaps
-  - namespaces
+  - clusteroperators
+  - operatorhubs
   verbs:
   - get
   - list
   - watch
 - apiGroups:
-  - ""
+  - config.openshift.io
+  resourceNames:
+  - marketplace
   resources:
-  - configmaps
+  - clusteroperators/status
   verbs:
+  - patch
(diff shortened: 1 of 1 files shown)


awgreene:
/hold
I'm going to double check that we're using only the RBAC we need by running the e2e suite and checking the RBAC used by the Marketplace Opeator using the [audit2rbac tool](https://github.com/liggitt/audit2rbac).

manifests/05_role.yaml
```diff
@@ -106,53 +40,16 @@ metadata:
...
-  - roles
-  - rolebindings
+  - catalogsources
   verbs:
   - get
   - create
   - delete
   - update
+  - list
+  - watch
```
anik120, line 52: git diff is making it look like we're adding these verbs new here. Why do we need to add these verbs, if they're in fact new?

stevekuznetsov: Changing from RBAC to catalogsources in the rule

awgreene:
The e2e suit worked locally, retesting.
/retest

awgreene:
Updated based on findings from audit2rbac.
/unhold

awgreene:
/hold
I'm going to play around with this more on Monday

PR #530, opened 2023-08-08
OCPBUGS-17157: go.mod: bump operator-framework/api

stevekuznetsov:
The clients in this controller deserialize the CatalogSource objects that are defaulted, so the version of the API in the client libraries of this controller determines what fields exist in the objects that get created, regardless of what the server knows about.


stevekuznetsov:
Looks like the setting for community operators is too low, will fix later.

jianzhangbjz:
It failed at `  Error: open /proc/8505/stat: no such file or directory: container process not found`, details in https://issues.redhat.com/browse/OCPBUGS-17157
/hold

PR #570, opened 2024-07-24
[release-4.15] OCPBUGS-37524: Set required-scc for openshift workloads

kramaranya:
Cherry-pick of #561 
[OCPBUGS-37524](https://issues.redhat.com/browse/OCPBUGS-37524)

This PR explicitly sets the required SCC to be used to admit pods of the `marketplace-operator` deployments. The SCC chosen is the one that the pods are already getting admitted with, which means that this brings no change to the SCC used.

In some cases, custom SCCs can have higher priority than default SCCs, which means that they will be chosen over the default ones. This can lead to unexpected results; in order to protect openshift workloads from this, we must explicitly pin the required SCC to all our workloads in order to make sure that the expected one will be used.


everettraven (approved):
/lgtm

PR #572, opened 2024-08-21
 OCPBUGS-38813: Replace 4.16 image references with 4.17

dtfranz:


dtfranz:
I'm noticing a pattern with this test failing:
`[sig-operator] an end user can use OLM can subscribe to the operator [apigroup:config.openshift.io]`

Running this one more time and I'll try and see if this could be connected to the changes somehow.
/test e2e-gcp

perdasilva:
:thinking: is this a CI platform issue? [events under openshift-operator-lifecycle-manager](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/operator-framework_operator-marketplace/572/pull-ci-operator-framework-operator-marketplace-master-e2e-gcp/1829078238459072512/artifacts/e2e-gcp/gather-must-gather/artifacts/event-filter.html)

dtfranz:
Sorry, should have posted here as well as slack. The `cluster-logging` operator is unavailable in the 4.17 catalog, which the `e2e-gcp` test uses. There may also be some other CI issues going on, but this is separate.

PR #592, opened 2025-02-28
OCPBUGS-52340: Ensure that PSA label use k8s version supported on the project

camilamacedo86:
In this commit, a new Makefile target `update-k8s-values` was created to automatically update the
`pod-security.kubernetes.io/*-version` values (`enforceVersion`, `auditVersion`, `warnVersion`)
in the Helm chart's `values.yaml` file.

These values now align with the Kubernetes API version defined in `go.mod`, instead of using `latest`. This ensures better compatibility and avoids issues with unsupported versions in Kubernetes PSA.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,3 +31,24 @@ vendor:
 .PHONY: manifests
 manifests:
 	./hack/update-manifests.sh
+
+KUBE_MINOR ?= $(shell go list -m k8s.io/client-go | cut -d" " -f2 | sed 's/^v0\.\([[:digit:]]\{1,\}\)\.[[:digit:]]\{1,\}$$/1.\1/')
+.PHONY: update-k8s-values  # HELP Update PSA labels with k8s version used
+update-k8s-values:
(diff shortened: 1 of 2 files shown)


camilamacedo86:
I will need to create an OCPBUG for each version :-( 
I also pushed the check: https://github.com/openshift/release/pull/62239 
But we can only merge that after cherry-picking it up to the OCP release 4.16. 

@tmshort

camilamacedo86:
/test e2e-gcp-console-olm

The error faced has no relation with the change :-(

jianzhangbjz:
Based on https://redhat-internal.slack.com/archives/C06KP34REFJ/p1741224828143489?thread_ts=1739880491.760029&cid=C06KP34REFJ, we should use the `latest` instead of the pinned version.

camilamacedo86:
Closing since QE would like to set latest instead

jianzhangbjz:
This PR was replaced by https://github.com/operator-framework/operator-marketplace/pull/600

PR #596, opened 2025-03-05
OCPBUGS-52362: [release-4.18] Upgrade dependencies and k8s versions to match with OCP release version

camilamacedo86:
In this commit we are upgrading the k8s version deps from v0.29.2 to v0.31.0


camilamacedo86:
According to the discussion, we should not do that for a bug fix.
We can fix for 4.19, but leave it as it is unless someone asks for that.

PR #599, opened 2025-03-05
NO-ISSUE: Add dependabot to help keep the project maintained

camilamacedo86:
Add dependabot and workflow to run go mod tidy and vendor for the opened PRs

Help us to keep this project maintained

**Reviewer Checklist**

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      prefix: "NO-ISSUE:"
+


perdasilva (approved):

anik120 (changes requested):
What is the motivation behind this PR? 

From what I can understand, it's introducing a GitHub action to run regularly, and the action is checking out the repo, doing a `go mod tidy` and `go mod vendor` and then doing a `git push`? Where is this pushing to? https://blog.frankel.ch/dont-git-push/

camilamacedo86:
Hi @anik120 

The issue is that Marketplace-Operator has not been properly updated.
What we need is a way to be notified when a dependency needs to be updated there. The main problem is that we must be informed when things are outdated.
Dependabot alone is enough for that, so I have adjusted it to rely only on Dependabot.
If we see an open PR, we will be aware of the update and can manually create a PR with the vendoring if needed.

Does that make sense?

anik120:
> Hi @anik120
> 
> The issue is that Marketplace-Operator has not been properly updated. What we need is a way to be notified when a dependency needs to be updated there. The main problem is that we must be informed when things are outdated. Dependabot alone is enough for that, so I have adjusted it to rely only on Dependabot. If we see an open PR, we will be aware of the update and can manually create a PR with the vendoring if needed.
> 
> Does that make sense?

If this PR is trying to add Dependabot, my way of doing it would be something like https://github.com/operator-framework/operator-lifecycle-manager/pull/3226/files

anik120 (approved):
/lgtm

PR #647, opened 2025-07-24
OPRUN-3981: Upgrade Go from 1.23 to 1.24.3 and k8s deps to 1.33 for release 4.20

camilamacedo86:

Closes: https://github.com/operator-framework/operator-marketplace/pull/643
Closes: https://github.com/operator-framework/operator-marketplace/pull/640
**Reviewer Checklist**


camilamacedo86:
@perdasilva 🤞

pkg/status/status.go
```diff
@@ -246,6 +248,7 @@ func NewReporter(cfg *rest.Config, mgr manager.Manager, namespace string, name s
 		stopCh:              stopCh,
 		monitorDoneCh:       make(chan struct{}),
 		clusterOperatorName: name,
+		clock:               clock.RealClock{},
```
camilamacedo86, line 251: Required to fix breaking change

pkg/status/status.go
```diff
@@ -112,7 +114,7 @@ func (r *reporter) setStatusCondition(statusCondition configv1.ClusterOperatorSt
 	if statusCondition.Type == configv1.OperatorAvailable && statusCondition.Status == configv1.ConditionTrue {
 		r.setOperandVersion()
 	}
-	cohelpers.SetStatusCondition(&r.clusterOperator.Status.Conditions, statusCondition)
+	cohelpers.SetStatusCondition(&r.clusterOperator.Status.Conditions, statusCondition, r.clock)
```
camilamacedo86, line 117: Required to fix breaking change in this method

camilamacedo86:
We are not in code freeze yet.
An email and label turn-on was sent stating that the label should have been enabled just one hour ago, and now that the test flakes have passed. 
It seems that the idea is that the label is being used as a "be cautious" signal — not a full "freeze" where only critical fixes are allowed.

So, I believe it's okay to deliver this PR.
This PR updates dependencies to be compatible with the upcoming 4.20 release.
If we don’t merge it, we risk releasing 4.20 with outdated components, including CVEs, Go version and old Kubernetes versions, which are not aligned with what OCP uses.

All tests have passed, so this change appears safe to go inside at this point.

/acknowledge-critical-fixes-only

PR #662, opened 2025-09-17
OCPBUGS-61036: Update memoryTarget on catalog source pods

tmshort:
30Mi is too low, and causes performance issues in redhat-operators Bump them all up to 120Mi (which is what community-operators was set to, and is the largest of the bunch).

**Reviewer Checklist**


joelanford:
/hold

There is some discussion that this may cause issues for small form-factor clusters. Let's make sure those discussions are resolved before this merges.

tmshort:
We're going to do this from 4.17 and down.

jianzhangbjz:
It has been superseded by https://github.com/operator-framework/operator-marketplace/pull/663

PR #663, opened 2025-09-17
OCPBUGS-61036: Update memoryTarget on catalog source pods

tmshort:
The current values of memoryTarget are too low for redhat-operator, redhat-marketplace and certified-operators. The result is that golang GC is run too frequently, and this slows down transfers (causing DealineExceeded errors), increases CPU usage (causing CPU spikes) and because it's the GC looking through all of memory, hits the disk more often than it should.

The value for community-operators is likely too high, and can be reduced to lessen the impact of the other increases.

**Reviewer Checklist**

### DIFF
--- a/defaults/01_redhat_operators.cr.yaml
+++ b/defaults/01_redhat_operators.cr.yaml
@@ -32,7 +32,7 @@ spec:
       operator: "Exists"
       effect: "NoExecute"
       tolerationSeconds: 120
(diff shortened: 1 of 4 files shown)


jianzhangbjz:
Test passed, details: https://issues.redhat.com/browse/OCPBUGS-61036
image
/verified by @jianzhangbjz 
/lgtm
/approve
/label backport-risk-assessed

PR #750, opened 2026-05-15
OPRUN-4598: dynamically resolve catalog image tag from OCP release version

ankitathomas:
Dynamically set default CatalogSource image tags to match the OpenShift cluster version (e.g., v4.23, v5.0) when RELEASE_VERSION is available.

Behavior:
- Only affects default CatalogSources with tagged images
- Preserves digest-based images and non-OpenShift deployments
- Extracts major.minor from RELEASE_VERSION (4.23.0-rc.1 → v4.23)
- Skips if RELEASE_VERSION is unset or "0.0.1-snapshot"
- Skips if the `-default-catalog-version` is not set to `ocp-release` on the command line.
- Only done once at startup for the set of default CatalogSources.

Assisted by: Claude

**Reviewer Checklist**

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"runtime"
+	"strings"
 	"time"
 
 	ca "github.com/operator-framework/operator-marketplace/pkg/certificateauthority"
@@ -56,6 +57,11 @@ const (
 	defaultLeaseDuration               = 90 * time.Second
 	defaultPprofPort                   = 6060
 	healthPort                         = 8080
+
+	// defaultCatalogVersionModeOCPRelease is the value for -default-catalog-version
+	// that enables dynamic catalog tag resolution based on the cluster's OCP version.
+	// This only affects the default/operator-managed CatalogSources.
+	defaultCatalogVersionModeOCPRelease = "ocp-release"
 )
 
 func init() {
@@ -85,13 +91,14 @@ func main() {
 	printVersion()
 
 	var (
-		clusterOperatorName     string
-		tlsKeyPath              string
-		tlsCertPath             string
-		leaderElectionNamespace string
-		pprofAddress            string
-		version                 bool
-		loglvl                  string
+		clusterOperatorName         string
+		tlsKeyPath                  string
+		tlsCertPath                 string
+		leaderElectionNamespace     string
+		pprofAddress                string
+		version                     bool
+		loglvl                      string
+		defaultCatalogVersionMode   string
 	)
 	flag.StringVar(&clusterOperatorName, "clusterOperatorName", "", "configures the name of the OpenShift ClusterOperator that should reflect this operator's status, or the empty string to disable ClusterOperator updates")
 	flag.StringVar(&defaults.Dir, "defaultsDir", "", "configures the directory where the default CatalogSources are stored")
@@ -101,6 +108,7 @@ func main() {
 	flag.StringVar(&tlsCertPath, "tls-cert", "", "Path to use for certificate (requires tls-key)")
 	flag.StringVar(&leaderElectionNamespace, "leader-namespace", "openshift-marketplace", "configures the namespace that will contain the leader election lock")
 	flag.StringVar(&loglvl, "level", "info", "Sets level of logger with default verbosity info level. See https://github.com/sirupsen/logrus for other verbosity levels.")
+	flag.StringVar(&defaultCatalogVersionMode, "default-catalog-version", "", "When set to 'ocp-release', enables automatically updating the image tags of default catalog sources to the OCP version of the cluster they are running on. Other values are ignored.")
 	flag.Parse()
 	logger := logrus.New()
 
@@ -219,8 +227,26 @@ func main() {
 			}
 		}
 
+		var operatorReleaseVersion string
+		if strings.Trim(defaultCatalogVersionMode, "'\"") == defaultCatalogVersionModeOCPRelease {
+			// The RELEASE_VERSION env variable is set to current OpenShift version
+			// by the Cluster Version Operator. If missing or set to the default of
+			// "0.0.1-snapshot", the cluster does not have CVO running, and the image
+			// tag override will be skipped.
+			operatorReleaseVersion = os.Getenv("RELEASE_VERSION")
+		}
+		overrideTag, err := defaults.GetCatalogSourceImageTagOverride(operatorReleaseVersion)
+		if err != nil {
+			overrideTag = ""
+			logger.Warnf("failed to parse RELEASE_VERSION %q for image tag override: %v (skipping override)", operatorReleaseVersion, err)
+		}
+
+		if len(overrideTag) > 0 {
+			logger.Infof("applying image tag override %s to default CatalogSources based on RELEASE_VERSION %s", overrideTag, operatorReleaseVersion)
+		}
+
 		// Populate the global default CatalogSource definitions and config
-		if err := defaults.PopulateGlobals(); err != nil {
+		if err := defaults.PopulateGlobals(overrideTag); err != nil {
 			logger.Fatal(err)
 		}
 
--- a/manifests/09_operator-ibm-cloud-managed.yaml
+++ b/manifests/09_operator-ibm-cloud-managed.yaml
@@ -62,6 +62,7 @@ spec:
             - /var/run/secrets/serving-cert/tls.crt
             - -tls-key
             - /var/run/secrets/serving-cert/tls.key
+            - -default-catalog-version=""
           imagePullPolicy: IfNotPresent
           livenessProbe:
             httpGet:
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -2,11 +2,15 @@ package defaults
 
 import (
 	"context"
+	"fmt"
 	"io/ioutil"
 	"os"
 
 	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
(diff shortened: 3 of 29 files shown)


pkg/defaults/defaults.go
```diff
@@ -144,8 +151,73 @@ func populateDefsConfig(dir string) (map[string]olmv1alpha1.CatalogSource, map[s
...
 		config[catsrc.Name] = false
 	}
 	return catsrcDefinitions, config, nil
 }
+
+func GetCatalogSourceImageTagOverride(versionString string) (string, error) {
+	// Return empty if not in OpenShift or version is default/unknown
+	if len(versionString) == 0 || versionString == defaultCatsrcVersionString {
+		return "", nil
+	}
```
tmshort, line 172: How could `versionString == defaultCatsrcVersionString` ever be true? This function is used at one point, and the value retrieved is from an environment variable, so it will likely be either "5.0.something" or "4.23.something". 
So, I'm not sure what this part of the check is for?

It almost seems to me you want to override the value _only_ if the _existing_ version is `defaultCatsrcVersionString`, which you don't know about in this function. You know about that in the `overrideImageTag` function.

pkg/defaults/defaults.go
```diff
@@ -144,8 +151,73 @@ func populateDefsConfig(dir string) (map[string]olmv1alpha1.CatalogSource, map[s
 			config = make(map[string]bool)
 			return catsrcDefinitions, config, err
 		}
+
+		// Override image tags with ones matching OpenShift <major>.<minor> for
+		// default CatalogSources
+		if err = overrideImageTag(catsrc, imageTagOverride); err != nil {
+			return map[string]olmv1alpha1.CatalogSource{}, map[string]bool{},
+				fmt.Errorf("unable to update image tags for default CatalogSource %s: %w", catsrc.Name, err)
+		}
```
tmshort, line 160: An error in one catalog will block all other catalogs from loading; are we sure we want this behavior? 
Or is this pod only reading one catalog source?

tmshort (commented):
My concern here is that this always seems to be overriding the tag with the value from RELEASE_VERSION. But we currently don't have 4.23 nor 5.0 catalogs available yet. Unless I'm missing something? 

It seems to be that the OLMv1 mechanism ought to be followed; use a "special value" that gets written over. from RELEASE_VERSION. Such that if the reference is "v4.22", we remain using that catalog until it is changed to a value to be replaced.

The special value that was added seems to purposefully _not_ replace the version with RELEASE_VERION.

pkg/defaults/defaults.go
ankitathomas: This only breaks if we ship a bad default catalog, but it will break all default catalogs if we have even one bad one. We can have it be conditional, but the default behavior is to fail for all default catalogs if we have even one malformed one, as in the previous error here: https://github.com/operator-framework/operator-marketplace/blob/8673f5ae72d8bf78eccdfe5854ff8de84ba4c814/pkg/defaults/defaults.go#L141-L146

pkg/defaults/defaults.go
ankitathomas: It will be true for upstream, where CVO does not update the default RELEASE value for the marketplace operator to something besides `0.0.1-snapshot`. The purpose of this special value is so that we don't break the marketplace-operator in case it runs on a cluster without CVO. If we see `0.0.1-snapshot`, we use the default catalogs as they are, so for the current state of the catalogs they would use 4.22 as in https://github.com/operator-framework/operator-marketplace/blob/8673f5ae72d8bf78eccdfe5854ff8de84ba4c814/defaults/01_redhat_operators.cr.yaml#L11

This is based on https://github.com/operator-framework/operator-marketplace/blob/8673f5ae72d8bf78eccdfe5854ff8de84ba4c814/manifests/09_operator.yaml#L91-L93

tmshort: OK, so this is effectively an upstream-only check.

tmshort:
This still seems to _always_ override the catalog version with the RELEASE_VERSION variable. That's a problem because the `v4.23` and `v5.0` don't currently exist; so this will fail. As I suggested above, we should only swap in the RELEASE_VERSION when we're ready to do so (i.e. when those catalogs are ready). In OLMv1 we have a special `ocp-release` tag in the manifests that indicate, "yes, it's OK to replace this with the RELEASE_VERSION"

This is from the failed `e2e-gcp` test:
```
    state:
      waiting:
        message: 'Back-off pulling image "registry.redhat.io/redhat/certified-operator-index:v5.0":
          ErrImagePull: unable to pull image or OCI artifact: pull image err: initializing
          source docker://registry.redhat.io/redhat/certified-operator-index:v5.0:
          reading manifest v5.0 in registry.redhat.io/redhat/certified-operator-index:
          manifest unknown; artifact err: image reference: get manifest from ref:
          create image source: reading manifest v5.0 in registry.redhat.io/redhat/certified-operator-index:
          manifest unknown'
        reason: ImagePullBackOff
```
It's trying to load the `v5.0` version of the catalog, which doesn't exist.

grokspawn:
> My concern here is that this always seems to be overriding the tag with the value from RELEASE_VERSION. But we currently don't have 4.23 nor 5.0 catalogs available yet. Unless I'm missing something?
> 
> It seems to be that the OLMv1 mechanism ought to be followed; use a "special value" that gets written over. from RELEASE_VERSION. Such that if the reference is "v4.22", we remain using that catalog until it is changed to a value to be replaced.
> 
> The special value that was added seems to purposefully _not_ replace the version with RELEASE_VERION.

100% agree.  This functionality cannot be applied until some later date when the catalogs are all available under the desired tag.  It may be necessary in future releases to remove the sentinel from the config for some time after a release to avoid relying on nonexistent catalog images. 

If we update the default manifests to use a sentinel versioned tag in [.spec.image](https://github.com/operator-framework/operator-marketplace/blob/4ff8a12f5885e93ffdbd8039f78c7725492f6ffa/defaults/01_redhat_operators.cr.yaml#L11) like 'ocp-release' which will trigger the behavior. 

When discovering a match, then the tag replacement occurs based on RELEASE_VERSION.  

This allows current catalogs to function (including CI), and controlling the future behavior by updating the manifests.

catalog.version
tmshort, line 1: What is this file for?

manifests/09_operator-ibm-cloud-managed.yaml
```diff
@@ -5,9 +5,7 @@ metadata:
   namespace: openshift-marketplace
   annotations:
     config.openshift.io/inject-proxy: "marketplace-operator"
-    include.release.openshift.io/hypershift: "true"
     include.release.openshift.io/ibm-cloud-managed: "true"
-    capability.openshift.io/name: "marketplace"
```
tmshort, line 10: Why are these removed?

pkg/defaults/defaults.go
```diff
@@ -108,29 +113,31 @@ func IsDefaultSource(name string) bool {
...
-func populateDefsConfig(dir string) (map[string]olmv1alpha1.CatalogSource, map[string]bool, error) {
+func populateDefsConfig(dir, imageTagOverride string) (map[string]olmv1alpha1.CatalogSource, map[string]bool, error) {
 	catsrcDefinitions := make(map[string]olmv1alpha1.CatalogSource)
 	config := make(map[string]bool)
 	// Default directory has not been specified
 	if dir == "" {
 		return catsrcDefinitions, config, nil
 	}
 
-	_, err := os.Stat(Dir)
```
tmshort, line 128: This seems like it was a bug!

tmshort (commented):
/approve

rashmigottipati (approved):

catalog.version
ankitathomas: I missed cleaning up that catalog.version file from a previous version. Good catch, thanks!

pkg/defaults/defaults.go
ankitathomas: yeah, caught that while working on this so I thought I might as well fix it

manifests/09_operator-ibm-cloud-managed.yaml
ankitathomas: This seems to be what hack/update-manifests.sh does. I suspect we never updated it to add those annotations automatically to the ibm-cloud-managed manifests. Good catch, I've added a fix

ankitathomas (commented):
Updated to remove some leftover files from an earlier iteration, and fixed manifest generation

cmd/manager/main.go
```diff
@@ -219,8 +227,26 @@ func main() {
 			}
 		}
 
+		var operatorReleaseVersion string
+		if strings.Trim(defaultCatalogVersionMode, "'\"") == defaultCatalogVersionModeOCPRelease {
```
ankitathomas, line 231: Trimming quotes for `defaultCatalogVersionMode`  because of how kubernetes manifests pass quotes in commandline arguments, specifically when formatted in the following ways on the deployment manifest

```
args:
- -default-catalog-version="ocp-release"
```
```
args:
- -default-catalog-version
- "ocp-release"
```

rashmigottipati (approved):
/lgtm

rashmigottipati (commented):
/test e2e-gcp-ovn-upgrade

perdasilva:
/verified by @perdasilva 

- launched 4.23 cluster
- checked that setting `-default-catalog-version` to `"ocp-release"` as well as `ocp-release` updates the catalogs to tag :v4.23
- observed expected log line, e.g. `time="2026-06-01T08:54:21Z" level=info msg="applying image tag override v4.23 to default CatalogSources based on RELEASE_VERSION 4.23.0-0-2026-06-01-075432-test-ci-ln-r7pxb12-latest"`
- checked that setting `-default-catalog-version` to `""` returns the tag :v4.22

